PR #7, opened 2017-05-09
Openshift template

DuncanDoyle:
Initial implementation of the OpenShift template. Still needs some finetuning. Also misses Route configuration (which we'll add later).


DuncanDoyle:
I think we can merge this one. S2I build still does not properly work in OpenShift Online, so that needs a binary build. But that's something we need to fix at a later stage.

PR #12, opened 2017-08-29
[PLANNER-871] Updated project to use gwt:codeserver instead of gwt:devmode

Christopher-Chianelli:
This PR allows the project to be run using gwt:codeserver instead of gwt:devmode. However, there is one issue: The Roster REST methods do not work:

- rest/tenant/{tenantId}/roster/spotRosterView/current (GET)
- rest/tenant/{tenantId}/roster/spotRosterView?startDate={startDate}&endDate={endDate} (GET)
- rest/tenant/{tenantId}/roster/employeeRosterView/current (GET)
- rest/tenant/{tenantId}/roster/employeeRosterView?startDate={startDate}&endDate={endDate} (GET)

After from analysing, I found the source of the issue: the Roster REST methods use `LocalDateTime`, which gwt jackson is having trouble serializing:
```
java.lang.NoSuchMethodError: com.fasterxml.jackson.datatype.jsr310.ser.JSR310FormattedSerializerBase.findFormatOverrides(Lcom/fasterxml/jackson/databind/SerializerProvider;Lcom/fasterxml/jackson/databind/BeanProperty;Ljava/lang/Class;)Lcom/fasterxml/jackson/annotation/JsonFormat$Value;
	at com.fasterxml.jackson.datatype.jsr310.ser.JSR310FormattedSerializerBase.createContextual(JSR310FormattedSerializerBase.java:92)
	at com.fasterxml.jackson.datatype.jsr310.ser.LocalDateSerializer.createContextual(LocalDateSerializer.java:36)
	at com.fasterxml.jackson.databind.SerializerProvider.handlePrimaryContextualization(SerializerProvider.java:875)...
```
This is odd since `OptaShiftRosteringGwtJacksonConfiguration` seem like it handles this, but it appears the error is happening server side. Any ideas?


ge0ffrey:
Jackson (= server side) has support to serialize java.time and it's already using that, find in path (ctrl-shift-f in IntelliJ) for "<artifactId>jackson-datatype-jsr310</artifactId>".
Gwtjackson (= client side) doesn't, but OptaShiftRosteringGwtJacksonConfiguration.java fixes that, through rebinds.

Not sure how these changes break that from working, but I am not too surprised that it's brittle...

ge0ffrey:
NoSuchMethodError implies **the wildfly version is not correct** (at least not the one I am using). I 've seen that with an SA before.

We need to make this less brittle, so it can work with any wildfly version.

Christopher-Chianelli:
That fixed it! Thanks @ge0ffrey.

pom.xml
```diff
@@ -33,13 +33,23 @@
         <updatePolicy>daily</updatePolicy>
       </snapshots>
     </repository>
+
+    <repository>
+      <id>mvn-central</id>
+      <name>Maven Central</name>
```
ge0ffrey: no need to add maven central explicitly

pom.xml
```diff
@@ -33,13 +33,23 @@
...
+      <url>http://central.maven.org/maven2/</url>
+    </repository>
   </repositories>
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
-    <wildfly.home>../local/appserver/wildfly</wildfly.home>
+    <wildfly.version>10.1.0.Final</wildfly.version>
```
ge0ffrey: version properties should be in the form version.groupId, see 5 lines below for examples

optashift-employee-rostering-webapp/pom.xml
```diff
@@ -46,6 +46,50 @@
...
+        <artifactId>maven-war-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>war-exploded</id>
+            <phase>prepare-package</phase>
+            <goals>
+              <goal>exploded</goal>
+            </goals>
+            <configuration>
+              <webappDirectory>${wildfly.home}/standalone/deployments/${project.build.finalName}.war</webappDirectory>
```
ge0ffrey: By default, the exploded war directory is generated by the maven-war-plugin, IIRC.
So instead of needing this configuration by exception, can we avoid configuring the war-plugin by telling the wildfly plugin to take the standard exploded war dir? (And why isn't it doing that by default already?)

Less is more.

optashift-employee-rostering-webapp/pom.xml
```diff
@@ -46,6 +46,50 @@
...
+          </execution>
+        </executions>
+      </plugin>
+      
+      <plugin>
+        <artifactId>maven-resources-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>copy-resources</id>
+            <!-- here the phase you need -->
```
ge0ffrey: Most comments are good: those that explain complexity, non-obvious things and when some code is a workaround or hack (which information when that will become obsolete).
Comments that explain what the code already explains, or what the javadoc/maven pom spec already explains, aren't useful. Let's remove this line.

optashift-employee-rostering-webapp/pom.xml
```diff
@@ -46,6 +46,50 @@
...
+            <phase>package</phase>
+            <goals>
+              <goal>copy-resources</goal>
+            </goals>
+            <configuration>
+              <outputDirectory>${wildfly.home}/standalone/deployments/${project.build.finalName}.war/gwtui</outputDirectory>
+              <resources>          
+                <resource>
+                  <directory>${project.root}/optashift-employee-rostering-gwtui/src/main/resources/org/optaplanner/openshift/employeerostering/gwtui/public</directory>
+                  <filtering>true</filtering>
```
ge0ffrey: This PR enables filtering for those files (which can have dangerous consequences for binary files),
but I don't see changes in this PR to those files, to take advantage of that filtering?

pom.xml
```diff
@@ -194,11 +227,62 @@
           </execution>
         </executions>
       </plugin>
+
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
```
ge0ffrey: +1 to have this dependency-plugin do this, as it makes setting up the project a lot easier. very good idea.

However, this shouldn't happen in the openshift build on openshift online, as that's just going to slow down the build a lot.
So we don't want this to happen if `-Popenshift` is active.

pom.xml
```diff
@@ -236,20 +326,39 @@
...
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-war-plugin</artifactId>
           <version>2.6</version>
         </plugin>
+
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-resources-plugin</artifactId>
+          <version>3.0.2</version>
+        </plugin>
```
ge0ffrey: Can this be removed as discussed?

Christopher-Chianelli: Thanks for catching this; I modified it so it doesn't require `maven-resources-plugin` anymore, but forgot to remove it from `dependencyManagement`.

PR #13, opened 2017-08-31
[PLANNER-873] Added i18n

Christopher-Chianelli:


Christopher-Chianelli:
@ge0ffrey How do you like the naming the constants? Is there a better naming scheme or are they good as they are? I'm pretty sure I got everything, but let me know if I missed something.

.gitignore
```diff
@@ -1,4 +1,5 @@
 /target
+/bin
```
ge0ffrey: Any idea which tool is creating files in the bin directory?

All generated files from sources go to target (if they get cleaned) or local (if they should not get cleaned with "mvn clean"). Normally, Eclipse and IntelliJ pick this up correctly after being configured from the pom.xml's.

optashift-employee-rostering-gwtui/pom.xml
```diff
@@ -102,6 +102,27 @@
           <moduleName>org.optaplanner.openshift.employeerostering.gwtui.WorkerRosteringWebapp</moduleName>
         </configuration>
       </plugin>
+
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-resources-plugin</artifactId>
+        <version>2.3</version>
```
ge0ffrey: optashift-employee-rostering-gwtui needs to use the same plugin versions as the rest of the build.
So maven plugin versions should be set (if needed at all) in the parent pom, so in /pom.xml instead of the child pom.

I think we want the parent pom to extend the jboss parent pom so we don't need to set up the vanilla plugin versions:

```
  <parent>
    <groupId>org.jboss</groupId>
    <artifactId>jboss-parent</artifactId>
    <version>22</version>
  </parent>
```

(check if 22 is the latest)

optashift-employee-rostering-gwtui/pom.xml
```diff
@@ -102,6 +102,27 @@
...
+      </plugin>
     </plugins>
+    <resources>
+      <resource>
+        <directory>src/main/java</directory>
+        <includes>
+          <include>**/*.html</include>
+          <include>**/*.css</include>
```
ge0ffrey: Let's talk on bluejeans about this, I want to understand it.

optashift-employee-rostering-gwtui/pom.xml
```diff
@@ -102,6 +102,27 @@
...
+        </includes>
+      </resource>
+      <resource>
+        <directory>src/main/resources</directory>
+        <includes>
+          <include>**/*.properties</include>
```
ge0ffrey: this means that all non properties files aren't resources any more?
What if I have a png image resource?

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/resources/i18n/OptaShiftUIConstants.java
```diff
@@ -0,0 +1,90 @@
...
+	final String REFRESH = "REFRESH";
+	
+	@TranslationKey(defaultValue="Rest API")
+    final String REST_API = "REST_API";
+	
+	@TranslationKey(defaultValue="seconds")
+    final String SECONDS = "SECONDS";
+
+	@TranslationKey(defaultValue="Solving for another")
+    final String SOLVING_FOR = "SOLVING_FOR";
```
ge0ffrey: identation

optashift-employee-rostering-gwtui/src/main/resources/org/optaplanner/openshift/employeerostering/gwtui/client/resources/i18n/OptaShiftUIConstants_en.properties
```diff
@@ -0,0 +1,28 @@
+ACTIONS=Actions
+ADD=Add
+CLICK_TO_START_SOLVING=Click the <i>Solve</i> button to start solving.
```
ge0ffrey: Does wb also use capitals and underscore_case (as opposed to normal camelCase) for i18n keys?

optashift-employee-rostering-gwtui/src/main/resources/org/optaplanner/openshift/employeerostering/gwtui/client/resources/i18n/OptaShiftUIConstants_en.properties
```diff
@@ -0,0 +1,28 @@
...
+DELETE=Delete
+
+EMPLOYEE=Employee
+EMPLOYEE_NAME=Employee name
```
ge0ffrey: What would be the key for referencing the class concept "EmployeeAvailability"?
What would be the key for referencing the field concept "Employee.name"?

Ideally, the class/field -> i18n key mapping is reversable.
So FOO_BAR is either class FooBar or field Foo.bar

.gitignore
Christopher-Chianelli: Eclipse is; I think I imported the project before installing m2e, then reimport as a Maven project afterwards.

optashift-employee-rostering-gwtui/pom.xml
Christopher-Chianelli: I think this is from using codeserver instead of devmode. All html and css files were in src/main/java instead of src/main/resources. If I run gwt:codeserver without this, these files will not be in the classes directory, causing errai to complain. Perhaps they should be moved to src/main/resources.

optashift-employee-rostering-gwtui/pom.xml
Christopher-Chianelli: Good point, probably should change to **/* (I think maven wasn't copying resources after adding the src/main/java as a resource directory)

optashift-employee-rostering-gwtui/src/main/resources/org/optaplanner/openshift/employeerostering/gwtui/client/resources/i18n/OptaShiftUIConstants_en.properties
Christopher-Chianelli: They use a modified form of camelCase, where the first letter is also capitalized (so "variable name" becomes "VariableName"). See here: https://github.com/kiegroup/kie-wb-common/blob/500f4ab3315f6d1a62f9d20ff0f72372361644b1/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-client/src/main/java/org/kie/workbench/common/screens/server/management/client/container/status/empty/ContainerStatusEmptyView.html

optashift-employee-rostering-gwtui/src/main/resources/org/optaplanner/openshift/employeerostering/gwtui/client/resources/i18n/OptaShiftUIConstants_en.properties
Christopher-Chianelli: The issue here is many of the constants are shared (Name, Actions, Delete, Employee, Spot) and don't really belong to any particular class. Should we have a general "class" for constants that are used in many different files? (So "Name" becomes "General.Name", class "FooBar" is "FooBar", and field "bar" of "Foo" is "Foo.bar")

optashift-employee-rostering-gwtui/pom.xml
ge0ffrey: +1 to move them instead of messing with the resources plugin

Let's talk on bluejeans anyway if you have some time in your morning (my afternoon), I want to understand this setup and it's not entirely clear to me.

optashift-employee-rostering-gwtui/pom.xml
ge0ffrey: Specifying no includes should do the trick too, and it's less code to maintain (less configuration by exception, so less error prone too)

optashift-employee-rostering-gwtui/src/main/resources/org/optaplanner/openshift/employeerostering/gwtui/client/resources/i18n/OptaShiftUIConstants_en.properties
ge0ffrey: We should stick to wb conventions and follow that form of CamelCase.

optashift-employee-rostering-gwtui/src/main/resources/org/optaplanner/openshift/employeerostering/gwtui/client/resources/i18n/OptaShiftUIConstants_en.properties
ge0ffrey: What do the wb projects do in this situation?

.gitignore
ge0ffrey: ok, let's remove the bin directory from the gitignores.
Anyone accidentally creating it, will then see those files appear, switch to m2e and delete that bin dir, instead of leaving it around and potentially confuse the developers with the target dir.

optashift-employee-rostering-gwtui/src/main/resources/org/optaplanner/openshift/employeerostering/gwtui/client/resources/i18n/OptaShiftUIConstants_en.properties
Christopher-Chianelli: Couldn't find an example in kie-wb; look like they are duplicating values.

ge0ffrey:
Good to merge, but there are merge conflicts, probably due to the other PR in parallel. As the code base grows, there should be less of these.
@Christopher-Chianelli would you mind looking at resolving the merge conflicts?

Christopher-Chianelli:
@ge0ffrey Done. There weren't any proper merge conflicts (since git rebase worked without conflicts), but there were whitespace errors in the patches.

ge0ffrey:
Thanks Christopher :)

PR #14, opened 2017-09-12
[PLANNER-876] Added integrations tests for SkillRestService

Christopher-Chianelli:


ge0ffrey:
Discussed with Christopher - he'll investigate an arquillian approach first, so the REST and JSON code is tested too. Other options might be selenium or a Mockito mockout of EntityManager.
In any case, we don't want to implement and maintain a MockEntityManagers ourself.

Christopher-Chianelli:
@ge0ffrey How do I tell Jenkins the working directory for the test?

optashift-employee-rostering-server/src/test/java/org/optaplanner/openshift/employeerostering/server/skill/AbstractClientArquillianTest.java
```diff
@@ -0,0 +1,46 @@
...
+    public static WebArchive createDeployment() {
+        File file = findPomFile();
+        return ShrinkWrap.create(MavenImporter.class)
+                .loadPomFromFile(file)
+                .importBuildOutput()
+                .as(WebArchive.class);
+    }
+
+    private static File findPomFile() {
+        File file = new File("pom.xml");
```
ge0ffrey: this means it's using the working directory,which should be the module directory as per maven standards. optaplanner-webexamples does this assumption too and there it works fine.

optashift-employee-rostering-webapp/src/test/resources/arquillian.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<arquillian xmlns="http://jboss.org/schema/arquillian"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xsi:schemaLocation="http://jboss.org/schema/arquillian http://jboss.org/schema/arquillian/arquillian_1_0.xsd">
+
+  <container qualifier="wildfly-managed" default="true">
+    <configuration>
+      <property name="jbossHome">../local/appserver/wildfly-10.1.0.Final</property>
```
ge0ffrey: Can we use `<property name="jbossHome">target/wildfly-${version.org.wildfly}</property>` just like in optaplanner-webexamples to avoid hard coding the wildfly version?

optashift-employee-rostering-webapp/src/test/java/org/optaplanner/openshift/employeerostering/webapp/skill/SkillRestServiceTest.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+    
+    private Optional<Skill> getSkill(List<Skill> skills, String name) {
+        return skills.stream().filter((s) -> s.getName().equals(name)).findFirst(); 
+    }
+    
+    private static int generateDatabase(URL baseUrl) {
+        
+        
+        ResteasyClient client = new ResteasyClientBuilder().build();
```
ge0ffrey: code style: double empty line

optashift-employee-rostering-webapp/pom.xml
```diff
@@ -44,6 +44,110 @@
...
+      <groupId>org.jboss.shrinkwrap.resolver</groupId>
+      <artifactId>shrinkwrap-resolver-depchain</artifactId>
+      <type>pom</type>
+      <scope>test</scope>
+    </dependency>
+
+
+    <dependency>
+      <groupId>commons-logging</groupId>
+      <artifactId>commons-logging</artifactId>
```
ge0ffrey: commons-logging is bad - are we sure we need it? 
Normally we only have slf4j (= the api), logback (= the slf4j impl) and if needed bridges that reroute messages to other frameworks into slf4j.

optashift-employee-rostering-webapp/pom.xml
```diff
@@ -44,6 +44,110 @@
...
+      <version>4.12</version>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>2.10.0</version>
+      <scope>test</scope>
+    </dependency>
```
ge0ffrey: All version declerations should be in dependencyManagement in the root pom.

optashift-employee-rostering-webapp/src/test/resources/arquillian.xml
Christopher-Chianelli: No; the arquillian configuration file does not share properties with pom.xml. I got the error "jbossHome &apos;../local/appserver/wildfly-${version.org.wildfly}&apos; must exist" when using that setup.

optashift-employee-rostering-webapp/pom.xml
Christopher-Chianelli: It appears arquillian requires commons-logging to be present at runtime to work. Without this dependency, I get the following error message for tests: org/apache/commons/logging/LogFactory" type="java.lang.NoClassDefFoundError">java.lang.NoClassDefFoundError: org/apache/commons/logging/LogFactory

optashift-employee-rostering-webapp/pom.xml
Christopher-Chianelli: I would normally agree to this, but arquillian is really picky about what dependencies you are allowed to have. Many of these dependencies are out-dated and are only there since arquilian requires them. Updating them to the one the other modules use breaks the test, and I don't want the other module using the out-dated dependency. I'll try to make it clearer.

Christopher-Chianelli:
@ge0ffrey I fixed the code styling issue and remove versions where possible, creating properties for the dependencies required by arquillian to work. Unfortunately, I cannot remove commons-logging since arquillian requires it to work. Additionally, I cannot use pom properties in arquillian's configuration file.

PR #15, opened 2017-09-12
[PLANNER-874] Display a helpful error message when using wrong WildFly version

Christopher-Chianelli:
Without directly modifying the WildFly instance they are using, there is no way to intercept the Exception from the server code in this repo (since none of the server code is called), so the next best thing is to put an error message telling them what WildFly version to use.


optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/roster/EmployeeRosterViewPanel.java
```diff
@@ -94,6 +94,10 @@ protected void refreshTable() {
             @Override
             public void onSuccess(EmployeeRosterView employeeRosterView) {
                 EmployeeRosterViewPanel.this.employeeRosterView = employeeRosterView;
+                if (null == employeeRosterView) {
```
ge0ffrey: This might be null due to other reasons.
I guess we can't detect it on the server side? as it happens during the json marshalling?

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/roster/EmployeeRosterViewPanel.java
```diff
@@ -94,6 +94,10 @@ protected void refreshTable() {
...
+                    solveStatus.setTextContent("You are using the wrong version of Wildfly. Please use Wildfly 10.1.0.Final");
```
ge0ffrey: We need to have a decent error handler on the client side to show all kinds of errors in a clear, popup with a red icon. This is a good temporary workaround, so let's add a comment "// TODO use error handler"

We probably don't want to hardcode the wildfly version in the error message, but we do want to mention it indeed.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/roster/EmployeeRosterViewPanel.java
Christopher-Chianelli: Yeah; none of our server-side code is touched unfortunately. The only way to detect it server side that I can think of is to expose a new REST endpoint, serverup, which returns a JSON with a boolean "isUp", which is true iff the server is working property. To calculate if the server is running property, we can try {...}catch(Exception e){...} the call which throws the exception. But I think that kinda pointless. If there a way for the server to process the html files before severing them, then we can do a pre-emptive check using the above method, and return an error HTML page if it is setup incorrectly.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/roster/EmployeeRosterViewPanel.java
Christopher-Chianelli: There is an ErrorPopupPresenter in uberfire, but none in Errai. I can easily create one using DialogBox (which is a generic popup).

ge0ffrey: +1 good idea

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/roster/EmployeeRosterViewPanel.java
ge0ffrey: Agreed. Let's add some doubt in the message to the user (the word "maybe"), in case it's not failing due to the wrong WildFly version, for example let's change
  "You are using the wrong version of Wildfly. Please use Wildfly 10.1.0.Final"
into
  "The server returned an empty result which is impossible.\nMaybe the server is using the wrong WildFly version. Try using WildFly 10.1.0.Final."

PR #17, opened 2017-09-25
[PLANNER-889] Server's Error Stack Trace is shown in a Error Popup on Client

Christopher-Chianelli:


optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/exceptions/OptashiftExceptionMapper.java
```diff
@@ -0,0 +1,24 @@
+package org.optaplanner.openshift.employeerostering.server.exceptions;
```
ge0ffrey: I'd put this in the package "common", in the philosophy that we have these packages under each module package:
- common (code used by or extended all other packages)
- skill (normal package for functionalty)
- employee (normal package for functionalty)
- spot (normal package for functionalty)
- ... (normal packages based on functionalty, NOT type).
- app (bootstrap code and other code that use all other packages)

I'll merge this PR and do that adjustment myself.

PR #22, opened 2017-10-10
*DO NOT MERGE* PLANNER-900: Added basic REST API for continuous planning

Christopher-Chianelli:
@ge0ffrey I added a basic REST API for simple continuous planning: Added method planUntil(tenantId, baseDate, targetDate, duration), which takes shifts between baseDate and (baseDate + duration), and copies them, except time slots are changed so if a time slot starts at (baseDate + x), it starts at (targetDate + x) instead. Added a simple button to the employeeRosterView to copy all shifts to the last end date.


Christopher-Chianelli:
Closed since this feature isn't useful (since this done via the template anyways)

PR #24, opened 2017-10-19
PLANNER-903: Made all Serializers/Deserializers internal to fix conflicts from different versions

Christopher-Chianelli:
@ge0ffrey This works with Wildfly 10.0.0.Final. I haven't tested on EAP 7 yet; should I test it on EAP 7.1.0 Beta or EAP 7.0.0?

### DIFF
--- a/optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/jackson/LocalDateDeserializer.java
+++ b/optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/jackson/LocalDateDeserializer.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2013 FasterXML.com
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may obtain
+ * a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the license for the specific language governing permissions and
+ * limitations under the license.
+ */
+
+package org.optaplanner.openshift.employeerostering.shared.jackson;
+
+import java.io.IOException;
+import java.time.DateTimeException;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.core.JsonTokenId;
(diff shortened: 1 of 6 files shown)


ge0ffrey:
@Christopher-Chianelli EAP 7.1.0 Beta is more important for us because we need to think long term and we need to get all our ducks in a row by the BRMS 7.0 release.

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/jackson/LocalDateDeserializer.java
```diff
@@ -0,0 +1,35 @@
+package org.optaplanner.openshift.employeerostering.shared.jackson;
```
ge0ffrey: Copy the copyright/license text from the original file, for legal purposes.

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/jackson/LocalDateDeserializer.java
```diff
@@ -0,0 +1,35 @@
...
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.core.JsonTokenId;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+
+public class LocalDateDeserializer extends JsonDeserializer<LocalDate> {
```
ge0ffrey: Add a comment that clearly indicates when this class would become obsolete.

By copying them, we don't get automatic bugfixes as we upgrade to newer jackson versions, but that is indeed the lesser evil today. This is a very good, practical PR.
However, that situation won't stay the case: at some point every modern WildFly/EAP version will contain the jackson java.time support out of the box and we no longer need to duplicate them.
The comment should say something like "// TODO remove me when our minimum WildFly version is x and the minimum EAP version is y"

pom.xml
```diff
@@ -63,7 +63,7 @@
     <version.com.github.nmorel.gwtjackson>0.14.2</version.com.github.nmorel.gwtjackson>
     <version.com.github.nmorel.gwtjackson.restprocessor>0.4.1</version.com.github.nmorel.gwtjackson.restprocessor>
     <version.org.gwtbootstrap3>0.9.4</version.org.gwtbootstrap3>
-    <version.org.wildfly.wildfly-dist>10.1.0.Final</version.org.wildfly.wildfly-dist>
+    <version.org.wildfly.wildfly-dist>10.0.0.Final</version.org.wildfly.wildfly-dist>
```
ge0ffrey: we probably still want to push the latest WildfFly final version, despite also being compatible with 10.0

Christopher-Chianelli:
@ge0ffrey I have added your suggested changes

PR #26, opened 2017-11-02
PLANNER-906: Added edit functionality for employee, spot and skill

Christopher-Chianelli:


optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/employee/EmployeeRestServiceImpl.java
```diff
@@ -122,4 +123,29 @@ public Boolean removeEmployeeAvailability(Integer tenantId, Long id) {
...
+    @Override
+    @Transactional
+    public Boolean updateEmployee(Integer tenantId, Long id, Employee newValue) {
+        Employee employee = entityManager.find(Employee.class, id);
+        if (null == employee) {
+            return false;
+        }
+        validateTenantIdParameter(tenantId, employee);
+        employee.setName(newValue.getName());
+        //Hibernate workaround: you cannot set a list directly, only add and remove from it
```
ge0ffrey: Code style: A space after //

Merging regardless, will fix here.

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/employee/EmployeeRestServiceImpl.java
```diff
@@ -122,4 +123,29 @@ public Boolean removeEmployeeAvailability(Integer tenantId, Long id) {
         return true;
     }
 
+    @Override
+    @Transactional
+    public Boolean updateEmployee(Integer tenantId, Long id, Employee newValue) {
+        Employee employee = entityManager.find(Employee.class, id);
```
ge0ffrey: This should use the EntityManager.merge() functionality, which returns a new instance.
There's a bunch of important theory behind this, I can explain in a conf call.

This is stopping the PR merge for now.

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/employee/EmployeeRestService.java
```diff
@@ -40,6 +40,14 @@
     @POST
     @Path("/add")
     Long addEmployee(@PathParam("tenantId") Integer tenantId, Employee employee);
+    
+    /**
+     * @param employee never null
+     * @return never null, the id
```
ge0ffrey: it returns a boolean, not an id

the removeEmployee() method has the same bug, please fix that javadoc.
Let's also see if we need to return anything, boolean or Boolean.

Christopher-Chianelli:
@ge0ffrey Unfortunately `merge()` only works with persistent and detached objects (objects the database is aware of); we have a transient object (object that the database is not aware of), and there is no way to update an existing persistent or detached object from a transient object, since to use a transient object, we would need to call persist or merge, which will create a new row in the database, which we don't want. I managed to make a very generic update function, which updates all the field via reflection, assuming they all have getter and setter methods of the form `getFieldName()` and `setFieldName(value)`. In case either of those don't exist, it does throw an exception telling us which one didn't exist, but I am pretty sure we need to have them anyway.

ge0ffrey:
Using reflection here to avoid the maintenance problem is smart. Too smart. This falls under the category "If you think the problem is bad now, just wait until we fixed it.".

The merge() method indeed works with persistent objects (= already attached objects, so JPA is aware of them) and detached objects (= objects no longer known to JPA, but they have an id). IIRC, our objects (even though they json-serialized to the client and came back) are perfectly fine detached objects, because the have the id.
I 'll pull this branch locally and see if I can sort this out. There must be something I am missing here.

ge0ffrey:
Superseded by https://github.com/kiegroup/optashift-employee-rostering/pull/31 so closing this one

PR #29, opened 2017-11-10
PLANNER-918: Removed the openshift template limitations

Christopher-Chianelli:
I still need to edit the README.doc; I don't know what instructions to put for it. However, I got it running on my local OpenShift instance.

### DIFF
--- a/openshift/templates/optashift-employee-rostering-template.yaml
+++ b/openshift/templates/optashift-employee-rostering-template.yaml
@@ -41,13 +41,6 @@ objects:
     - type: ConfigChange
     - imageChange: {}
       type: ImageChange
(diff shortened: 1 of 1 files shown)


ge0ffrey:
Some people are using OpenShift Online (openshift.com, version 3) so they can't build the github url there due to resource limitations of that free service. The provision.sh script is a workaround for them, usage as defined by the current readme. Others are using OpenShift Dedicated (which is a paid service) or Origin (locally - such as yourself).
The Readme needs to somehow point people to the right thing to use, as quickly (= as short) and as clearly (= bullet or ordered lists instead of paragraphs usually) as possible, depending on what they use.

PR #30, opened 2017-11-15
Updated README

Christopher-Chianelli:
@ge0ffrey Let me know what you think of the updated README.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -8,19 +8,64 @@ See https://www.youtube.com/watch?v=sOWC4qrXxFk[the video demo].
 
 === Prerequisites
 
-. Install https://www.openshift.org/download.html[the oc client tools] and log in.
+. Install https://www.openshift.org/download.html[the oc client tools]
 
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -8,48 +8,76 @@ See https://www.youtube.com/watch?v=sOWC4qrXxFk[the video demo].
...
+`oc login -u=<username> -p=<password> --server=<your-openshift-server> --insecure-skip-tls-verify`
+
+(With OAuth token)
+
+`oc login <https://api.your-openshift-server.com> --token=<tokenID>`
+
+Create the project, build the `war` file locally and provision it on OpenShift:
+
+`./provision.sh setup employee-rostering --binary`
+
```
ge0ffrey: Can we add something here (in a new PR, this is merged already) to motivate why the normal way doesn't work?
Something like:

NOTE
====
Creating a new app from the GitHub URL on OpenShift online doesn't work because the build pod's maven process fails with error code 137 due to lack of memory for the GWT compilation.
====

PR #34, opened 2018-01-05
PLANNER-927: Upgraded Errai and OptaPlanner versions

Christopher-Chianelli:
After an extensively long time trying to find right dependencies versions for everything, I managed to upgrade both Errai and OptaPlanner to the latest versions. Unfortunately, I had to copy one class, `PortablePreconditions.java`, to super sources. Although it should be included in "org.kie.soup.commons.KIESoupCommons" (https://github.com/kiegroup/kie-soup/blob/master/kie-soup-commons/src/main/resources/org/kie/soup/commons/KIESoupCommons.gwt.xml), (https://github.com/kiegroup/kie-soup/blob/master/kie-soup-commons/src/main/java/org/kie/soup/commons/validation/PortablePreconditions.java), the gwt compiler could not find it unless it was in super sources. Additionally, the Arquillian Tests fails on my machine due to missing class definitions, probably same issue as https://stackoverflow.com/questions/46319292/java-lang-classnotfoundexception-org-wildfly-security-permission-abstractnamese, which does not have answer (as they weren't able to reproduce)


optashift-employee-rostering-gwtui/pom.xml
```diff
@@ -91,6 +101,12 @@
       <artifactId>errai-cdi-jboss</artifactId>
       <scope>runtime</scope>
     </dependency>
+    
+    <!-- GWT compiler logger -->
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
```
ge0ffrey: Why not logback just like the other modules? If we end up with 2 impl's in webapp (logback and slf4j-simple) then logging is broken
If this doesn't leak to webapp (transitive dep), can we make it `<dependency>...<optional>true</></>` to be sure?

optashift-employee-rostering-gwtui/src/main/resources/jboss-log4j.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- TODO Remove this file when https://issues.jboss.org/browse/LOGMGR-162 is fixed so logback.xml gets picked up -->
+<log4j:configuration xmlns:log4j="http://jakarta.apache.org/log4j/" debug="false">
```
ge0ffrey: There can only be one logging configuration. Otherwise people don't know which one to use.

optashift-employee-rostering-gwtui/src/main/resources/logback.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- TODO Currently ignored due to https://issues.jboss.org/browse/LOGMGR-162 so use jboss-log4j.xml instead. -->
+<configuration>
```
ge0ffrey: Does this one work? Change the logging to prove it does.

optashift-employee-rostering-webapp/pom.xml
```diff
@@ -28,135 +30,159 @@
...
       <groupId>com.fasterxml.jackson.jaxrs</groupId>
       <artifactId>jackson-jaxrs-base</artifactId>
       <version>${arquillian.version.jackson-jaxrs}</version>
       <scope>test</scope>
     </dependency>
-
+    
+   <dependency>
+     <groupId>dom4j</groupId>
+     <artifactId>dom4j</artifactId>
```
ge0ffrey: Do we really need this one? If it's a workaround, add a comment when the workaround can be removed.

optashift-employee-rostering-webapp/pom.xml
```diff
@@ -28,135 +30,159 @@
...
   </dependencies>
 
   <profiles>
     <profile>
-      <!-- Use this profile for any OpenShift specific customization your app will need. -->
-      <!-- By default that is to put the resulting archive into the 'deployments' folder. -->
+      <!-- Use this profile for any OpenShift specific customization your 
+        app will need. -->
+      <!-- By default that is to put the resulting archive into the 'deployments' 
+        folder. -->
```
ge0ffrey: Either it's 1 big comment (so one ), or it's 1 comment per line. Not 2 comments of 2 lines.

pom.xml
```diff
@@ -144,6 +145,11 @@
         <groupId>org.slf4j</groupId>
         <artifactId>slf4j-api</artifactId>
         <version>${version.org.slf4j}</version>
+      </dependency>
+       <dependency>
```
ge0ffrey: identation

optashift-employee-rostering-gwtui/pom.xml
```diff
@@ -73,6 +73,16 @@
       <groupId>org.jboss.errai</groupId>
       <artifactId>errai-navigation</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.kie.soup</groupId>
+      <artifactId>kie-soup-commons</artifactId>
```
ge0ffrey: @mbarkley This project doesn't depend on kie-wb nor appformer nor uberfire.
How come we need to import kie-soup?

optashift-employee-rostering-gwtui/pom.xml
Christopher-Chianelli: logback appears to be broken/incompatible with the current version of the logging system; using logback causes there to be no logger due to initialization failures (in both webapp and GWT). I can remove this, although it probably better to remove logback.

optashift-employee-rostering-gwtui/pom.xml
Christopher-Chianelli: Errai requires the source code of `Precondition` (https://github.com/kiegroup/kie-soup/blob/master/kie-soup-commons/src/main/java/org/kie/soup/commons/validation/Preconditions.java). It doesn't get imported with the other dependencies currently.

optashift-employee-rostering-webapp/pom.xml
Christopher-Chianelli: Work around for hibernate (https://stackoverflow.com/questions/35533847/wildfly-10-java-lang-classcastexception-org-dom4j-documentfactory-cannot-be-ca). I tried the accepted answer, still got the exception, so I went to the second answer.

optashift-employee-rostering-gwtui/pom.xml
Christopher-Chianelli: @mbarkley You're right, got confused since the log was full of Errai info messages before the error (and some Errai issues from using an incompatible gwt version which is now fixed)

ge0ffrey: Does this mean we can remove the explicit dependency on kie-soup?

mbarkley: I don't know as much about the new GWT maven plugin, but assuming it doesn't automatically add source jars, then @Christopher-Chianelli is correct that you'll need to explicitly add the source jar.

optashift-employee-rostering-gwtui/src/main/module.gwt.xml
```diff
@@ -10,6 +10,7 @@
 
   <extend-configuration-property name="gwtjackson.configuration.extension"
       value="org.optaplanner.openshift.employeerostering.gwtui.rebind.OptaShiftRosteringGwtJacksonConfiguration" />
+  <inherits name="org.kie.soup.commons.KIESoupCommons" />
```
ge0ffrey: Caused by https://issues.jboss.org/browse/ERRAI-1094

pom.xml
```diff
@@ -302,6 +314,7 @@
               <!-- Remote debugger: uncomment this line to debug the WildFly 
                 backend -->
               <!--<arg>-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005</arg> -->
+              <arg>-Derrai.server.classOutput=classes</arg>
```
ge0ffrey: What does this do? Do we need it?

optashift-employee-rostering-webapp/pom.xml
```diff
@@ -28,135 +30,159 @@
...
     <dependency>
       <groupId>org.optaplanner.openshift</groupId>
-      <artifactId>optashift-employee-rostering-server</artifactId>
+      <artifactId>optashift-employee-rostering-gwtui</artifactId>
+      <type>war</type>
+      <scope>runtime</scope>
     </dependency>
     <dependency>
-      <groupId>javax</groupId>
-      <artifactId>javaee-api</artifactId>
```
ge0ffrey: This is the JEE api, this one definitly can't be removed.

optashift-employee-rostering-webapp/pom.xml
```diff
@@ -28,135 +30,159 @@
...
       <artifactId>shrinkwrap-resolver-depchain</artifactId>
       <type>pom</type>
       <scope>test</scope>
     </dependency>
-
-
+    
     <dependency>
       <groupId>commons-logging</groupId>
       <artifactId>commons-logging</artifactId>
```
ge0ffrey: We can't have multiple logging implementations in our classpath.
There can only be slf4j-api (= the api) in all modules and logback (the implementation) only in end-user modules (the war and the benchmarker). All other logging jars, such as slf4j-simple and commons-logging are evil.
slf4j bridges are also allowed (they make sure that if we depend on spring (we don't) and spring logs to commons-logging (it doesn't) that spring's log messages get bridged into slf4j-api and so respect our logback.xml configuration.

ge0ffrey:
Reworked as https://github.com/kiegroup/optashift-employee-rostering/pull/52
So closing this one.

PR #38, opened 2018-01-15
PLANNER-944: Single-Click to change employee avaliability, Shift-Click for more complex edits

Christopher-Chianelli:
@ge0ffrey Let me know what you think; I modified it so a single click without shift key pressed down cycles employee availability (red -> yellow -> green -> gray -> red -> ...). If you single click with shift key pressed down, you get the original popup. I could put an embedded widget that you can click, but that can pose an issue for users with smaller screens.

### DIFF
--- a/optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/common/TimeoutRestCaller.java
+++ b/optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/common/TimeoutRestCaller.java
@@ -1,0 +1,50 @@
+package org.optaplanner.openshift.employeerostering.gwtui.client.common;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import com.google.gwt.user.client.Timer;
+
+public class TimeoutRestCaller {
+
+    private static Map<Object, TimeoutRestCaller> callerMap = new HashMap<>();
+    private Callee callee;
+    private Timer timer;
+
+    private TimeoutRestCaller(Callee callee, Object obj) {
+        this.callee = callee;
+        timer = new Timer() {
+
+            @Override
+            public void run() {
+                doCall(obj);
+            }
+        };
+    }
+
+    public static void call(Object obj, Callee callee) {
+        call(obj, callee, 1000);
+    }
+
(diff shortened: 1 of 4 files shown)


optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/common/TimeoutRestCaller.java
```diff
@@ -0,0 +1,50 @@
...
+        caller.timer.schedule(delayMilliSec);
+    }
+
+    private static void doCall(Object obj) {
+        TimeoutRestCaller caller = callerMap.get(obj);
+        callerMap.remove(obj);
+        caller.callee.call();
+    }
+
+    public interface Callee {
```
ge0ffrey: replace with java.util.Callable

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/common/TimeoutRestCaller.java
```diff
@@ -0,0 +1,50 @@
+package org.optaplanner.openshift.employeerostering.gwtui.client.common;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import com.google.gwt.user.client.Timer;
+
+public class TimeoutRestCaller {
```
ge0ffrey: @mbarkley Do we have anything like this in errai out of the box?

mbarkley: No, I don't think so.

ge0ffrey:
Nobody is going to know about holding the shift key down. Thinking aloud: can we do the hover over approach that they do for jbpm designer?
What this means: if you hover over a shift in the roster window, button appear around (just above and just below) the shift. Above we could do the 3 traffic light buttons (green, yellow, red), below we could do the edit, lock and the remove button?
This is an idea to share with UX in any case, to get their opinion.

Meanwhile, I do think this is an improvement (the shifts can be edited from the spot roster screen), so I am merging.

PR #39, opened 2018-01-15
PLANNER-943: Allow users to specify a default employee for a shift

Christopher-Chianelli:
The main functionality is implemented; But I still need to add more visual indicators to indicate when the employee is not the rotation employee in the spot and employee roster. @ge0ffrey can you verify I did the drools rule correctly?

### DIFF
--- a/optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/roster/RosterGenerator.java
+++ b/optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/roster/RosterGenerator.java
@@ -39,6 +39,7 @@ import org.optaplanner.openshift.employeerostering.shared.employee.EmployeeAvail
 import org.optaplanner.openshift.employeerostering.shared.employee.EmployeeAvailabilityState;
 import org.optaplanner.openshift.employeerostering.shared.lang.tokens.EmployeeTimeSlotInfo;
 import org.optaplanner.openshift.employeerostering.shared.lang.tokens.IdOrGroup;
+import org.optaplanner.openshift.employeerostering.shared.lang.tokens.OptionalEmployee;
 import org.optaplanner.openshift.employeerostering.shared.lang.tokens.ShiftInfo;
 import org.optaplanner.openshift.employeerostering.shared.roster.Roster;
 import org.optaplanner.openshift.employeerostering.shared.shift.Shift;
@@ -98,19 +99,20 @@ public class RosterGenerator {
     @PostConstruct
     public void setUpGeneratedData() {
         tenantNameGenerator.predictMaximumSizeAndReset(10);
-        generateRoster(10, 7, false);
-        generateRoster(10, 7 * 4, false);
-        generateRoster(20, 7 * 4, false);
-        generateRoster(40, 7 * 2, false);
-        generateRoster(80, 7 * 4, false);
-        generateRoster(10, 7 * 4, true);
-        generateRoster(20, 7 * 4, true);
-        generateRoster(40, 7 * 2, true);
-        generateRoster(80, 7 * 4, true);
+        generateRoster(10, 7, false, false);
+        generateRoster(10, 7 * 4, false, false);
+        generateRoster(20, 7 * 4, false, true);
+        generateRoster(40, 7 * 2, false, false);
+        generateRoster(80, 7 * 4, false, true);
+        generateRoster(10, 7 * 4, true, false);
+        generateRoster(20, 7 * 4, true, true);
+        generateRoster(40, 7 * 2, true, false);
+        generateRoster(80, 7 * 4, true, true);
     }
 
     @Transactional
-    public Roster generateRoster(int spotListSize, int timeSlotListSize, boolean continuousPlanning) {
+    public Roster generateRoster(int spotListSize, int timeSlotListSize, boolean continuousPlanning,
+            boolean assignDefaultEmployee) {
         int employeeListSize = spotListSize * 7 / 2;
         int skillListSize = (spotListSize + 4) / 5;
         Integer tenantId = createTenant(spotListSize, employeeListSize);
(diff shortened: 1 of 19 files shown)


optashift-employee-rostering-server/src/main/resources/org/optaplanner/openshift/employeerostering/server/solver/employeeRosteringScoreRules.drl
```diff
@@ -118,3 +118,13 @@ rule "Desired time slot for an employee"
     then
         scoreHolder.addSoftConstraintMatch(kcontext, +$tenantConfiguration.getDesiredTimeSlotWeight());
 end
+
+rule "Employee is rotation employee"
+    when
+    	$tenantConfiguration : TenantConfiguration(rotationEmployeeMatchWeight != 0)
+        Shift(  
+                $e: rotationEmployee,
+                employee == $e)
```
ge0ffrey: Instead of rewarding if it's the same, I 'd punish if it's different. So != instead of == and -weight instead of +weight.

ge0ffrey: The variable $e is ambiguous: I'd expect it to represent the employee, not the rotationEmployee.
In any case, we can remove it, simply inline it.

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/roster/RosterGenerator.java
```diff
@@ -98,19 +99,19 @@ public RosterGenerator(EntityManager entityManager) {
...
+        generateRoster(80, 7 * 4, false, true);
+        generateRoster(10, 7 * 4, true, false);
+        generateRoster(20, 7 * 4, true, true);
+        generateRoster(40, 7 * 2, true, false);
+        generateRoster(80, 7 * 4, true, true);
     }
 
     @Transactional
-    public Roster generateRoster(int spotListSize, int timeSlotListSize, boolean continuousPlanning) {
+    public Roster generateRoster(int spotListSize, int timeSlotListSize, boolean continuousPlanning, boolean isRotation) {
```
ge0ffrey: I believe all datasets should do rotation:
- Either they do a rotation which defines the shift times and for each shift the rotationEmployee
- Or the do a ration that only defines the shift times, but leaves the rotatinEmployee null.

optashift-employee-rostering-server/src/main/resources/org/optaplanner/openshift/employeerostering/server/solver/employeeRosteringScoreRules.drl
ge0ffrey: If rotationEmployee == null, then this rule shouldn't apply (that's probably the reason why you went with a positive constraint). I still think it needs to be a negative constraint, from my experience, because otherwise the score will look funny (and harder to explain to users - a score of 0 might not be perfect).

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/roster/RosterGenerator.java
```diff
@@ -203,6 +203,15 @@ public Roster generateRoster(int spotListSize, int timeSlotListSize, boolean con
                 }
                 shiftSpots.add(new IdOrGroup(tenantId, false, spot.getId()));
             }
+            if (isRotation) {
+                shift.setRotationEmployees(extractRandomSubListOfLength(employees.stream().map((e) -> new OptionalEmployee(tenantId, e)).collect(Collectors.toList()), shiftSpots.size()));
+            } else {
+                List<OptionalEmployee> rotationEmployees = new ArrayList<>(shiftSpots.size());
+                for (int k = 0; k < shiftSpots.size(); k++) {
```
ge0ffrey: use a for each loop instead of introducing k (the jvm might inline that better)

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/roster/RosterGenerator.java
Christopher-Chianelli: Look like I named the parameter poorly; `isRotation` is basically `assignDefaultEmployee`; if true, all shifts get assigned a rotation employee, otherwise all shifts get a rotation employee of null (in short, it behaves like your second bullet point, all datasets use templates (shifts to repeat each week)).

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/lang/tokens/OptionalEmployee.java
```diff
@@ -0,0 +1,32 @@
...
+
+import javax.annotation.Nullable;
+import javax.persistence.Entity;
+import javax.persistence.ManyToOne;
+
+import org.optaplanner.openshift.employeerostering.shared.common.AbstractPersistable;
+import org.optaplanner.openshift.employeerostering.shared.employee.Employee;
+
+@Entity
+public class OptionalEmployee extends AbstractPersistable {
```
ge0ffrey: Why do we need this class? Can we get rid of it by just allow null's of Employee?

Christopher-Chianelli: It needed as `Employee` are used in a List, and Hibernate doesn't behave well with nulls in a list to say the least... (and I cannot use the `Optional<Employee>` class since Hibernate isn't aware of it). (See line 92 of `ShiftInfo`)

optashift-employee-rostering-server/src/main/resources/org/optaplanner/openshift/employeerostering/server/solver/employeeRosteringScoreRules.drl
Christopher-Chianelli: rule "Desired time slot for an employee" is also a positive constraint, should I change it to a negative (since a score of 0 might not be perfect)?

Christopher-Chianelli:
Fixed merge conflicts from renaming variables (and fixed a bug that caused rotation employee to not be assigned)

ge0ffrey:
I need to look into this one, I 'll try beginning of next week.

ge0ffrey:
jenkins test this

ge0ffrey:
@Christopher-Chianelli Can you make this mergable again. I 'll merge - all the non-info stuff looks good and the info classes need a more high-level vision review anyway, so it's fine to merge as is.

PR #43, opened 2018-01-18
PLANNER-954: Allow a spot to require multiple skills

Christopher-Chianelli:


optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/spot/Spot.java
```diff
@@ -31,28 +33,28 @@
...
 
-    @NotNull @Size(min = 1, max = 120)
+    @NotNull
+    @Size(min = 1, max = 120)
     private String name;
     @NotNull
-    @ManyToOne
-    private Skill requiredSkill;
+    @ManyToMany(fetch = FetchType.EAGER)
+    private List<Skill> requiredSkills;
```
ge0ffrey: Change "requiredSkills" to "requiredSkillList" for consistency with the rest of the domain code. Also apply that name change through the rest of this PR (including screens etc).
The philosophy is that we use plural for arrays (items), List suffix for lists (itemList), Set suffix for sets (itemSet), and so forth (itemMap), ...

Christopher-Chianelli:
@ge0ffrey Updated all Collection typed variables to use naming consistent with domain code.

PR #51, opened 2018-01-30
Configure PatternFly 3.18.1

tiagobento:
I'm not entirely sure I'm missing something here because it seems like PatternFly has many dependencies which I didn't add. See: https://mvnrepository.com/artifact/org.webjars.bower/patternfly/3.18.1

IMO, we should stick to this minimal configuration for now and add the other dependencies if we happen to encounter some issues in the future.

`3.18.1` is the version used by the Workbench btw. PatternFly is now at `3.29.13`, but since we're maybe copying stuff from the Workbench to this project, I guess we should stick to this version for now and change in the future if it seems suitable.


optashift-employee-rostering-gwtui/pom.xml
```diff
@@ -14,6 +14,10 @@
 
   <name>Employee Rostering as a Service GWT UI</name>
 
+  <properties>
+    <version.org.webjars.bower.org.patternfly>3.18.1</version.org.webjars.bower.org.patternfly>
```
ge0ffrey: all versions must be declared in the parent pom

optashift-employee-rostering-gwtui/pom.xml
```diff
@@ -119,6 +123,63 @@
           <moduleName>org.optaplanner.openshift.employeerostering.gwtui.WorkerRosteringWebapp</moduleName>
         </configuration>
       </plugin>
+
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
```
ge0ffrey: I wonder if there's a plugin that just does all this magic with a much less verbose configuration.
I looked for a patternfly plugin, but that doesn't exist (and it doesn't need to because patternfly is just a webjar).
So I looked for a webjars plugin, for which at least one plugin exists, but it's unclear if it's stable enough to use.

So, let's merge this PR and let's open a separate issue to research/investigate:
- confirm that the webjars is a consistently used standard. So that Central Maven has much many of such webjars.
- Try the use of a webjars maven plugin. Which one?

optashift-employee-rostering-gwtui/pom.xml
tiagobento: Oops. Moved.

pom.xml
```diff
@@ -65,11 +65,13 @@
     <version.org.gwtbootstrap3>0.9.4</version.org.gwtbootstrap3>
     <version.org.wildfly.wildfly-dist>10.1.0.Final</version.org.wildfly.wildfly-dist>
     <version.org.wildfly-plugins.wildfly-maven-plugin>1.2.0.Final</version.org.wildfly-plugins.wildfly-maven-plugin>
+    <version.net.ltgt.gwt.maven.gwt-maven-plugin>1.0-rc-7</version.net.ltgt.gwt.maven.gwt-maven-plugin>
     <version.net.ltgt.gwt.maven.gwt-maven-plugin>1.0-rc-9</version.net.ltgt.gwt.maven.gwt-maven-plugin>
```
ge0ffrey: duplicate version.net.ltgt.gwt.maven.gwt-maven-plugin

optashift-employee-rostering-gwtui/pom.xml
tiagobento: Yeah, it's too much configuration for just a dependency.. Good idea, let's make it a separate issue.

pom.xml
tiagobento: Sorted alphabetically in b4220c6

ge0ffrey:
Merged, but had to squash instead of rebase because the latter didn't work

tiagobento:
@ge0ffrey Glad you brought that up. Do you see any problems in "Squashing and merging"? If you "Rebase and merge" a multi-commit PR, the target branch commit history will contain every single commit of that PR and not the entire "feature" that PR added. IMO, that makes it harder to revert things and the commit history looks less clean, wdyt?

ge0ffrey:
I prefer rebasing, as then no commit history is lost, but I don't mind it that it gets squashed. It's bikeshedding I 'd say. If rebasing works, I use that, otherwise I try squashing.

Nice work on add the PatternFly webjars btw :)

tiagobento:
@ge0ffrey Great, thx :)

PR #55, opened 2018-02-01
General architectural improvements and faster page loading time

tiagobento:
On this PR:

- Separated former `MenuPanel` responsibilities
- Improved the application startup time by making pages lazy load.
- Created a very simple lifecycle mechanism to handle page changes.
- Briefly introduced Promises to handle callback chains. [1]
- Applied PatternFly styles to the Header of the application. [2]
- Created a nicer API to provide callbacks to the RestServices calls.
- Replaced old `IsElement` usages by Elemental2's `IsElement` interfaces. [3]

Known issues:

- When clicking the "Add" button on in-page forms to add data in the Skills, Spots and Employees tables, the page is refreshing.
- `TenantSelectorView` select element is not in the PatternFly style.

**Notes:**
- I recommend building this branch and running the application to best sense the changes made.
- There are some FIXMEs I left so we can better discuss what's the best approach to solve them.

--

[1] There's a much better "PromiseUtils" class available here: [Promises.java](https://github.com/tiagobento/appformer/blob/AF-933/uberfire-client-api/src/main/java/org/uberfire/client/promise/Promises.java), but since we're not using Errai's Callers, I don't think we should worry about that too much right now. Also, if we decide we could really benefit of this class, we should move it from `appformer` to `errai`, but we'd have to upgrade our Errai version. Something to keep in mind for the future, I guess.

[2] This is just a sketch of how it's going to be after we receive the definitive designs from the UX team.

[3] We should try to favor Elemental2 over GWT Widgets because they've been discontinued, while Elemental2 is under active development. If we judge we're going to make much faster progress if we use Widgets, then I think it's okay to let Elemental2's API aside.


ge0ffrey:
[3] +1
[1] Can you create an issue for moving that from uberfire to errai? As soon as it's in errai, we can upgrade our errai version. If Max approves that issue and you want to do it, I think it's a good idea. Part of the optashift work is "scratching our own back" in the upstream projects, such fixing upstream Errai issues that impact us.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/OptaShiftRosteringEntryPoint.java
```diff
@@ -23,19 +25,24 @@
...
     @PostConstruct
     public void onModuleLoad() {
-        document.getBody().appendChild(menuPanel.getElement());
-        /*GWT.setUncaughtExceptionHandler(new
-                GWT.UncaughtExceptionHandler() {
-                public void onUncaughtException(Throwable e) {
-                  ErrorPopup.show(Arrays.asList(e.getStackTrace()).stream().reduce("", (a,b) -> a.toString() + "\n" + b.toString(), (a,b) -> a + "\n" + b));
-              }
-          });*/
+        tenantStore.init(); //FIXME: Shouldn't this call be made by the Container once it's annotated with @PostConstruct?
```
mbarkley: I think the scopes of this and the `TenantStore` are inverted. It looks like the the `TenantStore` should be an `@EntryPoint` with a `@PostConstruct` that initializes itself, and this class should be an `@ApplicationScoped` bean that will be created by the container when the `TenantsReady` event fires.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/util/PromiseUtils.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.openshift.employeerostering.gwtui.client.util;
+
+import elemental2.promise.Promise;
+
+public class PromiseUtils {
```
mbarkley: I would suggest making this utility non-static and injectable so that it is more easily mocked when we start adding more unit tests.

optashift-employee-rostering-gwtui/src/main/resources/org/optaplanner/openshift/employeerostering/gwtui/client/app/AppView.html
```diff
@@ -0,0 +1,26 @@
...
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+<div class="app">
+
+    <style>
+        /* Place dev CSS here to take advantage of live reloading */
```
mbarkley: That's clever :)

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/OptaShiftRosteringEntryPoint.java
tiagobento: That makes sense. In my reasoning, when an `@EntryPoint` bean depends on a `@ApplicationScoped` bean, the `@PostConstruct` method of the `@ApplicationScoped` bean is invoked before the `@PostConstruct` of the `@EntryPoint` bean. I guess that makes sense too?

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/util/PromiseUtils.java
tiagobento: Good suggestion. I'll make this change.

tiagobento:
@ge0ffrey Thanks for the review. I'll make sure to create this issue.

@Christopher-Chianelli Thanks for the review. The "skills" page choice is only temporary, I'll certainly change it after the new designs are ready. We can also take that opportunity to make sure that the current menu button is properly highlighted when the application starts.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/util/PromiseUtils.java
tiagobento: ...in a future PR, wdyt? :P

tiagobento:
@mbarkley Thanks for the review. I didn't consider using Errai navigation. I'm actually not sure how it works, to be honest. Do you think we would benefit from using it?

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/OptaShiftRosteringEntryPoint.java
mbarkley: You need to be careful calling another bean's `@PostConstruct`. There is no special logic to ensure it's only called once, so you should only do this if your sure the method is idempotent.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/util/PromiseUtils.java
mbarkley: That's fine by me. I don't consider it a blocking issue.

mbarkley:
@tiagobento Correct me if I'm wrong, but I think your current page functionality doesn't support bookmark-able URLs and browser back/forward button navigation, right? So that's something you'd get for free from Errai Navigation. Other than that it has the same concepts of lifecycle events for pages, but with a declarative API.

[Here](http://docs.jboss.org/errai/latest/errai/reference/html_single/#sid-54493676_ErraiUINavigation-HowitWorks) is the relevant section of the Errai docs. TLDR: add `@Page` annotation on a UI component and you have a page you can navigate to by URL (with optional annotations for lifecycle methods).

mbarkley:
@tiagobento Let me know if you plan to make any changes in this PR, otherwise I can merge it for you.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/OptaShiftRosteringEntryPoint.java
tiagobento: That's valuable information. Thanks for sharing it :)

tiagobento:
@mbarkley That's great. We can certainly aim to add Errai Navigation in a near future. I think we can merge this PR.

PR #58, opened 2018-02-07
PLANNER-991: Added <!doctype html> to AppView.html to make it work in Firefox

Christopher-Chianelli:


mbarkley:
@kkufova I don't think the broken rotations link is related to this change, so I'm inclined to merge this.

@tiagobento Were you aware of this issue?

tiagobento:
@mbarkley I wasn't aware, only tested with Chrome and Safari. I guess I'll include Firefox on my list as well.

kkufova:
@mbarkley, alright, I agree, thank you for looking into this!

PR #59, opened 2018-02-07
PLANNER-992: Changed <form> to <div> to prevent forced redirect after POST

Christopher-Chianelli:


Christopher-Chianelli:
@mbarkley Those form fields do not support auto-completion (since you are inputting new data, there nothing to auto-complete). I'll keep your solution in mind though (and why was the solution not Google-able, the closest thing I got was making the target an empty iframe on Stack Overflow)...

PR #61, opened 2018-02-09
PLANNER-981: Added a health check + show a semi-readable stack trace when an uncaught exception occurs

Christopher-Chianelli:


optashift-employee-rostering-gwtui/src/main/module.gwt.xml
```diff
@@ -30,4 +30,7 @@
   <!--<set-property name="locale" value="default"/>-->
   <!--<extend-property name="locale" values="en"/>-->
   <!--<set-property-fallback name="locale" value="en"/>-->
+  <set-property name="compiler.stackMode" value="emulated" />
```
mbarkley: I'm not familiar with these settings. Do they give readable exceptions in production compiles too?

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/OptaShiftRosteringEntryPoint.java
```diff
@@ -45,4 +72,13 @@ public void onTenantsReady(final @Observes TenantStore.TenantsReady tenantsReady
...
\ No newline at end of file
+
+    private void healthCheck() {
+        TenantRestServiceBuilder.getTenantList(FailureShownRestCallback.onSuccess(tenantList -> {
+            if (null == tenantList) {
+                throw new IllegalStateException("The server cannot be contacted on url (" + Window.Location.getHref() + ").");
+            }
+            tenantStore.init(); //FIXME: Shouldn't this call be made by the Container once it's annotated with @PostConstruct?
+        }));
+    }
```
tiagobento: I'm not sure it's optimal do use this endpoint as a health-checker. Also, we'll have a duplicated call for the tenant list because `tenantStore.init()` does that too. I think it's okay to merge this but maybe adding a FIXME or a TODO marker would be a good idea.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/OptaShiftRosteringEntryPoint.java
```diff
@@ -35,7 +42,27 @@
...
+            public Throwable unwrap(Throwable e) {
+                if (e instanceof UmbrellaException) {
+                    UmbrellaException ue = (UmbrellaException) e;
+                    if (ue.getCauses().size() == 1) {
+                        return unwrap(ue.getCauses().iterator().next());
+                    }
+                }
+                return e;
+            }
+        });
```
tiagobento: Not a blocker, but I think this code block could use a method extraction.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/OptaShiftRosteringEntryPoint.java
Christopher-Chianelli: It might be a good idea to create an endpoint specificity for health checks. All other endpoints would require advance knowledge of existing tenants (this is the only one with no parameters).

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/OptaShiftRosteringEntryPoint.java
Christopher-Chianelli: Extract what? There is only one statement in `onModuleLoad`. The only thing we can do is move the implementation to another class.

optashift-employee-rostering-gwtui/src/main/module.gwt.xml
Christopher-Chianelli: From https://lists.fedorahosted.org/pipermail/rhq-devel/2011-October/001252.html, apparently yes (couldn't find any reference to in the GWT docs, other than http://www.gwtproject.org/javadoc/latest/com/google/gwt/core/server/StackTraceDeobfuscator.html).

mbarkley:
@tiagobento What is the status of this PR? I see some suggestions inline. Do you want these changes made or should I merge this as is?

tiagobento:
@mbarkley @Christopher-Chianelli It has my approval, none of my comments were blockers.

PR #64, opened 2018-02-15
PLANNER-957: Added Validation for Employee

Christopher-Chianelli:


optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/employee/EmployeeSubform.java
```diff
@@ -0,0 +1,151 @@
...
+    public void setValue(Employee value) {
+        setEmployeeModel(value);
+    }
+
+    @Override
+    public Employee getValue() {
+        return employeeModelDataBinder.getModel();
+    }
+
+    public static class SingleValueTagsInputExtension extends SingleValueTagsInput<Skill> implements HasText {
```
mbarkley: Okay, I think I understand why you use HasText here: because the supertype is already `HasValueChangeHandler<String>`, right?

IMO, I think you could simplify this (and remove the need for the external converter) by using encapsulation instead of inheritance.

Here is a sketch of a replacement:

    public class SkillValueTagsInputWrapper implements HasValue<Set<Skill>>, IsWidget {
      @Inject
      private SingleValueTagsInput<Skill> wrapped;
    
      @Override 
      public Widget asWidget() {
        return wrapped;
      }
      
      /*
       * Implement HasValue methods here, which defers to the SingleValueTagsInput change
       * handlers + conversion logic.
       * The logic could be simplified because you could set a tenant id as a field of this type.
       */
    }

This should work because `@DataField` respects `IsWidget` and will composite the object returned by `asWidget` onto the screen, but `@Bound` does not care about `IsWidget` and will respect the `HasValue` implementation of the wrapper class.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/common/SkillConvertor.java
```diff
@@ -0,0 +1,51 @@
...
+            skillNameToSkillMap.put(skill.getName(), skill);
+        }
+    }
+
+    @Override
+    public Set<Skill> toModelValue(String componentValue) {
+        String[] skillNames = componentValue.split(",");
+        Set<Skill> output = new HashSet<>();
+        for (String skillName : skillNames) {
+            if (skillNameToSkillMap.containsKey(skillName)) {
```
mbarkley: I think we can get rid of this logic entirely (see my comment on the SingleValueTagsInputExtensions), but if we do keep it we should throw an exception when converting a skill that isn't in this map.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/employee/EmployeeSubform.java
mbarkley: Okay, I spoke too soon and doing IsWidget won't work, but there are still other ways you could try to encapsulate this.

You could make a `@Templated` component that has the `SingleValueTagsInput` as it's only element. It's a bit of boiler plate (and a silly one element HTML file) but I still think it is cleaner.

mbarkley: You could also extend Composite, which is a widget type meant exactly for encapsulating and adding behaviour to a single Widget.

Christopher-Chianelli:
@mbarkley A lot of the work done here is redone/conflicts with https://github.com/kiegroup/optashift-employee-rostering/pull/65. It probably be better to review that (although I will readd validation logic). Currently seeing if I can get ListComponent to replace CellTable without needing to reimplement paging.

Christopher-Chianelli:
Closed, since this was put into https://github.com/kiegroup/optashift-employee-rostering/pull/65.

PR #71, opened 2018-03-01
PLANNER-1027:  Fixed benchmark

Christopher-Chianelli:


optashift-employee-rostering-benchmark/src/main/resources/META-INF/persistence.xml
```diff
@@ -2,16 +2,26 @@
 <!-- OVERWRITES the normal persistence.xml in the server module -->
 <persistence xmlns="http://xmlns.jcp.org/xml/ns/persistence" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
              version="2.1" xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/persistence http://xmlns.jcp.org/xml/ns/persistence/persistence_2_1.xsd">
-  <persistence-unit name="optashift-employee-rostering-persistence-unit">
+  <persistence-unit name="optashift-employee-rostering-persistence-unit" transaction-type="RESOURCE_LOCAL">
```
ge0ffrey: This resource local might be the reason you're getting empty lists.
IIRC, resource local means you're say you will explicitly open and close all transactions everywhere in your code.
So the generated testdata might not get commited to the database.

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/roster/RosterGenerator.java
```diff
@@ -93,6 +95,7 @@ public RosterGenerator() {}
      */
     public RosterGenerator(EntityManager entityManager) {
         this.entityManager = entityManager;
+        shiftRestService = new ShiftRestServiceImpl(entityManager);
```
ge0ffrey: The generator should probably not talk to any rest service, if possible.

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/shift/ShiftRestServiceImpl.java
```diff
@@ -64,6 +65,17 @@
     @Inject
     private EmployeeRestService employeeRestService;
 
+    @SuppressWarnings("unused")
+    public ShiftRestServiceImpl() {
+    }
+
+    // For benchmark
+    public ShiftRestServiceImpl(EntityManager entityManager) {
```
ge0ffrey: I 'd rather see the Generator refactor so it doesn't need the ShiftRestService to begin with.

optashift-employee-rostering-benchmark/src/main/resources/META-INF/persistence.xml
Christopher-Chianelli: Changing it to JTA causes a NPE when building the entityManager (no other info is given).

rsynek: @Christopher-Chianelli for JTA transactions, I would suggest configuring a Datasource on the container - I can't see any used in your persistence.xml

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/roster/RosterGenerator.java
```diff
@@ -91,8 +90,9 @@ public RosterGenerator() {
      * For benchmark only
      * @param entityManager never null
      */
-    public RosterGenerator(EntityManager entityManager) {
+    public RosterGenerator(EntityManager entityManager, ShiftFileParser shiftGenerator) {
```
ge0ffrey: What is ShiftFileParser? What does it do.

The variable name called generator but the type is FileParser - that's confusing.

Christopher-Chianelli: ShiftFileParser probably should be renamed to `ShiftGenerator`. It converts `ShiftTemplate` to `Shift`.

Christopher-Chianelli: (FileParser part is from when it parses files in the October 2017 version)

ge0ffrey: Now that ShiftTemplate with have a createShift(start, end) method, is Shiftgenerator/ShiftTemplate still needed?

yurloc:
Are you sure this is the fix? Why is test coverage missing? How is `OptaShiftEmployeeRosteringBenchmarkApp` supposed to be used? It still fails if I run
```
mvn -nsu exec:java -Dexec.mainClass=org.optaplanner.openshift.employeerostering.benchmark.OptaShiftEmployeeRosteringBenchmarkApp
```

yurloc:
I'm sorry I messed up the local build for the first time. Now the benchmark runs OK.

Only the missing test coverage remark is still valid. If the benchmark breaks in the future we won't notice it without at least a basic test.

PR #74, opened 2018-03-05
PLANNER-1019 Create a PoC of REST API tests for Skill endpoint

rsynek:
This is a proposal of OptaShift Employee Rostering REST API testing. The PR contains:
- introduction of maven Cargo as a tool for bootstrapping container and deploying the app to it
- removal of Arquillian used be preceding test
- test for Skill REST endpoint
- smaller fix to an issue found by tests

It's fair to say that although using Cargo simplifies dependency tree, using Arquillian could still bring one benefit - test execution time in case we do in-container testing. Although this way a single test may be 10 times faster, we still need to test the REST API itself from a client, so the difference won't be so big. Also all the logic of the application is inside the REST endpoint classes, there is no layer behind them to focus on.


rsynek:
@Christopher-Chianelli question is - do we have some use for the `equals` and `hashcode` based on ID? If not, I would recommend reimplementing them in all entities according to https://stackoverflow.com/questions/1638723/how-should-equals-and-hashcode-be-implemented-when-using-jpa-and-hibernate
"Basing equals() / hashCode() on a business key - e.g. a unique combination of attributes that is not going to change during object lifetime".

In general this is a result of disobeying a separation of concerns principle - the entity is used for JPA and also for serialization into JSON for REST interface and in UI too. So the best approach would be to use different set of classes for each of this purpose.

optashift-employee-rostering-client/src/main/java/org/optaplanner/openshift/employeerostering/client/ServiceClientFactory.java
```diff
@@ -0,0 +1,67 @@
...
+import org.optaplanner.openshift.employeerostering.shared.employee.EmployeeRestService;
+import org.optaplanner.openshift.employeerostering.shared.roster.RosterRestService;
+import org.optaplanner.openshift.employeerostering.shared.shift.ShiftRestService;
+import org.optaplanner.openshift.employeerostering.shared.skill.SkillRestService;
+import org.optaplanner.openshift.employeerostering.shared.spot.SpotRestService;
+import org.optaplanner.openshift.employeerostering.shared.tenant.TenantRestService;
+
+/**
+ * Creates client proxies to remote REST API endpoints.
+ * */
```
ge0ffrey: **/

optashift-employee-rostering-client/src/main/java/org/optaplanner/openshift/employeerostering/client/ServiceClientFactory.java
```diff
@@ -0,0 +1,67 @@
...
+public final class ServiceClientFactory {
+
+    private static final String REST_API_CONTEXT = "rest";
+
+    private URL baseUrl;
+    private ResteasyClient resteasyClient;
+    private ResteasyWebTarget target;
+
+    /**
+     * Constructor.
```
ge0ffrey: Let's remove this line. Javadocs should only include info that the method signature doesn't already say.

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/skill/SkillRestServiceImpl.java
```diff
@@ -43,6 +44,9 @@
     @Transactional
     public Skill getSkill(Integer tenantId, Long id) {
         Skill skill = entityManager.find(Skill.class, id);
+        if (skill == null) {
+            throw new EntityNotFoundException("No Skill entity found with ID " + id);
```
ge0ffrey: Use general exception style: "... with id (" + id + ").");

optashift-employee-rostering-webapp/pom.xml
```diff
@@ -14,6 +14,20 @@
 
   <name>Employee Rostering as a Service Webapp</name>
 
+  <properties>
+    <wildfly.version>11.0.0.Final</wildfly.version>
```
ge0ffrey: All version properties should be declared in the parent pom.

optashift-employee-rostering-webapp/src/test/java/org/optaplanner/openshift/employeerostering/webapp/skill/AbstractRestServiceIT.java
```diff
@@ -0,0 +1,56 @@
...
+import org.optaplanner.openshift.employeerostering.client.ServiceClientFactory;
+import org.optaplanner.openshift.employeerostering.webapp.tools.ClientResponseContextAssert;
+import org.optaplanner.openshift.employeerostering.webapp.tools.RecordingClientResponseFilter;
+import org.optaplanner.openshift.employeerostering.webapp.tools.TestConfig;
+
+public class AbstractRestServiceIT {
+
+    private static final String BASE_TEST_URL = TestConfig.getApplicationUrl();
+
+    //Consequence of reusing JPA entities on client - cannot rely on equals()
```
ge0ffrey: space after //

optashift-employee-rostering-webapp/src/test/java/org/optaplanner/openshift/employeerostering/webapp/skill/AbstractRestServiceIT.java
```diff
@@ -0,0 +1,56 @@
...
+
+    protected final ServiceClientFactory serviceClientFactory;
+    protected final RecordingClientResponseFilter recordingClientResponseFilter;
+
+    protected AbstractRestServiceIT() {
+        URL baseTestUrl;
+        try {
+            baseTestUrl = new URL(BASE_TEST_URL);
+        } catch (MalformedURLException e) {
+            throw new IllegalArgumentException("Base URL is invalid.", e);
```
ge0ffrey: throw new IllegalArgumentException("The baseTestUrl (" + baseTestUrl + ") is invalid.", e);

optashift-employee-rostering-webapp/src/test/java/org/optaplanner/openshift/employeerostering/webapp/skill/SkillRestServiceIT.java
```diff
@@ -0,0 +1,82 @@
...
+import org.optaplanner.openshift.employeerostering.shared.skill.Skill;
+import org.optaplanner.openshift.employeerostering.shared.skill.SkillRestService;
+
+import static org.assertj.core.api.Assertions.*;
+
+import java.util.List;
+
+public class SkillRestServiceIT extends AbstractRestServiceIT {
+
+    private static final int TENANT_ID = new Random().nextInt(Integer.MAX_VALUE);
```
ge0ffrey: The random is unseeded.
This is a randomly failing test that fails if the nextInt is the same as a tenant id in the auto generated testdata.
Just hard corde TENANT_ID -1 or -1000 instead.

optashift-employee-rostering-webapp/src/test/java/org/optaplanner/openshift/employeerostering/webapp/skill/SkillRestServiceIT.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Test
+    public void testUpdateNonExistingSkill() {
+        final long nonExistingSkillId = 123456L;
+        Skill nonExistingSkill = new Skill(TENANT_ID, "Non-existing skill");
+        nonExistingSkill.setId(nonExistingSkillId);
+        Skill updatedSkill = skillRestService.updateSkill(TENANT_ID, nonExistingSkill);
+
+        assertThat(updatedSkill.getName()).isEqualTo(nonExistingSkill.getName());
```
ge0ffrey: JUnit trumps assertJ when in direct competition :)

ge0ffrey:
Looks good. Very nice work. I made a few suggestions - the only one that really has to go is "new Random()" which creates randomly failing tests.

optashift-employee-rostering-webapp/pom.xml
rsynek: Thanks, will fix

optashift-employee-rostering-webapp/src/test/java/org/optaplanner/openshift/employeerostering/webapp/skill/SkillRestServiceIT.java
rsynek: The test will not necessarily fail in case it jumps into tenant ID that already has some data associated with - the @Before method clears the Skill data in such a case. On the other hand I agree the random here is needless complication, will remove it.

optashift-employee-rostering-webapp/src/test/java/org/optaplanner/openshift/employeerostering/webapp/skill/AbstractRestServiceIT.java
rsynek: Ok

rsynek:
@ge0ffrey thank you for your comments, all suggestions have been worked in.

@Christopher-Chianelli please let me know about your opinion on the equals/hashcode. If we are going to change it, let's start with this PR - I am happy to do it for Skill entity and corresponding tests.

Christopher-Chianelli:
@rsynek There is only one case where ID is important: Shifts; We currently represent a spot having two Shifts at the same time as two identical Shifts with different ID's. So you have my permission to change `equals` and `hashCode` on everything BUT `Shift`. Currently updating the architecture (https://issues.jboss.org/browse/PLANNER-1037).

ge0ffrey:
@rsynek Great work, thanks :)

PR #76, opened 2018-03-06
PLANNER-1038: Removed Old Canvas UI for Roster Screens

Christopher-Chianelli:


Christopher-Chianelli:
TODO: Remove binary files and find out why target is not in the `.gitignore`

PR #79, opened 2018-03-07
PLANNER-1033: Allow you to sort rows by any column in Spot

Christopher-Chianelli:


ge0ffrey:
Nice work, we'll want these on the other 2 CRUD screens as well :)

PR #75, opened 2018-03-06
PLANNER-1037: Refactored Model

Christopher-Chianelli:
Just posting so you can see progress

PR #84, opened 2018-03-12
PLANNER-1051: Added Employee Roster Screen

Christopher-Chianelli:


optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/common/Observable.java
```diff
@@ -0,0 +1,27 @@
+package org.optaplanner.openshift.employeerostering.gwtui.client.common;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import org.optaplanner.openshift.employeerostering.gwtui.client.interfaces.Updatable;
+
+public class Observable<T> {
```
ge0ffrey: On one hand, I'd expect errai to have something like this already
On the other hand, I wonder if the abstraction is worth it (it's not like `listeners.forEach((l) -> l.onUpdate(newValue))` is a lot of code).

@tiagobento / @mbarkley wdyt?

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/common/TimeSlotUtils.java
```diff
@@ -7,6 +7,10 @@
 
 public class TimeSlotUtils {
 
+    public static LocalDate toLocalDate(OffsetDateTime time) {
+        return LocalDate.of(time.getYear(), time.getMonthValue(), time.getDayOfMonth());
```
ge0ffrey: (hard) Use OffsetDateTime.toLocalDate()

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/common/TimeSlotUtils.java
```diff
@@ -7,6 +7,10 @@
...
+    }
+
     public static LocalTime toLocalTime(OffsetDateTime time) {
```
ge0ffrey: delete this one too (see above)

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/common/Observable.java
Christopher-Chianelli: In fact, `java.util` has this exactly, but for some reason it not in GWT Java emulation libary. Might be able to simulate this with events (`@Observes Event<X> e` and `@Inject Event<X>`

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/common/TimeSlotUtils.java
Christopher-Chianelli: I would. but our emulated libary does not have ` OffsetDateTime.toLocalDate()` (causes a GWT compile error)

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/common/TimeSlotUtils.java
Christopher-Chianelli: See above (we should probably update our emulated library, it is very out of date)

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/roster/RosterRestServiceImpl.java
```diff
@@ -193,8 +190,10 @@ protected EmployeeRosterView getEmployeeRosterView(Integer tenantId, LocalDate s
                 .getResultList();
         Map<Long, List<ShiftView>> employeeIdToShiftViewListMap = new LinkedHashMap<>(employeeList.size());
         for (Shift shift : shiftList) {
-            employeeIdToShiftViewListMap.computeIfAbsent(shift.getEmployee().getId(), k -> new ArrayList<>()).add(
-                    new ShiftView(shift));
+            if (shift.getEmployee() != null) {
```
ge0ffrey: (important to think about) Shouldn't the EmployeeRosterView know the shifts of the employee "unassigned" (so employee = null)? Even if the UI doesn't use it, it's interesting to include in the REST api.
Remember that the end-user might hire external temp-nurses to fill in the unassigned shifts, so he'll want to know how many there are.

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/roster/RosterGenerator.java
```diff
@@ -432,6 +433,44 @@ private Integer createTenant(int spotListSize, int employeeListSize, int lengthI
...
+            employeesListCopy.removeAll(unavailableEmployees);
+            List<Employee> undesiredEmployees = new ArrayList<>(extractRandomSubList(employeesListCopy, 0.3));
+            employeesListCopy.removeAll(undesiredEmployees);
+            List<Employee> desiredEmployees = new ArrayList<>(extractRandomSubList(employeesListCopy, 0.3));
+            employeesListCopy.removeAll(desiredEmployees);
+
+            unavailableEmployees.forEach((e) -> {
+                EmployeeAvailability toAdd = new EmployeeAvailability(tenantId, e, date, OffsetTime.of(LocalTime.MIN, config.getTimeZone().getRules().getOffset(date.atStartOfDay())), OffsetTime.of(LocalTime.MAX, config
+                        .getTimeZone().getRules().getOffset(date.atTime(LocalTime.MAX))));
+                toAdd.setState(EmployeeAvailabilityState.UNAVAILABLE);
```
ge0ffrey: I am not seeing any randomness in here.
Ideally we want 80% stability and 20% randomness - just like in real life.

Let's merge this and talk about improving the generated data as a separate issue.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/common/Observable.java
Christopher-Chianelli: Did a refactor to use ErraiMessage bus (a weakly typed equivalent, register SpotListProvider and the like as "Services" and have subscriber subscribe to their events)

tiagobento:
@Christopher-Chianelli Looks like `spotroster/ShiftBlob` had its label overwritten by #75. See: https://github.com/kiegroup/optashift-employee-rostering/pull/75/files#diff-1f3758159fa4b36f21ddd7d7259b4927R85. Could you please revert that?

I'm still in the process of reviewing it but I think that the EmployeeRoster should have employees as its lanes and different blob colors depending on the chosen availability. Idk if you guys discussed otherwise, though. @ge0ffrey could you please verify that?

I noticed that this PR has the PLANNER-1051 label but there's no description provided in the ticket or on the PR. It's a good practice to give some context of what problems you're solving or what features you're adding on a PR to make it easier to review. For example, I don't understand why the need to add the `*ListProvider` classes to the `*ListPanel` classes, could you please explain that?

Christopher-Chianelli:
@tiagobento in regards to your `*ListProvider` classes, it is because I noticed we are using the Employee List/Spot List/Skill List in multiple places, and we are making unneeded REST calls (since if we kept them all in one place, we don't need to a REST call to fetch the list (provided it already been fetched). This PR is more of a discussion PR: it provides a basis to a discussion (in this case, how should the Employee Roster be implemented). It also includes some refactoring (namely: creating a central repository for the Employee/Spot/Skill Lists so we do not need to do multiple REST calls)

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/pages/employeeroster/AvailabilityLane.java
```diff
@@ -0,0 +1,41 @@
...
+package org.optaplanner.openshift.employeerostering.gwtui.client.pages.employeeroster;
+
+import java.time.OffsetDateTime;
+import java.util.List;
+
+import org.optaplanner.openshift.employeerostering.gwtui.client.rostergrid.model.Lane;
+import org.optaplanner.openshift.employeerostering.gwtui.client.rostergrid.model.SubLane;
+import org.optaplanner.openshift.employeerostering.shared.employee.EmployeeAvailabilityState;
+
+public class AvailabilityLane extends Lane<OffsetDateTime> {
```
tiagobento: I think that there should be one employee per lane perhaps?

tiagobento:
@Christopher-Chianelli Ok. I don't think this extra layer of complexity pays off when compared to making many repeated REST calls.

If our REST calls are fast enough, we can simplify our codebase by a lot. Having this async caching mechanism is a case of _Premature Optimization_, which, [according to some](http://wiki.c2.com/?PrematureOptimization), is "the root of all evil" 😛 

In the end, I think we will never want to fetch the entire list of skills/employees/spots anyways, because they are known to have the potential to be very big. For the list screens, we will probably have paginated REST calls, so no need to cache anything. For the BlobPopoverContents, we will probably fetch the top 10 results each time the user types something on the "Employee" input, so also no need to cache.

From my local tests, your implementation works very well, but for the reasons mentioned above I'll kindly ask you to remove this mechanism since I don't think the application will benefit from it right now.

Regarding the use of this PR as a "discussion proposal", I think code can be a very powerful way to express ideas, but the risk of having to re-work it is big, so we have to take that into consideration. For this particular case, I think we and the rest of the team should've exchanged some comments on the JIRA or even in an email thread, so the implementation (which is often the most expensive part of the process) is much more accurate when it comes together.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/pages/spotroster/SpotRosterViewportFactory.java
```diff
@@ -75,45 +78,49 @@
...
+                    cssGridLinesFactory.newWithSteps(2L, 12L),
+                    ticksFactory.newTicks(scale, 2L, 12L),
+                    lanes);
         });
     }
 
     private Map<Spot, List<ShiftView>> buildSpotRosterModel(final SpotRosterView spotRosterView) {
 
-        final Map<Long, Spot> spotsById = indexById(spotRosterView.getSpotList());
+        final Map<Long, Spot> spotsById = spotListProvider.getUnmodifiableSpotList().stream().collect(Collectors.toMap((e) -> e.getId(), (e) -> e));
```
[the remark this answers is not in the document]
Christopher-Chianelli: Will change (provided it does the same thing) (although should we also cache this so we don't need to rebuild it everytime?)

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/pages/employeeroster/AvailabilityLane.java
Christopher-Chianelli: Can you explain an advantage of having Employee per lane versus Availability per lane? This is something a manager will be using, and I don't think they are interested in when a particular employee is off, but rather what employees are off. Plus if it was per employee, we will probably have a lot of empty lanes, making it harder to find out who is off.

Christopher-Chianelli:
@tiagobento One thing I worry about frequent REST calls is what if the user has a slow connection. I do my testing on localhost, so REST calls are almost instantaneous for me. And how frequent will these REST calls be? If they will be happening every 10s or so, perhaps having a WebSocket API in addition to our REST API will be a good idea (we still need REST for outside applications using this application). I will remove the providers, but a rework of the API is something to consider. Implementation in this case was not very expensive, since you already did all the work for me. I consider this a prototype: something that is extremely cheap to implement that give you a good idea of what it will be when it is implemented. And like all prototypes, I am willing to throw it to the can if you can suggest something else.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/pages/spotroster/SpotRosterViewportFactory.java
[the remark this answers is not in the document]
tiagobento: If we detect that's a bottle neck for the performance of the application, absolutely! :)

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/pages/employeeroster/AvailabilityLane.java
tiagobento: I'm not sure about the advantages, I just thought we were rewriting the old version of the Employee Roster using the HTML/CSS grid. If the UX team, @mbarkley and @ge0ffrey agree, we can certainly keep the Availability lanes. Did we already discussed that with them?

Christopher-Chianelli: It a discussion to be had; I can easily change it back to Employee Lanes; need to do a meetup on this.

mbarkley: Not sure I even understand what "Availability per lane" even means. So far in all our discussions, we have talked about the "Employee Roster" screen being a view of the Spot Roster data but organized by employee. IMO we should stick with that for now, and discuss alternatives later.

ge0ffrey:
The only way to know what happens on slow connections is to emulate it locally (Data is better than opinions). After starting up the server, configure linux to add a delay to a TCP connections to the HTTP port used by our app. Then see what happens. A delay of 500ms for every HTTP request is a good start - it will flush out "n+1 query problems", especially when looking at the tenants with more data.

Notice there are 2 properties:
- latency (delay): how long does it take for a HTTP request to reach the other side. 
- bandwidth: how much does the size of the HTTP request impact the time for the last bit to reach the other side.
Don't worry about bandwidth - no need to emulate it reducing it for now. Worry about latency - and emulate it.

Christopher-Chianelli:
@tiagobento @mbarkley @ge0ffrey Implemented suggested changes, uses Employee lanes now to simulate the old Canvas UI (for better or for worst is to be discussed).

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/header/MenuView.java
```diff
@@ -35,10 +35,11 @@
 import org.optaplanner.openshift.employeerostering.gwtui.client.pages.Pages;
 
 import static org.optaplanner.openshift.employeerostering.gwtui.client.pages.Pages.Id.EMPLOYEES;
+import static org.optaplanner.openshift.employeerostering.gwtui.client.pages.Pages.Id.EMPLOYEE_ROSTER;
 import static org.optaplanner.openshift.employeerostering.gwtui.client.pages.Pages.Id.ROTATIONS;
 import static org.optaplanner.openshift.employeerostering.gwtui.client.pages.Pages.Id.SKILLS;
 import static org.optaplanner.openshift.employeerostering.gwtui.client.pages.Pages.Id.SPOTS;
-import static org.optaplanner.openshift.employeerostering.gwtui.client.pages.Pages.Id.SPOT_ROSTER;
+import static org.optaplanner.openshift.employeerostering.gwtui.client.pages.Pages.Id.SPOT_ROSTER;;
```
ge0ffrey: Double ;

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/pages/Pages.java
```diff
@@ -58,10 +61,13 @@
...
     @Inject
-    private ManagedInstance<SpotRosterPage> spotRosterDemoPage;
+    private ManagedInstance<SpotRosterPage> spotRosterPage;
 
     @Inject
-    private ManagedInstance<RotationPage> rotationsDemoPage;
+    private ManagedInstance<EmployeeRosterPage> employeeRosterPage;
+
+    @Inject
+    private ManagedInstance<RotationPage> rotationsPage;
```
ge0ffrey: rename to singular: "rotationPage"

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/pages/employeeroster/EmployeeAvailabilityBlob.java
```diff
@@ -0,0 +1,104 @@
...
+public class EmployeeAvailabilityBlob implements Blob<OffsetDateTime> {
+
+    private final LinearScale<OffsetDateTime> scale;
+    private EmployeeAvailability availability;
+    private Long sizeInGridPixels;
+
+    private Long positionInGridPixelsCache = null;
+    private Long endPositionInGridPixelsCache = null;
+
+    EmployeeAvailabilityBlob(final LinearScale<OffsetDateTime> scale, final EmployeeAvailability availability) {
```
ge0ffrey: public or protected constructor? (the default - friendly - we don't use)

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/pages/employeeroster/EmployeeRosterPage.java
```diff
@@ -0,0 +1,191 @@
...
+    //Events
+    @EventHandler("refresh-button")
+    public void onRefreshButtonClicked(@ForEvent("click") final MouseEvent e) {
+        refreshWithLoadingSpinner();
+    }
+
+    @EventHandler("previous-page-button")
+    public void onPreviousPageButtonClicked(@ForEvent("click") final MouseEvent e) {
+
+        /*if (spotsPagination.isOnFirstPage()) {
```
ge0ffrey: spots?

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/pages/employeeroster/EmployeeRosterPage.java
```diff
@@ -0,0 +1,191 @@
...
+            RosterRestServiceBuilder.getCurrentEmployeeRosterView(tenantStore.getCurrentTenantId(),
+                    onSuccess(v -> {
+                        currentEmployeeRosterView = v;
+                        resolve.onInvoke(v);
+                    }));
+        });
+    }
+
+    public EmployeeRosterView getCurrentEmployeeRosterView() {
+        // TODO Auto-generated method stub
```
ge0ffrey: remove comment

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/pages/employeeroster/EmployeeRosterPage.java
```diff
@@ -0,0 +1,191 @@
...
+    public void onForwardInTimeButtonClicked(@ForEvent("click") final MouseEvent e) {
+        viewportFrame.scrollLeft += TIME_SCROLL_SIZE;
+    }
+
+    @EventHandler("back-in-time-button")
+    public void onBackInTimeButtonClicked(@ForEvent("click") final MouseEvent e) {
+        viewportFrame.scrollLeft -= TIME_SCROLL_SIZE;
+    }
+
+    //API calls
```
ge0ffrey: I'd remove this comment - Seems obvious from the use of "..RestService..."

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/pages/employeeroster/Positive2HoursScale.java
```diff
@@ -0,0 +1,82 @@
...
+    private final OffsetDateTime end;
+
+    public Positive2HoursScale(final OffsetDateTime start, final OffsetDateTime end) {
+        this.start = start;
+        this.end = end;
+    }
+
+    @Override
+    public Long toGridPixelsWithFactor1(final OffsetDateTime valueInScaleUnits) {
+        final OffsetDateTime date;
```
ge0ffrey: a variable called "date" that is a DateTime?

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/pages/spotroster/SpotRosterPage.java
```diff
@@ -186,6 +191,10 @@ public void onTenantChanged(@Observes final TenantStore.TenantChange tenant) {
         return refreshWithoutLoadingSpinner().then(i -> {
             loadingSpinner.hideFor("refresh-spot-roster");
             return resolve();
+        }).catch_(e -> {
+            ErrorPopup.show(e.toString());
```
ge0ffrey: Now this is a piece of code that can use a comment :)
No code should every call ErrorPopup.show() directly, because that's being handled as "an aspect" (which is a fancy way of saying that there's global exception handler that makes sure the pop up is shown).
But I bet there's a good reason to do it manually here  (the loadingSpinner.hide maybe?). Add a comment to explain that good reason.

optashift-employee-rostering-gwtui/src/main/resources/org/optaplanner/openshift/employeerostering/gwtui/client/pages/employeeroster/EmployeeRosterPage.css
```diff
@@ -0,0 +1,493 @@
...
+}
+
+/* VIEWPORT */
+
+.viewport-frame {
+    user-select: none;
+    overflow: scroll;
+    position: relative;
+    padding: 5px;
+}
```
ge0ffrey: This entire css file seems very long - does it duplicate anything from the spot roster css?
If so, can we pull that into a more global css file?

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/roster/RosterGenerator.java
```diff
@@ -432,6 +438,50 @@ private Integer createTenant(int spotListSize, int employeeListSize, int lengthI
...
+        for (LocalDate date : datesBetween) {
+            List<Employee> employeesListCopy = new ArrayList<>(employeeList);
+            List<Employee> unavailableEmployees = new ArrayList<>(extractRandomSubList(employeesListCopy, 0.3));
+            employeesListCopy.removeAll(unavailableEmployees);
+            List<Employee> undesiredEmployees = new ArrayList<>(extractRandomSubList(employeesListCopy, 0.3));
+            employeesListCopy.removeAll(undesiredEmployees);
+            List<Employee> desiredEmployees = new ArrayList<>(extractRandomSubList(employeesListCopy, 0.3));
+            employeesListCopy.removeAll(desiredEmployees);
+
+            unavailableEmployees.forEach((e) -> {
```
ge0ffrey: For consistency, we use a simple old-fashioned "for ( Foo f : foos)" over "foo.forEach()", unless there's a good reason to use forEach() (such as the reuse of a method).

pom.xml
```diff
@@ -259,6 +259,59 @@
...
+            <configuration>
+              <sources>
+                <source>${project.build.directory}/generated-sources/annotations/</source>
+              </sources>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+      
+      <!-- TODO: Remove me when Eclipse recongize generated sources -->
```
ge0ffrey: Are you using the latest Eclipse version?

ge0ffrey: typo "recognize"

Anyway, this sucks. Eclipse could handle maven's generated sources in 2008 so then the created these messy workarounds. It's 2018 now. I can't believe Eclipse hasn't fixed this. I am sure they have - it's probably a setting somewhere that says "do the right thing with maven files" that is defaulted to "false".
Let's not pollute our build with these old workarounds - let's look for today's real solution with eclipse.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/pages/spotroster/SpotRosterPage.java
Christopher-Chianelli: It appears the global exception handler does not work with promises, so to actually show the error, it need to be explicitly catched. Plus, we need to hide the loading spinner; I add that as a comment.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/pages/employeeroster/EmployeeRosterPage.java
Christopher-Chianelli: It to separate method types, to make it easier to find what method does a particular thing

PR #87, opened 2018-03-14
PLANNER-1055: Added visual indication of Historic, Published and Draft shifts

Christopher-Chianelli:
Historic shifts are not visible in the current implementation (as there are none currently), but would look have no outline
Pinning a shift cause a pin to appear next to the shift text
Published shifts have a solid line, draft shifts have a dotted line

### DIFF
--- a/optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/pages/spotroster/SpotRosterPage.java
+++ b/optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/pages/spotroster/SpotRosterPage.java
@@ -20,6 +20,7 @@ import java.time.OffsetDateTime;
 import java.util.Optional;
 
 import javax.annotation.PostConstruct;
+import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.event.Observes;
 import javax.inject.Inject;
 import javax.inject.Named;
@@ -50,9 +51,9 @@ import static elemental2.dom.DomGlobal.setTimeout;
 import static java.lang.Math.max;
 import static org.optaplanner.openshift.employeerostering.gwtui.client.common.FailureShownRestCallback.onSuccess;
 import static org.optaplanner.openshift.employeerostering.gwtui.client.util.PromiseUtils.resolve;
-import static org.optaplanner.openshift.employeerostering.shared.roster.RosterRestServiceBuilder.getCurrentSpotRosterView;
 
 @Templated
+@ApplicationScoped
 public class SpotRosterPage implements Page {
 
     private static final Integer SOLVE_TIME_IN_SECONDS = 30;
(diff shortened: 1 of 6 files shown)


optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/pages/spotroster/SpotRosterPage.java
```diff
@@ -102,7 +103,7 @@
 
     @Inject
     @DataField("previous-page-button")
-    private HTMLAnchorElement previousPageButton;
+    private HTMLAnchorElement cuurentSpotRosterViewpreviousPageButton;
```
Christopher-Chianelli: Will fix (is suppose to be RosterState currentSpotRosterView;, somehow got merged with the previous variable and still compile and work fine)

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/roster/RosterState.java
```diff
@@ -103,21 +103,20 @@ public boolean isHistoric(Shift shift) {
         return shift.getEndDateTime().isBefore(OffsetDateTime.of(getFirstPublishedDate().atTime(LocalTime.MIDNIGHT), shift.getEndDateTime().getOffset()));
```
Christopher-Chianelli: Should a shift that is in progress be historic?

ge0ffrey: Good question. For now, let's always use the startDateTime to decide if a shift is historic/published/draft. But once I see some end-users of this, I'll ask whether or not that makes sense that shifts that are in progress are considered historic in visualization.
I'd argue: yes, because for non-disruptive replanning of published shifts (!= our current "solve" button) shifts that have already started can't change, so those are historic.

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/roster/RosterRestServiceImpl.java
```diff
@@ -274,4 +276,9 @@ public void updateShiftsOfRoster(Roster newRoster) {
                     ? null : employeeIdMap.get(shift.getEmployee().getId()));
         }
     }
+
+    // TODO: Make this a REST method?
```
ge0ffrey: Yes, once we build the roster state configuration screen (of which most fields will be read-only)

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/roster/RosterState.java
```diff
@@ -103,21 +103,20 @@ public boolean isHistoric(Shift shift) {
...
     }
 
-    @JsonIgnore
+    // Do we need this, since if a shift exists, it would be draft...
+    /*@JsonIgnore
```
ge0ffrey: Agreed, remove the method isUnplanned

PR #89, opened 2018-03-15
Fix loose ends of WildFly update and Cargo replacement of arquillian

ge0ffrey:
- Remove arquillian-bom dependencies etc
- Use the cargo skip test enviromnent variable
- Skip cargo test in profile openshift (doing that on pod builds takes many resources)
- Update wildfly used template for openshift also to version 11 (like the rest of the wildfly upgrade)


optashift-employee-rostering-gwtui/pom.xml
```diff
@@ -110,10 +110,6 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-simple</artifactId>
     </dependency>
-    <dependency>
-      <groupId>org.jboss.errai</groupId>
-      <artifactId>errai-validation</artifactId>
-    </dependency>
```
Christopher-Chianelli: I think we need this; we use validators on client side (try inputting an emptying name for a Skill on SkillListPanel)

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/resources/i18n/OptaShiftUIConstants.java
```diff
@@ -5,7 +5,7 @@
 public interface OptaShiftUIConstants {
 
     @TranslationKey(
-            defaultValue = "The server returned an empty result which is impossible.\nMaybe the server is using the wrong WildFly version. Try using WildFly 10.1.0.Final.")
+            defaultValue = "The server returned an empty result which is impossible.\nMaybe the server is using the wrong WildFly version.")
```
Christopher-Chianelli: Should mention a working WildFly Version (if not here somewhere else where it is easily accessible)

optashift-employee-rostering-webapp/pom.xml
```diff
@@ -225,28 +225,8 @@
         </property>
```
Christopher-Chianelli: Should we add another activation here for cargo.maven.skip?

Christopher-Chianelli: (It just above this line, but GitHub won't let me comment on it)

optashift-employee-rostering-gwtui/pom.xml
ge0ffrey: Yes, we do, but we don't need it twice :) The dependency is also declared about 10 lines above it.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/resources/i18n/OptaShiftUIConstants.java
ge0ffrey: Problem is that this message was already out of date.
Either we inject it somewhere (think filtered resources based on the maven property version.org.wildfly) or we don't mention it. Otherwise it's unreliable (wildfly gets updated often and it's always forgotten).

optashift-employee-rostering-webapp/pom.xml
ge0ffrey: The maven-cargo-plugin directly supports the environment variable "cargo.maven.skip". There's no need to add an activation for that.

The goal if this code is to make sure that "-DskipTests" also skips cargo. Ideally, the maven-cargo-plugin should also pick up the "skipTests", but it doesn't.
There's no need to do any activation when a user only wants to skip cargo.

PR #91, opened 2018-03-15
PLANNER-1061: Added the ability to publish in the Spot Roster

Christopher-Chianelli:


optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/tenant/TenantRestServiceImpl.java
```diff
@@ -55,14 +55,15 @@ public Tenant addTenant(Tenant tenant) {
...
-                                            tenantConfiguration.getTenantId() + ").");
-        }
-        tenant.setConfiguration(tenantConfiguration);
-        return entityManager.merge(tenant);
+    public TenantConfiguration updateTenantConfiguration(TenantConfiguration tenantConfiguration) {
+        return entityManager.merge(tenantConfiguration);
+    }
+
+    @Override
+    public TenantConfiguration getTenantConfiguration(Integer tenantId) {
```
ge0ffrey: method order, to make it easier to read/maintain and see if it's consistenty:
get before add, add before update, update before delete.

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/roster/RosterRestService.java
```diff
@@ -30,6 +30,14 @@
 @GenRestBuilder
 public interface RosterRestService {
 
+    // ************************************************************************
+    // RosterState
+    // ************************************************************************
+    @ApiOperation("Fetches the current Roster State")
```
ge0ffrey: All the other ApiOperations use the wording "get".
Even in the swagger docs, it's nice to use consistent wording, so when a different word is used, we signal we mean there's some sort of difference with the other methods.

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/roster/RosterRestService.java
```diff
@@ -97,9 +105,20 @@ EmployeeRosterView getEmployeeRosterViewFor(@ApiParam(required = true) @PathPara
     @Path("/terminate")
     void terminateRosterEarly(@ApiParam(required = true) @PathParam("tenantId") Integer tenantId);
 
+    @ApiOperation("Publishes the next set of draft shifts and creates new draft shift from the rotation template.")
```
ge0ffrey: Use active tense, just like the other docs, so "publish" instead of "Publishes" and "create" instead of "creates".
There's a general docs/javadocs convention to prefer active tense whenever possible and to avoid past or future tense.

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/tenant/TenantRestService.java
```diff
@@ -55,11 +55,18 @@ Tenant addTenant(
     // ************************************************************************
     // TenantConfiguration
     // ************************************************************************
+    /**
+     * @param id never null
+     * @return never null, the id
+     */
```
ge0ffrey: Consistency: either we javadoc all methods on all RestService interfaces, or we don't. "Do or do not" ;) Hybrid states hurt long term quality.

Given that we swaggerdoc them all - and that's what really matters - I wouldn't bother: simply remove this javadoc.

PR #92, opened 2018-03-16
PLANNER-1058 mvn -DwithoutGwt profile + rename client to restclient

ge0ffrey:
- Renamed optashift-employee-rostering-client to optashift-employee-rostering-restclient
- Added maven profile to not build nor include the GWT stuff. Activate it with -DwithoutGwt. Needed for middleware production deployments.
- Cleaned up some pom files
- Fix cargoWildfly profile so it is activated by default, even if other (non-cargo) profiles are activated.
- Removed the use of scope in dependencyManagement.


ge0ffrey:
@rsynek because I broke it - in the intellij refactor rename from client to restclient I also had it scan for strings/comments - and I apparently forgot to exclude one of those false negatives.

PR #93, opened 2018-03-16
PLANNER-1062: Created a default error handler for promises

Christopher-Chianelli:


tiagobento:
Like @mbarkley suggested once, I think making the `PromiseUtils` class be injected instead of statically used will make our code more prepared to be tested in the future. Since this PR is very straightforward I think we could use the opportunity to change that, wdyt @Christopher-Chianelli?

Christopher-Chianelli:
@tiagobento Good idea, could also do the same with CommonUtils

Christopher-Chianelli:
@tiagobento @ge0ffrey Made CommonUtils and PromiseUtils be injected

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/util/PromiseUtils.java
```diff
@@ -16,11 +16,53 @@
...
         return Promise.resolve((Promise<Void>) null);
     }
+
+    public <T> Promise<T> promise(PromiseExecutorCallbackFn<T> executor) {
+        return manage(new Promise<>(executor));
+    }
+
+    public <T> Promise<T> manage(Promise<T> promise) {
+        return promise.catch_(this.<T> getDefaultCatch());
+    }
```
tiagobento: I know i kinda suggested it but now I'm starting to think that "manage" is not a good name. We're creating a promise with a fixed catch statement which always returns a resolved promise. Maybe it's not a good idea to do that. We can hand the exception over to GWT but maybe have it reject anyways.. wdyt @Christopher-Chianelli?

Christopher-Chianelli: I tried rejecting it like `return new Promise<>((res,rej) -> rej.onInvoke(e));`, but that created an infinite loop on `getDefaultCatch()`. Maybe `attempt` for the new name?

ge0ffrey:
@Christopher-Chianelli Can we give CommonsUtil a better name?

Christopher-Chianelli:
@ge0ffrey `CommonUtils`is for common operations. It has several methods: `delimitCollection()`, which converts a collection into a delimited string, `flatten()`, which makes an `Iterable<T>` from `Interable<Iterable<T>>` and `stringWithIntCompareTo()`, which is a lexlogical comparator, except for strings `aX` `aY` where a is a string and X and Y are different numbers, in which case It will compare base off of X and Y numerical values. I guess it can be called `CollectionUtils`

ge0ffrey:
@mbarkley @tiagobento Is this ok to merge now?

(This is also part of https://github.com/kiegroup/optashift-employee-rostering/pull/95 I believe.)

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/util/PromiseUtils.java
```diff
@@ -16,11 +16,60 @@
...
+                GWT.getUncaughtExceptionHandler().onUncaughtException((Throwable) e);
+            } else if (e instanceof Error) {
+                Error error = (Error) e;
+                GWT.getUncaughtExceptionHandler().onUncaughtException(new Throwable(error.toString()));
+            } else {
+                if (null != e) {
+                    JsObject error = new JsObject(e);
+                    // Exceptions seems to have a reference to the actual exception stored in "__java$exception"
+                    // Hopefully this is also true in production mode, else most stack traces will be useless
+                    if (error.hasOwnProperty("__java$exception")) {
```
mbarkley: Errai has a [utility class](https://github.com/errai/errai/blob/master/errai-common/src/main/java/org/jboss/errai/common/client/logging/util/StackTraceFormatter.java) that does this for the SLF4J logger in a way that gives translated stack traces in Google Chrome (provided source maps are available).

You might just want to use an SLF4J logger here and log the error so that Errai prints it to the console.

Christopher-Chianelli: https://github.com/kiegroup/optashift-employee-rostering/blob/2ebc8b5964a4eef7bb38598d917e4939af3b9d5a/optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/OptaShiftRosteringEntryPoint.java#L46-L66 does the stack trace translating. Unfortunately the error is usually not an exception object except if we explicitly throw the exception (which is usually never). Ideally, we would not need this method, but since Promises are native js, it seems GWT does not use `getUncaughtExceptionHandler` to report the exception. It does appear I might be able to simplify this logic: the stack that cause an exception is visible from the catch method, so I can just turn this to `GWT.getUncaughtExceptionHandler().onUncaughtException(new Throwable(e.toString()))`.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/util/PromiseUtils.java
tiagobento: I think it's a good option.

Christopher-Chianelli:
Was merged with #95

PR #97, opened 2018-03-20
provision.sh improvements

ge0ffrey:
- Prefer running the script with #!/bin/bash instead of #!/bin/sh so it finds "oc" and "mvn" more easily (if they are declared in .bashrc)
- remove old github url references
- remove "demo" and straighten out naming

### DIFF
--- a/provision.sh
+++ b/provision.sh
@@ -1,10 +1,9 @@
 #!/bin/bash
-#!/bin/sh
 set -e
 
 command -v oc >/dev/null 2>&1 || {
-  echo >&2 "The oc client tools need to be installed to connect to OpenShift Online.";
-  echo >&2 "Download it from https://www.openshift.org/download.html and confirm that \"oc version\" runs.";
+  echo >&2 "ERROR: The oc client tools need to be installed to connect to OpenShift Online.";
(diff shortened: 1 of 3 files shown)


provision.sh
```diff
@@ -1,5 +1,5 @@
-#!/bin/sh
 #!/bin/bash
+#!/bin/sh
```
Christopher-Chianelli: https://stackoverflow.com/a/42966709 states that there can only be one shebang, so it will attempt to launch bash, and if it is not available, report it cannot find bash. If you want it to attempt bash, then default to sh, you need to create a minor program to send the file to the correct shell, and make that your shebang. it would look like this:
```
#!/bin/sh
if which /bin/bash
then
/bin/bash $1
else
/bin/sh $1
fi

ge0ffrey: That code doesn't work, but you're correct that there can only be one shebang, so this needs to change.

We enforce the bash for all our scripts in 
  https://github.com/kiegroup/droolsjbpm-build-bootstrap/tree/master/script
for about 7 years now and we haven't received any complaints about that.
So, let's just be consistent with those scripts and only use "bash". (And if someone submits a PR for a good working "if" approach, we'll apply it there too).

PR #99, opened 2018-03-23
PLANNER-1047: Auto-focus on first input element when editing/creating a Skill/Employee/Spot

Christopher-Chianelli:
You can switch between input elements using tab, type into the select box to find and select skills, and click on save/cancel/whatever using enter


tiagobento:
@Christopher-Chianelli Although this seems a concise solution, depending on the HTML hierarchical structure is rarely a good design choice. I think you could either create an abstract method on `TableRow<T>` and force its implementations to define the element to focus there or call `.focus()` on each implementation's @PostConstruct method.

Note that on the first approach you're making `TableRow<T>` always have an element to be focused on when editing. On the second approach you make `TableRow<T>` more flexible, but you have to write a similar code thrice.

Christopher-Chianelli:
@tiagobento probably going with your first approach; the `@PostConstruct` one won't work since that would be called for ALL rows, including ones that are not being edited.

Christopher-Chianelli:
@tiagobento Fixed; I made the implementations do the actual focusing, so they don't need to convert their inputs to HTMLElement

PR #100, opened 2018-03-23
PLANNER-955: Added Heat-Map to Spot Roster

Christopher-Chianelli:


optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/solver/WannabeSolverManager.java
```diff
@@ -49,6 +56,8 @@
 
     private ConcurrentMap<Integer, SolverStatus> tenantIdToSolverStateMap = new ConcurrentHashMap<>();
     private ConcurrentMap<Integer, Solver<Roster>> tenantIdToSolverMap = new ConcurrentHashMap<>();
+    private ConcurrentMap<Integer, Map<Shift, IndictmentView>> tenantIdToShiftIndictmentMap = new ConcurrentHashMap<>();
+    private ConcurrentMap<Integer, Map<Employee, IndictmentView>> tenantIdToEmployeeIndictmentMap = new ConcurrentHashMap<>();
```
ge0ffrey: Race condition alert.
These are 4 different fields that on their own are thread-safe, but together they might not be.

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/solver/WannabeSolverManager.java
```diff
@@ -108,4 +115,40 @@ public Roster getRoster(final Integer tenantId) {
         Solver<Roster> solver = tenantIdToSolverMap.get(tenantId);
         return solver == null ? null : solver.getBestSolution();
     }
+
+    public Map<Shift, IndictmentView> getShiftIndictmentMap(Integer tenantId) {
```
ge0ffrey: SolverManager can't return "IndictmentView", because the later isn't optaplanner-api and SolverManager will be pushed upstream into optaplanner-core someday (in some form, it might be ripped apart).
Extract the logic to create IndictmentViews into the RosterService.

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/solver/WannabeSolverManager.java
```diff
@@ -108,4 +115,40 @@ public Roster getRoster(final Integer tenantId) {
...
+        if (tenantIdToSolverMap.containsKey(tenantId)) {
```
ge0ffrey: race condition

```
if (tenantIdToSolverMap.containsKey(tenantId)) {
  
  foo = tenantIdToSolverMap.get(tenantId);
  ...
}
```

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/solver/WannabeSolverManager.java
```diff
@@ -108,4 +115,40 @@ public Roster getRoster(final Integer tenantId) {
...
+            Roster roster = tenantIdToSolverMap.get(tenantId).getBestSolution();
+            ScoreDirector<Roster> scoreDirector = tenantIdToSolverMap.get(tenantId).getScoreDirectorFactory()
```
ge0ffrey: Called tenantIdToSolverMap.get(tenantId) twice

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/solver/WannabeSolverManager.java
```diff
@@ -108,4 +115,40 @@ public Roster getRoster(final Integer tenantId) {
...
+            Map<Shift, IndictmentView> indictmentMap = tenantIdToShiftIndictmentMap.computeIfAbsent(tenantId, (k) -> new HashMap<>());
+            indictmentMap.clear();
+            scoreDirector.getIndictmentMap().forEach((k, v) -> {
+                if (k instanceof Shift) {
+                    indictmentMap.put((Shift) k, new IndictmentView(v));
+                }
+            });
+            return indictmentMap;
+        }
+        return tenantIdToShiftIndictmentMap.getOrDefault(tenantId, Collections.emptyMap());
```
ge0ffrey: So, if we can't find it, there we no indictments?
This is "let's sweep the problem under the rug" code. The UI then incorrectly reports that the schedule is perfect (while it is not).

If it can't find it, it must fail fast. It must stick out like soar thumb. And if that's an ugly red popup in the UI, that's a good thing for us. Much better than an "all is fine" look with invalid data. Then we either fix it properly or we figure out a way to notify the user in the UI in a pretty way (think notification) that we don't know.

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/roster/view/IndictmentView.java
```diff
@@ -0,0 +1,87 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.optaplanner.core.api.score.buildin.hardsoft.HardSoftScore;
+import org.optaplanner.core.api.score.constraint.ConstraintMatch;
+import org.optaplanner.core.api.score.constraint.Indictment;
+import org.optaplanner.persistence.jackson.api.score.ScoreJacksonJsonSerializer;
+import org.optaplanner.persistence.jackson.api.score.buildin.hardsoft.HardSoftScoreJacksonJsonDeserializer;
+
+public class IndictmentView {
+
+    private List<Rules> ruleMatchedList;
```
ge0ffrey: Rename "Rules" to "Constraints". Adjust all fields/methods accordingly, of course.

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/roster/view/IndictmentView.java
```diff
@@ -0,0 +1,87 @@
...
+                case "No 2 shifts within 10 hours from each other":
+                    return EMPLOYEE_HAS_SHIFTS_LESS_THAN_1O_HOURS_APART;
+                case "Undesired time slot for an employee":
+                    return UNDESIRED_TIMESLOT_FOR_EMPLOYEE;
+                case "Desired time slot for an employee":
+                    return DESIRED_TIMESLOT_FOR_EMPLOYEE;
+                case "Employee is not rotation employee":
+                    return EMPLOYEE_IS_NOT_ROTATION_EMPLOYEE;
+                default:
+                    return UNKNOWN;
```
ge0ffrey: Fail fast.

throw IllegalStateException, remove UNKNOWN.
copy the "throw new IllegalStateException(...)" from one of the other switches I already wrote (here or in optaplanner-core), so the exception message is consistent.

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/roster/view/IndictmentView.java
```diff
@@ -0,0 +1,87 @@
...
+    public void setRuleMatchedList(List<Rules> ruleList) {
+        this.ruleMatchedList = ruleList;
+    }
+
+    public void setScoreList(List<HardSoftScore> scoreList) {
+        this.scoreList = scoreList;
+    }
+
+    public static enum Rules {
+        MISSING_REQUIRED_SKILL,
```
ge0ffrey: An enum can have a field, for example:

public enum Constraint {
   MISSING_REQUIRED_SKILL("Required skill for a shift"),
   EMPLOYEE_IS_NOT_ROTATION_EMPLOYEE(...);

  private final String name;
  private Constraint(String name) {
    this.name = name;
  }

  public static Constraint findFromConstraintMatch(ConstraintMatch constraintMatch) {
    for (Constraint constraint : Constraint.values() {
       if (constraint.name.equals(constraintMatch.getConstraintName()) {
           return constraint;
       }
   }

}

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/roster/view/IndictmentView.java
```diff
@@ -0,0 +1,87 @@
...
+        UNDESIRED_TIMESLOT_FOR_EMPLOYEE,
+        DESIRED_TIMESLOT_FOR_EMPLOYEE,
+        UNAVALIABLE_TIMESLOT_FOR_EMPLOYEE,
+        EMPLOYEE_IS_NOT_ROTATION_EMPLOYEE,
+        UNKNOWN;
+
+        public static Rules forConstraintMatch(ConstraintMatch cm) {
+            switch (cm.getConstraintName()) {
+                case "Required skill for a shift":
+                    return MISSING_REQUIRED_SKILL;
```
ge0ffrey: The constraint name (= the string in the drl after the word rule) must match the enum constraint name exactly, except for camel case.

So either change the drl to "Missing required skill" or change the enum value to "REQUIRED_SKILL_FOR_A_SHIFT". Take whatever you'd put on the label in the english i18n in the UI.

The English UI labels must match the code field/property names, always. If you change the UI label, you must change the field/property names too. Otherwise you're creating technical debt, which is cancer for a project.

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/roster/view/IndictmentView.java
```diff
@@ -0,0 +1,87 @@
...
+        EMPLOYEE_ALREADY_HAS_A_SHIFT_ON_DATE,
+        EMPLOYEE_HAS_SHIFTS_LESS_THAN_1O_HOURS_APART,
+        UNDESIRED_TIMESLOT_FOR_EMPLOYEE,
+        DESIRED_TIMESLOT_FOR_EMPLOYEE,
+        UNAVALIABLE_TIMESLOT_FOR_EMPLOYEE,
+        EMPLOYEE_IS_NOT_ROTATION_EMPLOYEE,
+        UNKNOWN;
+
+        public static Rules forConstraintMatch(ConstraintMatch cm) {
+            switch (cm.getConstraintName()) {
```
ge0ffrey: In theory, matching the constraintName isn't enough. A constraintMatchTotal is unique by the combination of constraintPackage and constraintName.
Ideally, with the code trick above, each Constraint enum has a 2 String fields: String constraintPackage and String constraintName. Both are being checked to be equal in that for loop.

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/roster/view/IndictmentView.java
```diff
@@ -0,0 +1,87 @@
...
+        this.ruleMatchedList = ruleMatched;
+        this.scoreList = score;
+    }
+
+    public List<Rules> getRuleMatchedList() {
+        return ruleMatchedList;
+    }
+
+    @JsonSerialize(contentUsing = ScoreJacksonJsonSerializer.class)
+    @JsonDeserialize(contentUsing = HardSoftScoreJacksonJsonDeserializer.class)
```
ge0ffrey: This can now be removed

Christopher-Chianelli: Not yet, requires https://github.com/kiegroup/optashift-employee-rostering/pull/101

Christopher-Chianelli:
@ge0ffrey Fixed; I also changed the version property to OffsetDateTime, which allows us to get the last modification date of any of the tables (so we can detect if any of the maps are out-dated). Unfortunately, there isn't really a clean way to pass an exception to the client from the server when we detect outdated data, so I made the "currentIndictmentMap" return null if the data is outdated, and "latestIndictmentMap" get the calculated indictment map (regardless if the data is outdated). Still need to put an alert on client side. Should all server-side code be taken to be concurrent?

ge0ffrey:
jenkins retest this

optashift-employee-rostering-restclient/pom.xml
```diff
@@ -29,7 +29,7 @@
     </dependency>
     <dependency>
       <groupId>org.jboss.resteasy</groupId>
-      <artifactId>resteasy-jackson-provider</artifactId>
+      <artifactId>resteasy-jackson2-provider</artifactId>
```
ge0ffrey: The main pom also references "resteasy-jackson-provider", please change it there too.

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/common/AbstractRestServiceImpl.java
```diff
@@ -12,9 +11,13 @@
...
     protected void validateTenantIdParameter(Integer tenantId, AbstractPersistable persistable) {
         if (!Objects.equals(persistable.getTenantId(), tenantId)) {
-            throw new IllegalStateException("The tenantId (" + tenantId
-                    + ") does not match the persistable (" + persistable + ")'s tenantId (" + persistable.getTenantId() + ").");
+            throw new IllegalStateException("The tenantId (" + tenantId + ") does not match the persistable (" + persistable + ")'s tenantId (" + persistable.getTenantId() + ").");
         }
     }
 
+    protected OffsetDateTime getLastUpdateDateTime(Integer tenantId, EntityManager entityManager, Class<? extends AbstractPersistable> entityClass) {
+        final String queryString = "SELECT MAX(lastUpdateDateTime) FROM " + entityClass.getSimpleName() + " row WHERE row.tenantId = " + tenantId;
```
ge0ffrey: Security problem, this is vulnerable to JPA-QL or SQL injection hacking.
Use setParameter("entityClass", entityClass.getSimpleName()) instead

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/roster/view/IndictmentView.java
```diff
@@ -0,0 +1,100 @@
...
+import org.optaplanner.core.api.score.buildin.hardsoft.HardSoftScore;
+import org.optaplanner.core.api.score.constraint.ConstraintMatch;
+import org.optaplanner.core.api.score.constraint.Indictment;
+import org.optaplanner.persistence.jackson.api.score.ScoreJacksonJsonSerializer;
+import org.optaplanner.persistence.jackson.api.score.buildin.hardsoft.HardSoftScoreJacksonJsonDeserializer;
+
+public class IndictmentView {
+
+    private List<Constraint> constraintMatchedList;
+    private List<HardSoftScore> scoreList;
```
ge0ffrey: A list of scores? If this scoreList is the same length as constraintMatchedList, it should be one list of ConstraintMatchView's

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/roster/view/IndictmentView.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public void setConstraintMatchedList(List<Constraint> constraintMatchedList) {
+        this.constraintMatchedList = constraintMatchedList;
+    }
+
+    public void setScoreList(List<HardSoftScore> scoreList) {
+        this.scoreList = scoreList;
+    }
+
+    public static enum Constraint {
```
ge0ffrey: Let's rename this to RosterConstraintType and put it as a normal class (not an inner class) in this shared.roster.view package. I suspect we'll need to this more later, once we do the Parameterization feature.

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/common/AbstractPersistable.java
```diff
@@ -39,11 +41,10 @@
     protected Integer tenantId;
 
     @Version
-    protected Long version;
```
ge0ffrey: This needs to stay, we're using it for optimistic locking.

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/common/AbstractPersistable.java
```diff
@@ -39,11 +41,10 @@
...
+    protected OffsetDateTime lastUpdateDateTime;
```
ge0ffrey: I 'd rather not see this added.

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/roster/RosterRestServiceImpl.java
```diff
@@ -67,6 +77,11 @@
     @Inject
     private ShiftRestService shiftRestService;
 
+    private Map<Integer, Map<Shift, IndictmentView>> tenantIdToShiftIndictmentMap = new ConcurrentHashMap<>();
+    private Map<Integer, Map<Employee, IndictmentView>> tenantIdToEmployeeIndictmentMap = new ConcurrentHashMap<>();
+    private Map<Integer, OffsetDateTime> tenantIdToShiftIndictmentMapUpdateTime = new ConcurrentHashMap<>();
+    private Map<Integer, OffsetDateTime> tenantIdToEmployeeIndictmentMapUpdateTime = new ConcurrentHashMap<>();
```
ge0ffrey: If they need to exist, they need to existing in the SolverManager, not in the RosterRestService.

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/roster/RosterRestServiceImpl.java
```diff
@@ -359,4 +376,146 @@ public RosterState getRosterState(Integer tenantId) {
...
+                    }
+                });
+                tenantIdToEmployeeIndictmentMap.put(tenantId, indictmentMap);
+                tenantIdToEmployeeIndictmentMapUpdateTime.put(tenantId, OffsetDateTime.now());
+                return indictmentMap;
+            }
+        } else {
+            OffsetDateTime lastUpdateTime = getLastUpdateTime(tenantId);
+            OffsetDateTime lastMapUpdateTime = tenantIdToEmployeeIndictmentMapUpdateTime.getOrDefault(tenantId, OffsetDateTime.MIN);
+            if (lastUpdateTime.isBefore(lastMapUpdateTime)) {
```
ge0ffrey: use the optimistic version number instead.

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/roster/RosterRestServiceImpl.java
ge0ffrey: I'll work through this. Fix the other comments in the PR and let me know - I 'll work on top of it to design RosterRestServiceImpl and WannabeSolverManager to how I see it. Also rebase this PR to master if you haven't already.

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/common/AbstractRestServiceImpl.java
ge0ffrey: https://imgs.xkcd.com/comics/exploits_of_a_mom.png

Christopher-Chianelli: This is not exposed to client side; to perform the exploit you would have to already hacked the server, but I changed it to native query with parameter set

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/common/AbstractPersistable.java
Christopher-Chianelli: `@Version` can be of type `OffsetDateTime`, and can still be used for optimistic locking (changed by hibernate on insert and update operation)

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/roster/RosterRestServiceImpl.java
Christopher-Chianelli: is it guaranteed that ALL future versions are greater than ALL past versions (I am guessing that since when Version is OffsetDateTime, version is a timestamp, it is still a timestamp when it is a long)?

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/common/AbstractRestServiceImpl.java
Christopher-Chianelli: Actually impossible: https://stackoverflow.com/a/31159912; I need to escape it when building the string...

optashift-employee-rostering-restclient/pom.xml
Christopher-Chianelli: Done

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/common/AbstractRestServiceImpl.java
Christopher-Chianelli: I check if the name is a valid table name, if not, I throw an IllegalStateException. Unfortunately, JPA does not provide a utility for escaping table names/verifying table names, so I have to create one from its spec

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/roster/view/IndictmentView.java
Christopher-Chianelli: Done

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/roster/view/IndictmentView.java
Christopher-Chianelli: Done

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/common/AbstractPersistable.java
Christopher-Chianelli: Done

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/common/AbstractPersistable.java
Christopher-Chianelli: Done

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/roster/RosterRestServiceImpl.java
Christopher-Chianelli: I use "[count(id), sum(version)]" as a hash

ge0ffrey:
jenkins retest this

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/pages/spotroster/ShiftBlobView.java
```diff
@@ -153,4 +204,32 @@ public void remove() {
     public Blob<OffsetDateTime> getBlob() {
         return blob;
     }
+
+    private void setBlobColorAccordingToScore(HardSoftScore score) {
+        if (score.getHardScore() != 0) {
```
ge0ffrey: hard score can be positive in the future if we ever add hard rewards, so != 0 is a bit risky.
It's safer to use !score.isFeasible() (which checks if hardsScore >= 0)

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/common/AbstractRestServiceImpl.java
```diff
@@ -12,9 +12,98 @@
 
     protected void validateTenantIdParameter(Integer tenantId, AbstractPersistable persistable) {
         if (!Objects.equals(persistable.getTenantId(), tenantId)) {
-            throw new IllegalStateException("The tenantId (" + tenantId
-                    + ") does not match the persistable (" + persistable + ")'s tenantId (" + persistable.getTenantId() + ").");
+            throw new IllegalStateException("The tenantId (" + tenantId + ") does not match the persistable (" + persistable + ")'s tenantId (" + persistable.getTenantId() + ").");
         }
     }
 
+    protected List<Long> getTableCountAndVersionSum(Integer tenantId, EntityManager entityManager, Class<? extends AbstractPersistable> entityClass) {
```
ge0ffrey: Instead of the hash approach, we'll look into the optimistic version number approach in our next meeting.

Christopher-Chianelli:
Doing a rewrite in another PR

PR #101, opened 2018-03-26
PLANNER-1071: Remove @JsonFormat for time fields and instead set the default formatter of time fields.

Christopher-Chianelli:


optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/roster/view/AbstractRosterView.java
```diff
@@ -84,7 +81,7 @@ public void setEmployeeList(List<Employee> employeeList) {
         this.employeeList = employeeList;
     }
 
-    @JsonSerialize(using = ScoreJacksonJsonSerializer.class)
+    @JsonSerialize(using = HardSoftScoreJacksonJsonSerializer.class)
```
ge0ffrey: +1

But if you ObjectMapper.register the new OptaPlannerJacksonModule.createModule() (see 7.7.0 optaplanner docs), you can get rid of these entirely.
Do check if the client still works then.

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/common/jaxrs/OptaShiftObjectMapperResolver.java
```diff
@@ -34,7 +31,10 @@
 
     public OptaShiftObjectMapperResolver() {
         objectMapper = new ObjectMapper()
-                .registerModule(new JavaTimeModule());
+                .registerModule(new JavaTimeModule())
+                .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
+                // This converts the server side offset to GMT, is that what we want?
+                .setDateFormat(new ISO8601DateFormat());
```
ge0ffrey: No, for example for OffsetDateTime, we want to have the same as it's toString(), which is an ISO standard that includes the offset. For example: "2007-12-03T10:15:30+01:00"

From OffsetDateTime's javadoc:

```
    /**
     * Outputs this date-time as a {@code String}, such as {@code 2007-12-03T10:15:30+01:00}.
     * <p>
     * The output will be one of the following ISO-8601 formats:
     * <ul>
     * <li>{@code uuuu-MM-dd'T'HH:mmXXXXX}</li>
     * <li>{@code uuuu-MM-dd'T'HH:mm:ssXXXXX}</li>
     * <li>{@code uuuu-MM-dd'T'HH:mm:ss.SSSXXXXX}</li>
     * <li>{@code uuuu-MM-dd'T'HH:mm:ss.SSSSSSXXXXX}</li>
     * <li>{@code uuuu-MM-dd'T'HH:mm:ss.SSSSSSSSSXXXXX}</li>
     * </ul>
     * The format used will be the shortest that outputs the full value of
     * the time where the omitted parts are implied to be zero.
     *
     * @return a string representation of this date-time, not null
     */
    @Override
    public String toString() {
        return dateTime.toString() + offset.toString();
    }
```
The abbreviation would be nice to have, but not required. The offset inclusion is required (otherwise there is data loss).
I suspect the jackson-java8Time module docs will document the best way to achieve this.

Christopher-Chianelli:
@ge0ffrey TimeZone offset is included in the output now. Unfortunately I cannot write a test for it since Cargo strongly dislike ShiftRestService.

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/common/jaxrs/OptaShiftObjectMapperResolver.java
```diff
@@ -16,25 +16,28 @@
...
 
     public OptaShiftObjectMapperResolver() {
+        // JavaTimeModule defaults date format to ISO, so we don't need to deal with the headache of setting
+        // a universal date format that is ISO and includes an offset
         objectMapper = new ObjectMapper()
-                .registerModule(new JavaTimeModule());
+                .registerModule(OptaPlannerJacksonModule.createModule())
+                .registerModule(new JavaTimeModule())
+                .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
+                .disable(DeserializationFeature.ADJUST_DATES_TO_CONTEXT_TIME_ZONE);
```
ge0ffrey: Let's add a comment above the first disable() call, such as:

// Write OffsetDateTime's (and similar) to JSON in the same format as OffsetDateTime.toString(), for example: 2017-01-01...(copy paste one from a json out)

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/common/jaxrs/OptaShiftObjectMapperResolver.java
```diff
@@ -16,25 +16,28 @@
...
+import org.optaplanner.persistence.jackson.api.OptaPlannerJacksonModule;
 
 @Provider
 public class OptaShiftObjectMapperResolver implements ContextResolver<ObjectMapper> {
 
     private final ObjectMapper objectMapper;
 
     public OptaShiftObjectMapperResolver() {
+        // JavaTimeModule defaults date format to ISO, so we don't need to deal with the headache of setting
+        // a universal date format that is ISO and includes an offset
```
ge0ffrey: This comment is confusing to me, let's remove it in favor of the other comment I am suggesting below.
The general idea of referring to the toString() is "I thrust the java.time creators to have chosen the best standarized OffsetDateTime format for it's toString().". On top of that, it adds an output example to clarify what we expect to see.

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/common/jaxrs/OptaShiftObjectMapperResolver.java
Christopher-Chianelli: JavaTimeModule includes an OffsetDateTime serializer, which does not use toString()

Christopher-Chianelli:
@ge0ffrey Updated the comment; it doesn't refer to `toString()` since it doesn't use `toString()` (see https://github.com/FasterXML/jackson-datatype-jsr310/blob/master/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/OffsetDateTimeSerializer.java)

Christopher-Chianelli:
jenkins please retest this

PR #102, opened 2018-03-29
PLANNER-1074: Correctly handle shifts that wrap around a rotation

Christopher-Chianelli:


optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/roster/RosterGenerator.java
```diff
@@ -277,24 +276,21 @@ public GeneratorType(String tenantNamePrefix, StringDataGenerator skillNameGener
...
-            ),
+                    Pair.of(LocalTime.of(19, 0), LocalTime.of(7, 0))),
             // Day:   A A A B B B B C C C A A A A B B B C C C C
             // Night: C C C A A A A B B B C C C C A A A B B B B
             21, 3, (startDayOffset, timeslotRangesIndex) -> {
                 int offset = timeslotRangesIndex == 0 ? startDayOffset : (startDayOffset + 14) % 21;
-                return offset < 3 ? 0 : offset < 7 ? 1 : offset < 10 ? 3 :
-                        offset < 14 ? 0 : offset < 17 ? 1 : offset < 21 ? 3 : -1;
-    });
+                return offset < 3 ? 0 : offset < 7 ? 1 : offset < 10 ? 3 : offset < 14 ? 0 : offset < 17 ? 1 : offset < 21 ? 3 : -1;
```
ge0ffrey: eclipse's policy to remove newlines that were specifically added for readability is a pain.
Is there any way to turn this off? Intellij doesn't do this, it respects existing newlines.

Christopher-Chianelli: I think it would be better to enforce formatting via maven, so we wouldn't need to worry about differences between editors. There is probably an option somewhere, but why does it not match Intellij formatter...

Christopher-Chianelli:
jenkins please retest this

PR #103, opened 2018-04-06
Use grid positioning instead of pixel positioning

Christopher-Chianelli:
This a major refactoring in how the UI works. It removes absolute positing of elements and instead uses the CSS Grid Layout. Quite a bit of logic need to be changed:

-jQuery draggable and resizable don't like grid, so needed to use mouse handlers to implement these features in the rotation screen

-Due to how grid positioning works, the html structure of viewport contains lane contains sublane contains blobs has been lost. Instead, it now viewport contains lane and sublane and blob. The html structure can be readded, but then we will have a grid inside a grid inside a grid and positioning logic change alot.

-Due to inability to rely on html structure to position the element, ListView keeps a reference to the model of the container, so that child elements can use it to position themselves.

-CSS grid line now places elements instead of using a custom css style, so changes to the grid do not require changes to the style

-SubLane keep a reference to the title of the parent lane so it can find its position within the parent lane, might be able to remove it...

The reason for this refactor is using absolute positioning is bad in general, and this allow the app to be responsive to screen size. Additionally, it allows us to build things as an area in a grid, which is helpful. Additionally, dates are now displayed as [date](time1 time2 time3 ...) where date is formated like "Wed Nov 3, 2018" and time is formatted like "8 PM". 

Added a utility that will probably be removed; PageUtils, which allow you to set the height of an element to the remainder of the page after removing all "height eating" elements `calc(100vh - eatenHeight - padding)`. Gonna try to use the window as the viewport frame instead (which removes the need of the utility).


optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/rostergrid/model/Orientation.java
```diff
@@ -16,48 +16,192 @@
...
+        }
+        return value;
+    }
+
+    abstract void absGroupPosition(final IsElement element, final Long positionInGridPixels, final Viewport viewport, final Long offsetInScreenPixels);
+
+    abstract void groupPosition(final IsElement element, final Long positionInGridPixels, final Viewport viewport, final Long offsetInScreenPixels);
+
+    // Use this for headers
+    abstract void absPosition(final IsElement element, final Long positionInGridPixels, final Viewport viewport, final Long offsetInScreenPixels);
```
mbarkley: Great comment. Maybe make this a javadoc so it shows up in an IDE when someone is trying to figure out which method to call?

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/rostergrid/model/Orientation.java
```diff
@@ -16,48 +16,192 @@
...
+
+    // Use this for everything else (ensures it doesn't conflict with header rows)
```
mbarkley: Same suggestion here as above.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/rostergrid/view/SubLaneView.java
```diff
@@ -81,34 +85,23 @@ public void onClick(final @ForEvent("click") MouseEvent e) {
...
-                lanes.remove(lane);
-            }
-        }
-
         // Add Blob (ALT + CLICK)
-        else if (e.altKey) {
-            final double offset = viewport.decideBasedOnOrientation(e.offsetY, e.offsetX);
-            final Long positionInGridPixels = viewport.toGridPixels(new Double(offset).longValue());
-            final T positionInScaleUnits = viewport.getScale().toScaleUnits(positionInGridPixels);
+        //if (e.altKey) {
```
mbarkley: Please remove commented out code.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/rostergrid/view/ViewportView.java
```diff
@@ -39,21 +46,101 @@
...
     }
+
+    @EventHandler("viewport")
+    private void onDragStart(@ForEvent("mousedown") MouseEvent e) {
+        if (!inMouseEvent && viewport != null && viewport.getMouseTarget() != null) {
+            try {
+                inMouseEvent = true;
+                e.target = viewport.getMouseTarget().getElement();
+                MouseEvent copy = new MouseEvent(e.type, (MouseEventInit) e);
+                viewport.getMouseTarget().getElement().dispatchEvent(copy);
```
mbarkley: Why do you have to manually dispatch a copy of the event? Shouldn't the target receive the event already?

optashift-employee-rostering-gwtui/src/main/resources/org/optaplanner/openshift/employeerostering/gwtui/client/app/AppView.less
```diff
@@ -71,7 +71,7 @@
 }
 
 .footer {
-  margin-top: -40px;
+  /* margin-top: -40px; */
```
mbarkley: Please remove commented out code.

optashift-employee-rostering-gwtui/src/main/resources/org/optaplanner/openshift/employeerostering/gwtui/public/roster.less
```diff
@@ -11,7 +11,7 @@
 
 .viewport-frame {
     user-select: none;
-    overflow: scroll;
+    /* overflow: scroll; */
```
mbarkley: More commented out code.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/rostergrid/view/ViewportView.java
Christopher-Chianelli: Not always; consider the case you drag the mouse fast enough so that it is on the viewport but not the element, or if you are dragging the edge and you haven't move the mouse enough to increase the length.

Christopher-Chianelli: @mbarkley Correct; viewport.setMouseTarget(element) is called on element mouseDown event handler and viewport.setMouseTarget(null) is called on element mouseUp handler.

tiagobento:
After reading the code this PR adds, I'm inclined not to approve it due to the refactor to use HTML-grid instead of absolute positioning.

The reasons are as follows:

_tl;dr_ I don't think HTML-grid is meant for a highly dynamic grid, which happens to be exactly what we have.

1. I failed to find any benefits of HTML-grid compared to absolute positioning. The PR description reads "this allow the app to be responsive to screen size" but I don't see how the current implementation can't. Also, I personally don't think that "using absolute positioning is bad in general" since we got a working mechanism that does essentially the same with much less code.
2. This PR adds very complex methods and no tests. We don't have a testing infrastructure for this kind of thing yet but I don't think that now is the time to think about having one it either. So the option is to keep complexity low to ensure sanity during development.
3. JQuery UI is used to make dragging and resizing work. As stated on the PR description, it doesn't work with HTML-grid, so this PRs introduces an in-house solution for what jQuery UI does and puts more complex code on our shoulders to maintain, which I think is bad on the medium/long term.
4. The vertical grid lines displayed are currently pure CSS and no elements on the DOM, which is already crowded with Blobs. By making these grid lines be `div`s we are slowing down the application and introducing the need to deal with more stuff on the DOM hierarchy.
5. Using HTML-grid makes we lose flexibility regarding Blob positioning. By using absolute positioning we can have custom-length Shifts like "6:20 to 15:20" without making it step 20 to 20 minutes while dragging/resizing. This is not possible with HTML-grid or even if it were, it would require us to make changes on the in-house drag/resize code. This is particularly important because UX might have big plans for us in the future, so I don't see why committing to a change that kills these possibilities while we can opt not to do so.
6. We lost Spots pagination along the way.
7. This PR is an overkill to what we wanted to achieve: better UI/UX on the roster screens.

Certainly this PR contains valuable changes that I really like, but unfortunately, they're tightly coupled with this major HTML-grid refactor.

I understand that timing is delicate here because we're 2 days away from a code-freeze and these changes are really important to be in the release. My suggestion is we merge this and all the other related PRs on `7.7.x` (which will be used to generate the upcoming release) and move on with `master` without these merges. We will have to reimplement these changes on `master` based on the current absolute positioning mechanism, but for the reasons stated above, I believe this is what we should do.

Of course, it's @ge0ffrey's decision how we're going to proceed.

Please feel free to add your comments @Christopher-Chianelli @mbarkley.

Christopher-Chianelli:
@tiagobento 
1. It decouples styling code from Java code, so if a designer wants to change the width of a blob, they can just do it in CSS without modifying the scales. Additionally, it allows us to "zoom-in" by adjusting grid-column-width/grid-row-height without modifying the position of every single blob.

2. Agreed, but the entirety of client side needs tests.

3. Agree-ish; I like JQuery since it means less code to maintain, but it doesn't automatically change the mouse cursor on edges/show a shift is movable.

4. That is easily revertible; I changed it so code is more maintainable (so if we change grid width, we don't need to update the code/css styling). I did a test and performance was the same between the two implementations.

5. Do we want to support this (that is, shifts to the minute granularity)? 

6. It actually still have spot pagination; I need to move the elements to the top toolbar (in fact, this PR should still have spot pagination on the bottom, the viewport frame removal one does not)

7. Can readd.

8. Define "overkill" (or rather, overkill how?)

PR #104, opened 2018-04-10
Use Window As Viewport

Christopher-Chianelli:
Depends on #103. This PR removes the viewport frame from the roster screen, causing the Window to be the viewport.


Christopher-Chianelli:
@mbarkley half-ish; this being ported to 7.7.x as is, but not to master (which will have everything but grid refactor)

PR #107, opened 2018-04-11
Add a simple SolverTest to smoke test the DRL, solver configuration, etc

ge0ffrey:
Currently this test fails, but once Christopher's DRL fix is in, I 'll ask jenkins to retest this.

### DIFF
--- a/optashift-employee-rostering-server/src/test/java/org/optaplanner/openshift/employeerostering/server/solver/SolverTest.java
+++ b/optashift-employee-rostering-server/src/test/java/org/optaplanner/openshift/employeerostering/server/solver/SolverTest.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.openshift.employeerostering.server.solver;
+
+import org.junit.Test;
+import org.optaplanner.core.api.solver.Solver;
+import org.optaplanner.core.api.solver.SolverFactory;
+import org.optaplanner.core.config.solver.termination.TerminationConfig;
+import org.optaplanner.openshift.employeerostering.server.roster.RosterGenerator;
+import org.optaplanner.openshift.employeerostering.shared.roster.Roster;
+
+import static org.junit.Assert.*;
+
+public class SolverTest {
+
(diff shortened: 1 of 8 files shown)


optashift-employee-rostering-server/src/test/java/org/optaplanner/openshift/employeerostering/server/solver/SolverTest.java
```diff
@@ -0,0 +1,40 @@
...
+import org.optaplanner.openshift.employeerostering.shared.roster.Roster;
+
+import static org.junit.Assert.*;
+
+public class SolverTest {
+
+    @Test(timeout = 600000)
+    public void solve() {
+        SolverFactory<Roster> solverFactory = SolverFactory.createFromXmlResource(WannabeSolverManager.SOLVER_CONFIG);
+        solverFactory.getSolverConfig().setTerminationConfig(new TerminationConfig().withBestScoreFeasible(true));
```
ge0ffrey: If asking for a feasible solution is too much to ask, then the data generator needs to be adjusted so it isn't.

optashift-employee-rostering-webapp/pom.xml
```diff
@@ -61,11 +61,6 @@
       <artifactId>optashift-employee-rostering-restclient</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
```
Christopher-Chianelli: I am pretty sure webapp has Unit tests

ge0ffrey: I move the junit dependency to the top level pom.
Junit (and maybe mockito) is the only "default" dependency (but only in test scope) for all modules.

ge0ffrey:
jenkins retest this

optashift-employee-rostering-server/src/test/java/org/optaplanner/openshift/employeerostering/server/solver/SolverTest.java
```diff
@@ -0,0 +1,40 @@
...
+        Solver<Roster> solver = solverFactory.buildSolver();
+        Roster roster = new RosterGenerator().generateRoster(10, 7);
+        solver.solve(roster);
+        assertNotNull(roster.getScore());
```
rsynek: It would be good to know whether the retrieved score is also different from HardSoftScore.ZERO; if it changed at all. Second thing, I would add also simple check for the returned solution too, not only score.

And one tip - if you like to make assertions easier for yourself and for anybody who is going to read it, try AssertJ. It became very popular among QE.

ge0ffrey: The return of a non-zero hard score is a responsibility for optaplanner's tests, I 'd argue (our users can thrust our code, we can't), but it can't hurt, although it's overweight maybe.

Checking the returned solution is a good idea to further improve this work.

ge0ffrey:
jenkins retest this

optashift-employee-rostering-server/src/test/java/org/optaplanner/openshift/employeerostering/server/solver/SolverTest.java
rsynek: Ok, probably no need to check the score in depth, only the returned solution.

Christopher-Chianelli:
jenkins retest this

PR #111, opened 2018-04-11
[PLANNER-1040] optashift-employee-rostering-shared persistence.xml parameterizing

calvinzhuca:
Signed-off-by: calvin zhu <calvinzhu@hotmail.com>

PR #112, opened 2018-04-17
PR-111 Fixed

Christopher-Chianelli:
@calvinzhuca @ge0ffrey This is a fixed version of #111

### DIFF
--- a/optashift-employee-rostering-webapp/pom.xml
+++ b/optashift-employee-rostering-webapp/pom.xml
@@ -228,6 +228,13 @@
                     <cargo.jboss.configuration>standalone</cargo.jboss.configuration>
                     <cargo.jvmargs>-Xmx512m</cargo.jvmargs>
                   </properties>
+                        <xmlReplacements>
+                          <xmlReplacement>
+                            <file>configuration/standalone.xml</file>
+                            <xpathExpression>//spec-descriptor-property-replacement</xpathExpression>
+                            <value>true</value>
+                          </xmlReplacement>
+                        </xmlReplacements>
                 </configuration>
               </configuration>
             </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -47,6 +47,7 @@
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
     <wildfly.parent>${project.basedir}/local/appserver</wildfly.parent>
+    <wildfly.server.config>wildfly-standalone-config.xml</wildfly.server.config>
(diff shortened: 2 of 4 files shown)


Christopher-Chianelli:
It appears the config file has some hard-coded paths for some reason... cargo made xml replacement easy, but I don't see any such option for the wildfly plugin (note: the plugin installs wildfly into local/appserver; we don't want users to modify the config file by hand)...

optashift-employee-rostering-webapp/pom.xml
```diff
@@ -228,6 +228,13 @@
                     <cargo.jboss.configuration>standalone</cargo.jboss.configuration>
                     <cargo.jvmargs>-Xmx512m</cargo.jvmargs>
                   </properties>
+                        <xmlReplacements>
```
ge0ffrey: identation

pom.xml
```diff
@@ -304,6 +305,10 @@
                   link="${wildfly.launchdir}/standalone/deployments/${deployment.name}.war"
                   resource="${deployment.parent}/${deployment.name}"
                   overwrite="true" />
+                <copy
+                  tofile="${wildfly.launchdir}/standalone/configuration/optashift-employee-rostering-standalone.xml"
```
ge0ffrey: -1 Veto on customizing the wildfly installation.
Our war file should "just run" on an out-of-the-box installation of wildfly, EAP, wildfly swarm or even weblogic some day.

ge0ffrey:
Using maven resource filtering might be a solution (although I hope there's an even simpler one):
https://maven.apache.org/plugins/maven-resources-plugin/examples/filter.html
Do note that we have a convention never to filter src/main/resources. Instead we define a second resources directory that is filtered: src/main/filtered-resources

wildfly-standalone-config.xml
```diff
@@ -0,0 +1,516 @@
+<?xml version='1.0' encoding='UTF-8'?>
+
+<server xmlns="urn:jboss:domain:5.0">
```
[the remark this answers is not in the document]
Christopher-Chianelli: The cargo plugin has a good way of dealing with this: it allow you to specify a xpath (kinda like a regex for xml), allowing you to change specific elements. All that I am doing is changing a single element from "false" to "true". Unfortunately, the wildfly plugin does not provide any utility for doing this. The reason I need for it to be "true" is that wildfly won't replace the variables if it is "false", causing a build error.

pom.xml
Christopher-Chianelli: Agreed, but if we are to use System variables, we MUST change an element in the configuration, otherwise build will fail (wildfly plugin does not provide any other way to change that configuration)

ge0ffrey:
Closing because we'll look into the filtered-resources approach

PR #116, opened 2018-04-20
Openshift fixes

ge0ffrey:


openshift/templates/optashift-employee-rostering-template-binary.yaml
```diff
@@ -136,7 +137,7 @@ parameters:
   displayName: Git source repository
   name: GIT_URI
   required: true
-  value: https://github.com/ge0ffrey/optashift-employee-rostering
+  value: https://github.com/kiegroup/optashift-employee-rostering
 - description: Git branch/tag reference
```
Christopher-Chianelli: Is that minus next to description suppose to be there?

openshift/templates/optashift-employee-rostering-template.yaml
```diff
@@ -135,7 +136,7 @@ parameters:
   displayName: Git source repository
   name: GIT_URI
   required: true
-  value: https://github.com/ge0ffrey/optashift-employee-rostering
+  value: https://github.com/kiegroup/optashift-employee-rostering
 - description: Git branch/tag reference
```
Christopher-Chianelli: Same here...

openshift/templates/optashift-employee-rostering-template-binary.yaml
ge0ffrey: the - is to signal the start of a new element in an array (yaml syntax)

PR #118, opened 2018-04-20
Fixed styling of headers

Christopher-Chianelli:
Fixes Selection_720.png and Selection_722.png of https://issues.jboss.org/browse/PLANNER-1097


optashift-employee-rostering-gwtui/src/main/resources/org/optaplanner/openshift/employeerostering/gwtui/public/roster.less
```diff
@@ -106,31 +106,31 @@
...
 }
 
 .rotation-page-viewport > .date-ticks {
-    top: 120px;
+    top: 190px;
 }
 
 .spot-roster-page-viewport > .time-ticks {
-    top: 205px;
+    top: 235px;
```
ge0ffrey: Can we still see 7 days on the screen?

PR #120, opened 2018-04-23
Clean up the provision script

ge0ffrey:


ge0ffrey:
The main reason is that the non-binary deployment never works, so we shouldn't con people in thinking it does.

ge0ffrey:
jenkins retest this

PR #122, opened 2018-04-24
Use browser timezones on client

Christopher-Chianelli:
-I use moment-timezone to add timezone support to client
-Due to javac using JDK-8 time, need to copy some classes from emulated to client for it to not complain about "final classes/protected constructors/etc" when adding my own zone id which is a wrapper for moment-timezone Zone
-Added a widget for selecting a date-time, DateTimeSelector
-Started some work on Tenant CRUD screen


optashift-employee-rostering-gwtui/src/main/java/java/time/ZoneId.java
```diff
@@ -0,0 +1,17 @@
+package java.time;
+
+import java.time.zone.ZoneRules;
+
+// Stub class since the java compiler do not know about supersources
+// See (supersources for REAL class)
+public class ZoneId {
```
ge0ffrey: We can't have a class in the package java.time in src/main/java
Let's talk about why this one is needed.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/util/DateTimeUtils.java
```diff
@@ -1,23 +1,173 @@
...
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+import jsinterop.annotations.JsPackage;
+import jsinterop.annotations.JsType;
+import org.optaplanner.openshift.employeerostering.gwtui.client.tenant.TenantStore;
+import org.optaplanner.openshift.employeerostering.shared.common.GwtJavaTimeWorkaroundUtil;
+
+@Singleton
 public class DateTimeUtils {
```
rsynek: This class contains some logic, partly in native part that will be compiled into JS. Is there any way of testing it?

Christopher-Chianelli: I think a test.js file came with the moment.js file from the mvn repo; I can make it inject something which is just an interface to moment, and we can inject something else to make sure it is using the right api calls

tiagobento:
@Christopher-Chianelli Is this PR still WIP?

Christopher-Chianelli:
@tiagobento I will close this, me and @ge0ffrey decided to change the views to use LocalDateTime (server still uses OffsetDateTime) so we don't need to import dependencies to handle ZoneId on client

tiagobento:
@Christopher-Chianelli Ok. Thx.

PR #123, opened 2018-04-24
Force hibernate to keep zone offset infomation

Christopher-Chianelli:
HIbernate will automatically convert ZoneOffset to system without this converter.


ge0ffrey:
Being researched in https://stackoverflow.com/questions/50028241/jpa-storing-offsetdatetime-with-zoneoffset

ge0ffrey:
JPA converters cannot go to 2 different columns, but the hibernate specific stuff can, as shown here:
https://github.com/kiegroup/optaplanner/blob/master/optaplanner-persistence/optaplanner-persistence-jpa/src/main/java/org/optaplanner/persistence/jpa/impl/score/AbstractScoreHibernateType.java#L39
https://github.com/kiegroup/optaplanner/blob/master/optaplanner-persistence/optaplanner-persistence-jpa/src/main/java/org/optaplanner/persistence/jpa/impl/score/buildin/hardsoft/HardSoftScoreHibernateType.java

Christopher-Chianelli:
Being worked on in https://github.com/kiegroup/optashift-employee-rostering/pull/128

PR #124, opened 2018-04-25
OpenShift template clean up

ge0ffrey:
* Reduce verbosity
* Improve order of attributes (input before strategy before output)
* Increase resource limits so it works on OpenShift Dedicated

### DIFF
--- a/openshift/templates/optashift-employee-rostering-template-binary.yaml
+++ b/openshift/templates/optashift-employee-rostering-template-binary.yaml
@@ -1,23 +1,29 @@
 apiVersion: v1
 kind: Template
+metadata:
+  name: optashift-employee-rostering-template
+  annotations:
+    description: "Employee rostering web application and REST service based on OptaPlanner."
+    iconClass: "icon-wildfly"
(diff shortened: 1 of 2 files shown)


openshift/templates/optashift-employee-rostering-template-binary.yaml
```diff
@@ -1,145 +1,114 @@
...
+          containers:
+          - name: "webapp-container"
+            image: "webapp-image"
+            ports:
+              - containerPort: 8080
+                protocol: TCP
+            resources:
+              limits:
+                cpu: 2
+                memory: 2Gi
```
Christopher-Chianelli: Do we need 2Gi, or did you just set it to the max available?

ge0ffrey: This says we need it. And I find it too unreliable with less - it might work most of the time on 1Gi, but as soon as they start solving multiple tenants - it can go bad.
This does mean we can't deploy this on OpenShift Starter...

PR #125, opened 2018-04-26
Use s2i wildfly docker image of docker hub + reduce verbosity

ge0ffrey:


ge0ffrey:
I've added the comment "WildFly 11.0" so that file shows up in a find-in-path for "11.0" (when wildfly gets upgraded)

PR #126, opened 2018-04-27
Generated test data must have a feasible solution

ge0ffrey:


optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/roster/RosterGenerator.java
```diff
@@ -55,12 +55,14 @@
 import org.optaplanner.openshift.employeerostering.shared.tenant.Tenant;
 import org.optaplanner.openshift.employeerostering.shared.tenant.TenantConfiguration;
 
-import static java.util.stream.Collectors.groupingBy;
+import static java.util.stream.Collectors.*;
```
Christopher-Chianelli: Are you sure we want to import all Collectors?

ge0ffrey: for static imports, using  a * is fine normally

PR #128, opened 2018-04-30
DO NOT MERGE - OffsetDateTimeHibernateType to fix the ZoneOffset time lost problem

ge0ffrey:
Fails with

```
15:32:25,210 ERROR [org.hibernate.internal.SessionFactoryImpl] (ServerService Thread Pool -- 5) HHH000177: Error in named query: Shift.filter: org.hibernate.hql.internal.ast.QuerySyntaxException: >= operator not supported on composite types. [select distinct sa from org.optaplanner.openshift.employeerostering.shared.shift.Shift sa left join fetch sa.spot s left join fetch sa.rotationEmployee re left join fetch sa.employee e where sa.tenantId = :tenantId and sa.endDateTime >= :startDateTime and sa.startDateTime < :endDateTime order by sa.startDateTime, s.name, e.name]
	at ...
Caused by: org.hibernate.hql.internal.ast.QuerySyntaxException: >= operator not supported on composite types.
         at ...
```

See https://stackoverflow.com/questions/50132548/hibernate-compositeusertype-that-is-comparible-in-jpa-ql-or-hql-query

### DIFF
--- a/optashift-employee-rostering-server/pom.xml
+++ b/optashift-employee-rostering-server/pom.xml
@@ -24,6 +24,13 @@
       <artifactId>javaee-api</artifactId>
       <scope>provided</scope>
     </dependency>
+    <!-- Hibernate -->
+    <dependency>
+      <groupId>org.hibernate</groupId>
+      <artifactId>hibernate-core</artifactId>
+      <!-- Used at compile time by OffsetDateTimeHibernateType -->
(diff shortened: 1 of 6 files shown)


ge0ffrey:
It all works in our testing, so we're going to leave this broken PR alone.

PR #130, opened 2018-05-01
Fixed SpotPopoverForm

Christopher-Chianelli:
-Created a widget that uses `input[date] + input[time]` to be a `LocalDateTime` Widget (in theory, should fall-back to text input okay)
-Removed RotationEmployee from SpotPopoverForm
-Allow setting the Employee if the shift is not pinned


optashift-employee-rostering-gwtui/src/main/resources/org/optaplanner/openshift/employeerostering/gwtui/client/common/LocalDateTimePicker.html
```diff
@@ -0,0 +1,9 @@
+<div class="local-date-time-picker">
+  <!-- Ideally by the year 9999 all browser support input type date so we don't need pattern -->
+  <!-- TODO: Load pattern from i18n (pattern is ignored by browser, so no way to set
+       date format on browsers that do support input type date) -->
+  <input class="form-control" data-field="date-picker" type="date"
+  pattern="(0[1-9]|1[0-9]|2[0-9]|3[01])/(0[1-9]|1[012])/[0-9]{4}" required></input>
```
mbarkley: This is a situation where it would probably be simpler to validate the date in the presenter and show a custom error message, but since this is a corner case it's probably best to leave it for now.

PR #132, opened 2018-05-01
Added a notification system for when shifts are published

Christopher-Chianelli:


manstis:
We have two other Notification systems in kie-wb-common; it might be nice to consider reuse of one. Is there good reason not to?

Christopher-Chianelli:
@manstis Agreed, but I dunno if I can port the notification system time-wise

Christopher-Chianelli:
(for this to be ready for Summit)

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/common/JQuery.java
```diff
@@ -0,0 +1,17 @@
+package org.optaplanner.openshift.employeerostering.gwtui.client.common;
+
+import elemental2.dom.HTMLElement;
+import jsinterop.annotations.JsMethod;
+import jsinterop.annotations.JsPackage;
+import jsinterop.annotations.JsType;
+
+@JsType(isNative = true, namespace = JsPackage.GLOBAL)
+public class JQuery {
```
ge0ffrey: I would expect that Errai might have support for this already? @mbarkley

optashift-employee-rostering-server/src/main/java/org/optaplanner/openshift/employeerostering/server/roster/RosterRestServiceImpl.java
```diff
@@ -339,12 +340,14 @@ public void updateShiftsOfRoster(Roster newRoster) {
 
     @Override
     @Transactional
-    public void publishAndProvision(Integer tenantId) {
+    public List<LocalDate> publishAndProvision(Integer tenantId) {
```
ge0ffrey: Let's use a strongly typed approach. Here we are hardcoding that that is a List of 2 values, and the elements are not self-explenatory.

Let's return a new class PublishResult with has 2 LocalDate fields.

Christopher-Chianelli: Done

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/common/JQuery.java
mbarkley: We have an example like this in the Errai Tutorial, but we don't provide wrappers for any JavaScript libraries in Errai.

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/roster/PublishResult.java
```diff
@@ -0,0 +1,37 @@
+package org.optaplanner.openshift.employeerostering.shared.roster;
+
+import java.time.LocalDate;
+
+public class PublishResult {
+    private LocalDate publishedFromInclusive;
```
ge0ffrey: => publishedFromDate

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/roster/PublishResult.java
```diff
@@ -0,0 +1,37 @@
...
+    private LocalDate publishedToExclusive;
```
ge0ffrey: => publishedToDate // exclusive

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/roster/PublishResult.java
mbarkley: Perhaps, `inclusivePublishedStart` (and `exclusivePublishEnd`)? I don't think it hurts to have the inclusive/exclusive in the name.

Christopher-Chianelli: It to be consistent with Shift/EmployeeAvaliability/All other date intervals fields; The inclusiveness of start and the exclusiveness of end is implict; if we were to change it here, we need to change it everywhere

PR #134, opened 2018-05-01
Use single click to add shifts; fixed delete in popup

Christopher-Chianelli:
Due to `click` being fired on the end of drag events, needed to use a complicated `mousedown`, `mousemove`, `mouseup` device to check if the user clicked without dragging.

### DIFF
--- a/optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/rostergrid/view/SubLaneView.java
+++ b/optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/rostergrid/view/SubLaneView.java
@@ -53,6 +53,8 @@ public class SubLaneView<T> implements ListElementView<SubLane<T>> {
     private Lane<T> lane;
     private ListView<Lane<T>> lanes;
 
+    private boolean hasMouseMoved = false;
+
     @Override
     @SuppressWarnings("unchecked")
     public ListElementView<SubLane<T>> setup(final SubLane<T> subLane,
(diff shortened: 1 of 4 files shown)


optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/rostergrid/view/SubLaneView.java
```diff
@@ -75,31 +77,36 @@ public void destroy() {
...
-            return;
+    @EventHandler("sub-lane")
+    public void onMouseMove(final @ForEvent("mousemove") MouseEvent e) {
+        hasMouseMoved = true;
+    }
+
+    @EventHandler("sub-lane")
+    public void onMouseUp(final @ForEvent("mouseup") MouseEvent e) {
+        if (!hasMouseMoved) {
+            onClick(e);
```
mbarkley: If a "click" fires after a "mouseup", then are you calling onClick twice per click? Also why don't we want to call onClick if the mouse moves during the click?

Christopher-Chianelli: OnClick once - when the user releases the click. The issue lies when dragging/resizing shifts (only available in rotation currently) - often after dragging/resizing the shift, either 1) A new shift is added or 2) The shift edit dialog pop ups. Niether is desired; by ensuring the mouse didn't move during the click, we can ensure that is what the user desired to do. Dragging/Resizing will eventually be added to the Roster Screens.

mbarkley: Right, I was getting confused by the `@ForEvent` annotation on your onClick parameter. Could you remove that?

Christopher-Chianelli: Thanks for finding that! Fixed

PR #144, opened 2018-05-04
Do not merge - experiment to remove distinct and use upper() for sorting

ge0ffrey:
fails currently


ge0ffrey:
Already resolved by Christopher's PR

PR #148, opened 2018-05-16
Refactored code to use DOM/CSS-Grid hybrid

Christopher-Chianelli:
This is a major refactor of the roster screens which is a hybrid of the DOM and Grid approaches. Each lane is an implicit CSS grid. This has several advantages:

- No need for collision detection logic, since CSS implict grid does it for us (see dense in [the spec](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-auto-flow))

- Automatic grid reordering when an object is moved/resized

- Ability to scale the entire grid by changing a single property

- Code reduction

There are several additional refactors included (they can be removed fairly easily):

- Grid Building is now an incremental task, which prevents the user getting sucks for long periods of time when the grid is building.

- Use CSS Global Variables, which allowed us to keep css separate from code (eliminate the need of CSSGridLines)

- Use `JsInterops`

The package `rostergrid` and roster grid classes in `pages` has been removed. The new roster classes can be found in the package `viewport`. 

Mapping from old to new classes:

- `CssGridLines`,`CssGridLineFactory` -> removed (not needed due to CSS global variables)

- `ListView` -> removed (not needed, DOM order doesn't matter in this implementation)

- `SubLane` -> removed (CSS implicit grid creates the sublanes for us)

- `CollisionDetector` -> removed (CSS implicit grid handles collisions for us)

- `Ticks`, `TicksFactory` -> `DateTimeHeader`

- `Lane`-> `Lane`

- `Blob` -> `GridObject`

- `LinearScale` -> `LinearScale`

- `Viewport`, `ViewportView` -> `DateTimeViewport`

- `Orientation` -> `GridObjectPlacer`

- `BlobPopover` -> `FormPopup`

- `*ViewportFactory` -> `*ViewportBuilder`

- `*Page` -> `*Page`, `*DateTimeViewport` (`DateTimeViewport` handles calling the builder)

Additionally, this refactor addresses most issues @tiagobento pointed out in the last PR:

1.  See above
 
2.  Added Tests
    
3.  Used JQuery UI to do resizing/dragging

4.  Vertical Grid lines are pure CSS

5,  Use `margin-left` and `margin-right` to fine-tune spot location

6.  We have pagnation

7.  I didn't add the red line back yet, but it should be simple
   
8.  Maybe overkill


optashift-employee-rostering-gwtui/pom.xml
```diff
@@ -128,6 +135,9 @@
         <artifactId>gwt-maven-plugin</artifactId>
         <configuration>
           <moduleName>org.optaplanner.openshift.employeerostering.gwtui.WorkerRosteringWebapp</moduleName>
+          <includes>
+            <include>**/*Test.java</include>
```
ge0ffrey: Having an include means that everything that is not explicitly included is excluded. Is that intended?

1) It seems weird to me that we want to include *Test.java, as we might have test helper classes in src/test/java.

2) I would except the tbroyer plugin to just do the right thing. Configuration by exception. Ideally, we don't need to do any includes, and we just follow the tbroyer plugin conventions.

ge0ffrey:
@tiagobento This is disrupting further changes (high risk to merge conflicts), would you mind reviewing this today? We're aware it's a big bang, so it's hard to grasp every detail...

optashift-employee-rostering-gwtui/pom.xml
Christopher-Chianelli: @ge0ffrey it due to how the gwt plugin run tests; by default, it only runs suites (see https://tbroyer.github.io/gwt-maven-plugin/test-mojo.html#includes and https://github.com/tbroyer/gwt-maven-plugin/issues/56). Some of these tests require different runners (GwtMockitoRunner), so I am not sure if it will work when run as a Suite.

ge0ffrey: tbroyer's choice to call this just "includes" is confusing. Only the goal gwt:test has it currently, but if other goals ever get an includes element (for example which sources to include to compile), this is going to create havoc.

Can we confine this includes configuration for the goal gwt:test? Something like:
     <execution>
                 <goal>gwt:test</>
                 <configuration>
                       <includes>...</></></>

ge0ffrey:
jenkins retest this

ge0ffrey:
@Christopher-Chianelli DynamicScaleTest.test seems to be failing due an assertion error.

expected:<1008.0> but was:<1000.2205071664829>
	at org.optaplanner.openshift.employeerostering.gwtui.client.viewport.impl.DynamicScaleTest.testScale(DynamicScaleTest.java:32)

Let's also try avoiding the method name "test()"? We shouldn't even prefix those methods names with the word "test" (that's JUnit 3 legacy). I 'd just rename this to "scale()", the Test annotation makes it clear it's a test.

ge0ffrey:
Once the gwt:test execution and the DynamicScaleTest is fixed, I 'll merge.

Christopher-Chianelli:
jenkins retest this

PR #149, opened 2018-05-23
Update Readme with SuperDevMode instructions for IDEA

yurloc:
Includes a single change in POM which shouldn't break the current command-line approach.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -93,11 +93,48 @@ To reproduce locally, add it explicitly (the `provision.sh` script already does 
 mvn clean package -P openshift
 ```
 
-=== IntelliJ IDEA 2017.3 (and above)
+=== IntelliJ IDEA Ultimate 2018.1
 
-. https://youtrack.jetbrains.com/issue/IDEA-171158[IntelliJ 2017.3 now supports the new GWT plugin]
-. `TODO: Test and document`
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -365,7 +365,7 @@
               <!--<arg>-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005</arg> -->
             </jvmArgs>
             <!-- GWT Super Dev Mode: deploy backend correctly -->
-            <launcherDir>${deployment.parent}/${deployment.name}</launcherDir>
+            <launcherDir>optashift-employee-rostering-webapp/target/optashift-employee-rostering-webapp-${project.version}</launcherDir>
```
Christopher-Chianelli: This will probably prevent automatic JavaScript update in the local Wildfly instance (that is, when you run with `mvn wildfly:start wildfly:deploy`); you should also modify the ant run script so the symlink points to `optashift-employee-rostering-webapp/target/optashift-employee-rostering-webapp-${project.version}`.

Christopher-Chianelli: Actually, `${deployment.parent}/${deployment.name}` resolves to the full path of optashift-employee-rostering-webapp/target/optashift-employee-rostering-webapp-${project.version}, so it should be fine, but why did you need to change it?

PR #150, opened 2018-05-23
PLANNER-1085: Added an "Add Tenant" form

Christopher-Chianelli:


optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/common/LocalDatePicker.java
```diff
@@ -0,0 +1,49 @@
...
+        datePicker.valueAsDate = new Date(Date.parse(value.toString()));
+    }
+
+    @Override
+    public LocalDate getValue() {
+        Date date;
+        if (reportValidity()) {
+            try {
+                date = datePicker.valueAsDate;
+            } catch (Exception e) {
```
ge0ffrey: Can we catch anything less general then Exception?
It's usually a bad idea to catch Throwable, Error, Exception or RuntimeException, which a few notable exceptions. For example: any NullPointerException or OutOfMemoryError should bubble up to the default exception handler and show a big fat red error dialog, and not been seen as a "the user filled in an invalid date".

optashift-employee-rostering-shared/src/main/java/org/optaplanner/openshift/employeerostering/shared/tenant/TenantRestService.java
```diff
@@ -66,5 +67,58 @@
...
     TenantConfiguration updateTenantConfiguration(@ApiParam(required = true) TenantConfiguration tenantConfiguration);
+    
+    // TODO: Where should this be?
+    @ApiOperation("Get supported timezones")
+    @GET
+    @Path("/supported/timezones")
+    List<ZoneId> getSupportedTimezones();
+    
+    // Workaround for addTenant since request can have at most one body parameter
+    public static class NewTenant {
```
ge0ffrey: Let's refactor the model to deal with this cleanly.

Obviously, we can't send in Tenant, as that needs to remain extremely lightweight (for the list tenants method that is always called). But what if we send in a RosterState and that embeds a Tenant? The TenantConfiguration is not needed to be send in anyway, as all those fields are defaulted and not in the form?

That being said, we need to discuss about TenantConfiguration - that should probably be renamed to RosterParametrization - and maybe we want to merge it with RosterState. Let's discuss after the meeting scheduling meeting.

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/common/LocalDatePicker.java
Christopher-Chianelli: Actually cannot test this case, the try/catch block is not checking for "the user filled in an invalid date?" (`reportValidity()` return true iff it is a valid date), but rather "does the browser support the `valueAsDate` computed field?" Both Chrome and Firefox support it, but I don't know if IE supports it.

Christopher-Chianelli:
jenkins retest this

kartiksura:
Is there a way I could use this PR? If i use it directly it is giving me a timezone related exception.

Christopher-Chianelli:
@kartiksura Cannot reproduce said exception; PR is working as intended on Wildfly 11 (tested on latest Chrome and Firefox). To check out this PR in git, execute `git fetch {remote} pull/150/head:{branchname}`, follow by `git checkout {branchname}`, where {remote} is the remote pointing to this repo, and {branchname} is what you want to name your branch. After checking out the PR, run `mvn clean install`. You can then test by copying the war file in "optashift-employee-rostering-webapp/target" to "local/appserver/wildfly-11.0.0.Final/standalone/deployments/". If "local/appserver/wildfly-11.0.0.Final/standalone/deployments/" is not empty, remove the symlink (which is used for `mvn gwt:codeserver`), and run with `local/appserver/wildfly-11.0.0.Final/bin/standalone.sh` when in "local/appserver/wildfly-11.0.0.Final/bin". The webapp will not work in OpenShift Online since it take too many resources; should work in OpenShift Dedicated or OpenShift Origin though. If the exception still occurs, please post the exception in the comments.

Christopher-Chianelli:
(alternatively, you can run the webapp via maven: `mvn clean install`, `mvn -N wildfly:start wildfly:deploy`, `mvn gwt:codeserver`)

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/tenant/NewTenantForm.java
```diff
@@ -0,0 +1,168 @@
...
+    @DataField("schedule-start-weekday")
+    private HTMLDivElement scheduleStartWeekday;
+
+    @Inject
+    @DataField("publish-notice")
+    private HTMLInputElement publishNotice;
+
+    @Inject
+    @DataField("publish-length")
+    private HTMLInputElement publishLength;
```
yurloc: This field should be read-only according to the feature description: https://issues.jboss.org/browse/PLANNER-1085.

@ge0ffrey is this requirement still valid?

optashift-employee-rostering-gwtui/src/main/java/org/optaplanner/openshift/employeerostering/gwtui/client/common/LocalDatePicker.java
ge0ffrey: I understand, but we don't support IE, only Chrome, Firefox and probably Safari and Edge. Feel free to completely ignore Internet Explorer, it's end of life.

I've removed the Exception catches and tested on Firefox and Chrome successfully.

As for Edge, I 'd rather see proof that it doesn't work before introducing code that we might not need. QA will be certifying against multiple browsers, so they'll detect it if it's a problem. Alternatively, we should try to get Edge and Safari running on our system and simply try it, but that's challenging.

PR #170, opened 2018-06-14
Overconstrained planning

ge0ffrey:
To try it out:
- open a dataset, solve it and notice that all shifts get assigned
- now pick a date and start adding new shifts for a spot there. Add so many shifts, that there are more shifts on that date then there are employees. Click solve. Notice that it still finds a schedule with no hard constraints broken, but now there are medium constraints broken: a few shifts didn't get assigned.

@Christopher-Chianelli Can you visualize the medium score in the UI too? And the indictement work will make the unassigned shifts stick out like a soar thumb, of course.


Christopher-Chianelli:
@ge0ffrey That easy enough, although maybe we should call it "unassigned shifts" since only unassigned shifts decrease the medium score currently. I will rebase the indictment work on this branch and change the color of the shifts to make them stick out.

ge0ffrey:
@Christopher-Chianelli merged, so you can leave you indictement branch to work from master

PR #178, opened 2018-07-25
PLANNER-1152 Persistence xml filtering to make it docker friendly

ge0ffrey:


pom.xml
```diff
@@ -464,6 +471,23 @@
         </plugins>
       </build>
     </profile>
+    <profile>
+      <!-- To activate use -DproductizedOpenshift (not -P) -->
```
Christopher-Chianelli: Non-blocker, but why did you set it to activate on a property?

ge0ffrey: because -DproductizedOpenshift actives profile openshift too (which it needs because the cargo tests will fail otherwise)

PR #184, opened 2018-08-09
Increase cargo deployment timeout

rsynek:
Increasing to two minutes which should satisfy the kieAllBuild.


ge0ffrey:
Merged early as the kieAllBuild is failing because of this

ge0ffrey:
@rsynek I presume you're also doing the same change on master?

rsynek:
@ge0ffrey I want to wait whether this solves the failure. If so, sure I will open a PR to master.

PR #188, opened 2018-08-24
PLANNER-1220: Added @JoinTable to Spot and Employe Many-to-Many relations

Christopher-Chianelli:
This change removes warnings when using ProgresSQL on Wildfly


Christopher-Chianelli:
@rsynek I would say #191 depends on this PR, but this PR addresses a separate issue (Posgres warnings) that MySQL also shares (hence the duplicate commit).

PR #191, opened 2018-08-29
PLANNER-1221: To decrease deployment time on MySQL, allowed user to select generation stategy

Christopher-Chianelli:
Hibernate 5 use TABLE id generation for MySQL since it doesn't support sequences. This is very slow,
making it take longer than 300s to generate the example data and start the application. Instead, it
should use IDENTITY, which allows it to generate all the example data in 75s (boot takes 122s in total).

By default, org.optaweb.employeerostering.persistence.id.generator has value AUTO.

Depends on #188, since MySQL have the same warnings

### DIFF
--- a/employee-rostering-server/src/main/filtered-resources/META-INF/persistence.xml
+++ b/employee-rostering-server/src/main/filtered-resources/META-INF/persistence.xml
@@ -4,6 +4,7 @@
   <persistence-unit name="optaweb-employee-rostering-persistence-unit" transaction-type="JTA">
     <description>OptaWeb Employee Rostering Persistence Unit</description>
     <jta-data-source>${org.optaweb.employeerostering.persistence.datasource}</jta-data-source>
+    <mapping-file>META-INF/orm.xml</mapping-file>
     <class>org.optaweb.employeerostering.shared.skill.Skill</class>
     <class>org.optaweb.employeerostering.shared.spot.Spot</class>
     <class>org.optaweb.employeerostering.shared.employee.Employee</class>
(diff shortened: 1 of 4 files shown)


employee-rostering-server/src/main/filtered-resources/META-INF/persistence.xml
```diff
@@ -4,6 +4,7 @@
   <persistence-unit name="optaweb-employee-rostering-persistence-unit" transaction-type="JTA">
     <description>OptaWeb Employee Rostering Persistence Unit</description>
     <jta-data-source>${org.optaweb.employeerostering.persistence.datasource}</jta-data-source>
+    <mapping-file>META-INF/orm.xml</mapping-file>
```
ge0ffrey: Check if the benchmark app still works - that has a copy of this persistence.xml without filtering

Christopher-Chianelli: @ge0ffrey The benchmark app still worked (checked by running `mvn org.codehaus.mojo:exec-maven-plugin:1.6.0:java -Dexec.mainClass=org.optaweb.employeerostering.benchmark.OptaWebEmployeeRosteringBenchmarkApp`)

ge0ffrey:
Please add the property in this file too:
  https://github.com/kiegroup/optaweb-employee-rostering/blob/master/employee-rostering-docs/src/main/asciidoc/Introduction/SystemProperties/SystemProperties-section.adoc

Looks good to me to merge then (feel free to merge once it's fixed).
Sorry that I created a merge conflict for you - I had to fast-track the hbm2ddl change to unblock David's work and forgot about this PR still being here.

Christopher-Chianelli:
Fixed the merge conflict and verified it still worked on a MySQL server. Will merge when green.

Christopher-Chianelli:
@ge0ffrey added property to docs, can you confirm it is good?

PR #187, opened 2018-08-24
PLANNER-938: Added IT test for REST services

Christopher-Chianelli:

PR #193, opened 2018-08-30
PLANNER-1240: Fixed bug in updateRotation which discarded rotation employee in ShiftTemplate

### LINKED ISSUE #175: Rotation savnig
if i save rotation it make all assigned slots as unassigned, am i missing something ?

Christopher-Chianelli:
Fixes #175. The updateRotation method may be removed to match single entity CRUD operation of other entities (Employee, Spot, Shift, etc.) instead of batch (slightly related to https://issues.jboss.org/browse/PLANNER-1224).


Christopher-Chianelli:
@yurloc agreed, added the majority of integration tests in https://github.com/kiegroup/optaweb-employee-rostering/pull/187 (doesn't does full, but does a lot more than it used to). Need to still add unit tests to server and client. I am a bit hesitant to do unit tests/integration tests on rotation, since currently `ShiftTemplate` does not follow the same convection as the other entities (in REST methods are batch operations, while the other entities REST methods are for single entities). Need to discuss this with @ge0ffrey.

PR #203, opened 2018-09-27
PLANNER-1287: Fix bug that cause EmployeeAvailability state to always be null in the Solver

Christopher-Chianelli:


Christopher-Chianelli:
The bug was a constructor not fully constructing an object, causing state to be null. I also modified `doTimeslotsIntersect(...)` so it counts edge cases as being intersecting (if a shift and availability has
the exact same start and end date times). Need to add regression tests, although this regression test is for a constructor, which might be a bit excessive.

PR #205, opened 2018-10-03
upgraded version com.fastewrxml.jackson to align to EAP 7.2.

mbiarnes:
to align to JBEAP 7.2 we have to align to com.fasterxml.jackson 2.9.5.
this was tested with this version - the build was successful and no test failed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
       https://github.com/jboss-integration/jboss-integration-platform-bom/blob/master/pom.xml
     -->
     <version.ch.qos.logback>1.2.3</version.ch.qos.logback>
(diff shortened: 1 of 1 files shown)


ge0ffrey:
jenkins retest this

ge0ffrey:
@mbiarnes The wildfly version for optaweb-employee-rostering is still set to 11, that should upgrade too, in a separate PR:
  https://github.com/kiegroup/optaweb-employee-rostering/blob/master/pom.xml#L83

@Christopher-Chianelli if this one works, please merge it. But would it make sense to upgrade any other components, such as gwt-jackson? (Of so, create jira's for that)

mbiarnes:
will close this PR as it will be raised soon a new PR with this changes

PR #210, opened 2018-11-12
PLANNER-1353 switch from jboss-parent to kie-parent

mareknovotny:
https://issues.jboss.org/browse/PLANNER-1353 and https://issues.jboss.org/browse/PLANNER-1342

Switched parent to kie-parent and fixed all issues with that related change:
- enforcer violations
- GWT compilation

PR #211, opened 2018-11-16
PLANNER-1358: Align versions with kie-parent and jboss-ip-bom

yurloc:
## First commit (clean up)

Improvements in POMs after adopting `kie-parent` in #210. For the most part it replaces `javax.javaee-api` with JBoss Spec APIs taken from https://github.com/jboss/jboss-javaee-specs/blob/master/pom.xml#L91.

Part of this commit fixes a GWT compilation classpath problem described in https://github.com/kiegroup/optaweb-employee-rostering/pull/210#issuecomment-439114090. I originally thought it breaks SuperDevMode but it was a local issue that I resolved by `rm -r /tmp`.

## Following commits (version alignment)

Adresses https://issues.jboss.org/browse/PLANNER-1358.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,12 +3,10 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
 
   <modelVersion>4.0.0</modelVersion>
-  <!-- Use JBoss parent so all maven-plugins have a decent base version. -->
   <parent>
     <groupId>org.kie</groupId>
     <artifactId>kie-parent</artifactId>
     <version>7.16.0-SNAPSHOT</version>
-    <relativePath/>
   </parent>
 
   <groupId>org.optaweb.employeerostering</groupId>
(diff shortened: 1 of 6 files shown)


yurloc:
@mbiarnes I'd like to get this merged sooner than later. It might make things easier for your Hibernate etc. upgrades but it might also introduce some conflicts. Let me know if you find this helpful and we can merge or if you prefer to wait with this until you're done with the upgrade. Thanks!

pom.xml
```diff
@@ -154,68 +138,39 @@
...
-        <groupId>com.fasterxml.jackson.datatype</groupId>
-        <artifactId>jackson-datatype-jsr310</artifactId>
-        <version>${version.com.fasterxml.jackson}</version>
-      </dependency>
-      <!-- KIE -->
-      <dependency>
-        <groupId>org.optaplanner</groupId>
-        <artifactId>optaplanner-bom</artifactId>
+        <groupId>org.jboss.errai.bom</groupId>
+        <artifactId>errai-internal-bom</artifactId>
```
mareknovotny: is this really good to reference internal BOM instead of user facing BOM `org.jboss.errai.bom:errai-bom`

yurloc: Yes. It's not internal in the sense that it shouldn't be used by Errai users. `errai-internal-bom` only contains Errai artifacts whereas `errai-bom` adds 3rd party dependencies that Errai works with and it **includes** (imports) `errai-internal-bom`. In other words, by using `errai-bom` you get `errai-internal-bom` anyway and a bunch of other dependencies that we either don't use or manage ourselves.

yurloc: Also, all parent POMs in kiegroup import `errai-internal-bom`, and not `errai-bom` - that was probably the deciding factor for me to use it here too.

https://github.com/search?q=org%3Akiegroup+errai-internal-bom&type=Code

mareknovotny: ok, I originally recall the Errai only with errai-bom there was really only needed Errai dependencies for the usage.

That internal-errai-bom is useless I think we should ask to merge it into one BOM. 

Anyway I got the point here to import only managed errai deps without its build deps.

PR #213, opened 2018-11-22
PLANNER-1266: Added Contracts

Christopher-Chianelli:


employee-rostering-shared/src/main/java/org/optaweb/employeerostering/shared/contract/Contract.java
```diff
@@ -0,0 +1,145 @@
...
+    public Contract(Integer tenantId, String name, Integer maximumMinutesPerDay, Integer maximumMinutesPerWeek, Integer maximumMinutesPerMonth, Integer maximumMinutesPerYear) {
+        super(tenantId);
+        this.name = name;
+        this.maximumMinutesPerDay = maximumMinutesPerDay;
+        this.maximumMinutesPerWeek = maximumMinutesPerWeek;
+        this.maximumMinutesPerMonth = maximumMinutesPerMonth;
+        this.maximumMinutesPerYear = maximumMinutesPerYear;
+    }
+
+    @AssertTrue
```
ge0ffrey: IIRc, you can do an @Min(0) on numbers with java bean validation instead of doing a positive check via @AssertTrue (which is less declarative).

See usage of @Min on https://www.baeldung.com/javax-validation

ge0ffrey: Btw, all these have a natural @Max too:

```
@Min(0) @Max(60L * 60L * 24L)
private Integer maximumMinutesPerDay
```

employee-rostering-shared/src/main/java/org/optaweb/employeerostering/shared/employee/Employee.java
```diff
@@ -104,7 +111,15 @@ public void setSkillProficiencySet(Set<Skill> skillProficiencySet) {
...
-    @Override
+    public Contract getContract() {
+        return contract;
+    }
+
+    public void setContract(Contract contract) {
+        this.contract = contract;
+    }
+
+    /*@Override
```
ge0ffrey: Don't comment it out, just delete it directly

employee-rostering-shared/src/main/java/org/optaweb/employeerostering/shared/contract/Contract.java
Christopher-Chianelli: Does `@min` work for null values (https://www.baeldung.com/javax-validation did not mentioned how the constraint handle null cases)? `maximumMinutesPerX` can be null if we want to ignore the constraint.

employee-rostering-shared/src/main/java/org/optaweb/employeerostering/shared/employee/Employee.java
Christopher-Chianelli: Thanks for pointing this out; I don't know if I meant to delete it/edit it or I just commented it out for testing. Will try uncommenting.

employee-rostering-server/src/main/resources/org/optaweb/employeerostering/server/solver/employeeRosteringScoreRules.drl
```diff
@@ -89,6 +91,64 @@ rule "No 2 shifts within 10 hours from each other"
         scoreHolder.addHardConstraintMatch(kcontext, -1);
 end
 
+rule "At most contract maximum minutes per day each day"
```
yurloc: Sounds strange. Maybe something like "Daily minutes must not exceed contract maximum"?

Christopher-Chianelli: That sounds better; will change

employee-rostering-webapp/src/test/java/org/optaweb/employeerostering/webapp/contract/ContractRestServiceIT.java
```diff
@@ -0,0 +1,115 @@
...
+        Contract invalidContract = new Contract(TENANT_ID, "invalid contract", -1, null, null, null);
+        assertThatExceptionOfType(javax.ws.rs.InternalServerErrorException.class)
+                .isThrownBy(() -> {
+                    contractRestService.updateContract(TENANT_ID, invalidContract);
+                });
+        assertClientResponseError(Response.Status.INTERNAL_SERVER_ERROR);
+    }
+
+    @Test
+    public void testCrudSkill() {
```
rsynek: This looks like a copy-paste error.

yurloc:
@Christopher-Chianelli I skimmed through the diff but I don't have time to do a full review, so don't wait for my approval. Thanks @rsynek for reviewing!

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/pages/contract/ContractsPage.java
```diff
@@ -0,0 +1,134 @@
...
+import org.optaweb.employeerostering.gwtui.client.common.KieSearchBar;
+import org.optaweb.employeerostering.gwtui.client.notification.NotificationFactory;
+import org.optaweb.employeerostering.gwtui.client.pages.Page;
+import org.optaweb.employeerostering.gwtui.client.tenant.TenantStore;
+import org.optaweb.employeerostering.gwtui.client.util.PromiseUtils;
+import org.optaweb.employeerostering.shared.contract.Contract;
+import org.optaweb.employeerostering.shared.contract.ContractRestServiceBuilder;
+
+@Templated
+public class ContractsPage
```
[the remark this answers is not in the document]
rsynek: We discussed the formatting of interfaces in the class declaration with @yurloc and I realized it might not be possible to achieve consistency with KIE group style across different IDEs. Please take this one comment as a soft suggestion.

employee-rostering-webapp/src/test/java/org/optaweb/employeerostering/webapp/contract/ContractRestServiceIT.java
Christopher-Chianelli: It is; will fix

employee-rostering-webapp/src/test/java/org/optaweb/employeerostering/webapp/contract/ContractRestServiceIT.java
```diff
@@ -0,0 +1,115 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.optaweb.employeerostering.shared.contract.Contract;
+import org.optaweb.employeerostering.shared.contract.ContractRestService;
+import org.optaweb.employeerostering.webapp.AbstractEntityRequireTenantRestServiceIT;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+
+public class ContractRestServiceIT extends AbstractEntityRequireTenantRestServiceIT {
```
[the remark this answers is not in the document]
Christopher-Chianelli: Same thing happens when deleting any entity that is referenced by other entity. Need to send and parse server-side exceptions to make it more friendly. https://issues.jboss.org/browse/PLANNER-933 is the corresponding issue.

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/common/NullableValueElement.java
```diff
@@ -0,0 +1,118 @@
...
+            valueInput.required = true;
+        }
+    }
+
+    @EventHandler("value")
+    public void onValueChange(@ForEvent("change") Event e) {
+        value = getValueFromInput();
+    }
+
+    public void init(Class<T> clazz, T value, String labelText) {
```
rsynek: Apart from using string comparison instead of `isAssignableFrom()` method, I wonder if we need such a complicated init method at all. Generally, conditions base on types can be replaced by class hierarchy - in this case for Integer, Double and String.

rsynek:
@Christopher-Chianelli please take a look at the SonarCloud analysis, there major and critical issues worth looking into. Also the test coverage of UI part of the new code is very small.

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/common/NullableValueElement.java
Christopher-Chianelli: GWT does not support the `isAssignableFrom()` method unfortunately. An unfortunate side effect of using class hierarchy is that the HTML template file would need to be duplicated with no changes and each would need to be maintained separately (although currently we only actually use the Integer version).

yurloc: Does it mean this change is not possible?
```diff
    @Inject
    @DataField("maximum-minutes-per-day")
-    private NullableValueElement<Integer> maximumMinutesPerDay;
    @Inject
    @DataField("maximum-minutes-per-day")
+    private NullableIntegerElement maximumMinutesPerDay;
```
Would it affect the HTML template? Why would I need to duplicate it?

Christopher-Chianelli: Nevermind; `@Templated` can take a fully qualified path, which would solve the duplication issue. (the issue have to do with the implementation of Nullable<Type>Element, which would be a Templated bean). Actually can probably remove the html code duplication from all the Entity List pages using this.

yurloc: Thanks, I have a better understanding now. Would it be possible to make the common ancestor (an abstract class `NullableElement`) `@Templated`, and still use its child, `NullableIntegerElement` like this:
```java
    @Inject
    @DataField("maximum-minutes-per-day")
    private NullableIntegerElement maximumMinutesPerDay;
```
even if `NullableIntegerElement` is not `@Templated`? If that worked, it would allow you to avoid using `@Templated("f.q.c.N")` on each concrete implementation of `NullableElement`.

Christopher-Chianelli: `@Templated("f.q.c.N")` is not the problem, but the solution (it prevent duplication of HTML files). Classes do not inherit annotations, so it is impossible to make the children inherit the `@Templated` annotation. Instead, the parent remains unannotated and we annotate the children.

yurloc: Makes sense. I know that annotations are not inherited but Errai could have been smart and traverse the class hierarchy towards root if I inject an unannotated class as a field in a `@Templated` class.

Your proposal seems good to me though.

Christopher-Chianelli:
@rsynek Can we disable/change some of the checks in SonarCloud? The constants in I18nKeys are named to have a one-to-one correspondence to the corresponding key in the OptaWebUIConstants file. Thus the "rename this constant name to match this regular expression [...]" is against this convention. Some duplicated code is impossible to remove as `@EventHandler(...)` is not repeatable and does not take in a list of argument (Errai threw an error when I tried passing in a list of arguments).

rsynek:
@Christopher-Chianelli we can definitely close some of the found issues - not sure if there is any process of learning based on our feedback to issues. Another option would be defining a new profile, if we agree that we see too many false positives.

"rename this constant name to match this regular expression [...]" - If I understand this correctly, it has nothing to do with the value of the constant, but with the name of with:
`final String ContractForm_perYear = "ContractForm.perYear";` is expected to be `final String CONTRACT_FORM_PER_YEAR = "ContractForm.perYear";`. Could you please explain the relationship between I18nKeys and OptaWebUIConstants? There seem to be 1-on-1 match on values of these constants; e.g. `ContractForm.perYear`.

Not sure which issue you are addressing with the "Some duplicated code is impossible to remove". If you login to the SonarCloud with your GitHub account, I can add you to the organization there and you may comment/close found issues.

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/common/NullableValueElement.java
rsynek: @Christopher-Chianelli if there is no way to replace the class name comparison by a class hierarchy due to limitations of the GWT and Errai framework, would it be at least possible to replace the `getSimpleName()` with an `instanceof` operator? In contrast to the `isAssignableFrom()`, for using `instanceof` the class must be known at compile time, so maybe GWT might support it.

Christopher-Chianelli:
@rsynek See https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/resources/i18n/StunnerWidgetsConstants.java - The actual name of the constant matches the i18n translation, replacing "." with "_". I follow the same convention here.

rsynek:
@Christopher-Chianelli picked some of the remaining issues SonarCloud found that seems to be more meaningful - see my latest comments.

+1 for ignoring those complains about improper constant names, if you already have such a convention.

Will take a look if it's possible to create a profile that would omit some of the checks.

rsynek:
@Christopher-Chianelli would you please mark the comments/suggestions that were addressed? It's becoming difficult to match code changes to individual comments.

Christopher-Chianelli:
@rsynek Done; All issues except for deleting a used contract (which is related to https://issues.jboss.org/browse/PLANNER-933) are addressed.

Christopher-Chianelli:
jenkins retest this

.gitignore
```diff
@@ -10,6 +10,8 @@
 /*.ipr
 /*.iws
 /*.iml
+/.project
+/.settings
```
ge0ffrey: The .gitignore file should be the same as in the optaplanner repo:
  https://github.com/kiegroup/optaplanner/blob/master/.gitignore
there it does /.* to ignore those eclipse files

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/common/EventManager.java
```diff
@@ -118,31 +118,33 @@ public void handleEvent(T value) {
...
+        public static final Event<Void> ROTATION_INVALIDATE = new Event<>("RotationInvalidate", Void.class);
 
         // Invalidation Events
         @SuppressWarnings("rawtypes")
-        public static Event<Class> DATA_INVALIDATION = new Event<>("DataInvalidation", Class.class);
+        public static final Event<Class> DATA_INVALIDATION = new Event<>("DataInvalidation", Class.class);
         @SuppressWarnings("rawtypes")
-        public static Event<Map> SKILL_MAP_INVALIDATION = new Event<>("SkillMapInvalidation", Map.class);
+        public static final Event<Map> SKILL_MAP_INVALIDATION = new Event<>("SkillMapInvalidation", Map.class);
+        @SuppressWarnings("rawtypes")
```
ge0ffrey: In a separate PR, we'll want to remove these "raw types" ignores, or comment why `Event<Map<Object, Object>>` isn't possible, because this half-use of generics leads to full problems.

employee-rostering-server/src/main/resources/org/optaweb/employeerostering/server/solver/employeeRosteringScoreRules.drl
```diff
@@ -89,6 +91,64 @@ rule "No 2 shifts within 10 hours from each other"
...
 end
 
+rule "Daily minutes must not exceed contract maximum"
+    when
+        $employee : Employee($contract : contract, $contract.getMaximumMinutesPerDay() != null)
+        $s : Shift(employee == $employee, $startDateTime : startDateTime)
+        Number( intValue > $contract.getMaximumMinutesPerDay() ) from accumulate(
+            Shift(employee == $employee, $shiftStart : startDateTime,
+            $shiftEnd : endDateTime,
+            DateTimeUtils.dayOf($shiftStart) == DateTimeUtils.dayOf($startDateTime)),
```
ge0ffrey: Remove the class DateTimeUitils, all those methods exist in java.time already.

For example:
- Check if 2 dateTimes are the same day:
-- shiftStart.toLocalDate().equals(startDateTime.toLocalDate())
- number of days between 2 dates:
  https://stackoverflow.com/questions/27005861/calculate-days-between-two-dates-in-java-8

employee-rostering-webapp/src/test/java/org/optaweb/employeerostering/webapp/contract/ContractRestServiceIT.java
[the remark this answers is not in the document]
rsynek: @Christopher-Chianelli would you please add this test for the REST API without implementing the referenced enhancement?

Do we have any general Jira issue describing the behaviour (when deleting any entity that is referenced by other entity) or is it just the PLANNER-933 ?

rsynek:
@Christopher-Chianelli regarding the issue about deleting a contract that is already referenced by some employees, I would rather not merge this PR with it. Instead of implementing error message parsing, you can check if the contract is in use and disable the delete button in the UI. REST API should return e.g. 403 Forbidden + a reasonable message. What do you think?

Screenshot from 2019-04-10 09-35-56

Christopher-Chianelli:
@rsynek There are several issues with that approach:

- First, we need to find out if a contract is in use in the UI. In order to determine this, we can either remove every contract that is in used by an employee in the employee list (adding another REST call, which will get more expensive as the number of employees increase), or we can add "# of references" to the Contract object (which bloats our REST).
- Second, it will create an inconsistency with the Employees, Skills and Spots pages. They currently have the same issue of raising an error if they are used in a Shift. I rather fix them all at once in the same PR.
- Third, this PR is large as is, and the suggested change is not trivial.
- Fourth, users may be confused why they cannot delete the contract; I think it is better to allow them to attempt to delete a contract and see an error which list all the relationships the contract participates in (i.e. the employees that have the contract), and maybe give them several options to fix it (give the employees an other contract, delete all employees that use the contract, or edit the contract). 

I agree that we should fix this eventually, but not in this PR.

Christopher-Chianelli:
@ge0ffrey I made .gitignore match the optaplanner .gitignore; I removed most methods from DateTimeUtils except:

- `doTimeslotsIntersect`, which is as simple as it can be, and is a convenience method
- `weekOf` was changed to `sameWeek`, which use ISO-8601 week numbers to determine if two timeslots begin in the same week.

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/common/EventManager.java
yurloc: Unfortunately this will require new classes (something like `SkillMap extends HashMap<String, Skill>`) and new events for data invalidation but I think it's the correct way to do it. Having a dedicated event for each data type might be some extra code but on the other hand it will allow to simplify this:
```java
eventManager.subscribeToEventForever(Event.DATA_INVALIDATION, clazz -> {
    if (clazz.equals(Employee.class)) {
        updateRowCount();
    }
});
```
to this:
```java
eventManager.subscribeToEventForever(Event.EMPLOYEE_INVALIDATION, aVoid -> updateRowCount());
```

I can do that.

PR #221, opened 2019-01-24
Make base URL handling more robust

rsynek:


rsynek:
> LGTM; but is there a way to append directly to the path via the URL object?
The same URL object cannot be changed. Anyway, the `ResteasyClient#taget()` does not take URL instance - conversion to either String or URI is required.

PR #223, opened 2019-01-31
RHDM-745 The one productized build should include ${} in the persistence.xml

ge0ffrey:


pom.xml
```diff
@@ -387,17 +382,17 @@
       </build>
     </profile>
     <profile>
-      <!-- To activate use -DproductizedOpenshift (not -P) -->
-      <id>productizedOpenshift</id>
+      <!-- To activate use -Dproductized (not -P) -->
+      <id>productized</id>
```
rsynek: If I understand it correctly, this profile leaves the place holders in persistence.xml and Wildfly/EAP has to be configured so that it replaces the place holders by system properties passed during startup.

As the profile activation has changed, has the same change been propagated into the build process?

PR #226, opened 2019-02-19
PLANNER-1438:  Fix Broken RESTEasy Tests after alignment

Christopher-Chianelli:
@mbiarnes I was able to reproduce the failing tests.  They were due to the exclusion of `jboss-jaxrs-api_2.1_spec`. Removing the exclusion fixes the test. What was the duplication issue with` jboss-jaxrs-api_2.1_spec` (i.e. what dependencies included it)?

### DIFF
--- a/employee-rostering-webapp/pom.xml
+++ b/employee-rostering-webapp/pom.xml
@@ -80,7 +80,7 @@
     </dependency>
     <dependency>
       <groupId>org.jboss.spec.javax.ws.rs</groupId>
-      <artifactId>jboss-jaxrs-api_2.0_spec</artifactId>
+      <artifactId>jboss-jaxrs-api_2.1_spec</artifactId>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/employee-rostering-webapp/pom.xml
+++ b/employee-rostering-webapp/pom.xml
@@ -177,6 +177,19 @@
 
   <profiles>
     <profile>
+      <id>productized</id>
+      <activation>
+        <property>
+            <name>productized</name>
+        </property>
+      </activation>
+      <properties>
+        <cargo.maven.skip>true</cargo.maven.skip>
+        <skipITs>true</skipITs>
+      </properties>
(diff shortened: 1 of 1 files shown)


Christopher-Chianelli:
The commented out exclusion will be removed once we find out if the dependency is duplicated elsewhere or not (serves as a reminder to check if it was duplicated).

mbiarnes:
@Christopher-Chianelli lets talk about this tomorrow

yurloc:
@Christopher-Chianelli you can run `mvn clean install -Dfull` to activate Enforcer during the build and make it fail locally to see if there are duplicate classes without having to wait for Jenkins PR build.

yurloc:
As far as I know JBoss Spec JARs for Java EE APIs are preferred over anything `javax.*` (if an equivalent JBoss Spec exists). I'll try to find the source for this statement if you disagree. There is a small chance I might be wrong.

Given the above, the correct solution is to
1. Remove dependency on `javax:javaee-api`.
2. Add required dependencies that were previously provided by `javax:javaee-api`. For example `employee-rostering-webapp` needs `javax.servlet.http.HttpServlet`. Use `org.jboss.spec.javax.servlet:jboss-servlet-api_3.1_spec` (presuming we're on Java EE 7).
3. Remove unnecessary exclusions (e.g. do not exclude `javax.validation:validation-api` as that has no equivalent JBoss Spec).
4. Especially remove exclusions of any `org.jboss.spec.*` dependencies, as these are the preferred ones.

See https://github.com/jboss/jboss-javaee-specs/blob/javaee7/jboss-javaee-web/pom.xml (you might need to switch between branches and tags). You'll find there the complete list JBoss Specs and external APIs.

Christopher-Chianelli:
Closed in favour of https://github.com/kiegroup/optaweb-employee-rostering/pull/229

rsynek:
...and in favour of https://github.com/kiegroup/optaweb-employee-rostering/pull/230

PR #228, opened 2019-02-20
PLANNER-936: Enable Checkstyle.failOnViolation

yurloc:
Should have been done "yesteryear".

Let's merge ASAP to avoid conflicts.


Christopher-Chianelli:
Will merge when green

yurloc:
> Looks good, but PLANNER-936 is not just about enabling checkstyle, but also adding a maven plugin to do the formatting for us (ex: running `mvn format` will format the source code).

Ok, I'll return PLANNER-936 to OPEN and reassign back to you.

PR #230, opened 2019-02-27
[RHDM-860] Employee rostering unable to deploy due to unresolvable placeholder in orm.xml

rsynek:
- removes orm.xml
- configures property replacement on Wildfly via maven cargo

Image and template should not need update as the removed id.generator parameter referenced from orm.xml is not declared among them.

Once you approve, will prepare a cherry-pick to 7.18.x


Christopher-Chianelli:
Wildfly (the one launched using mvn -N wildfly:start wildfly:deploy) will fail to deploy if built with -Dproductized (doesn't do variable substitution) (Non-blocker). However, the application works without the flags. I verified the integration tests now  work with -Dproductized. Changes look good. Merging.

rsynek:
@Christopher-Chianelli yep, you would need to run the script I included in the PR, or just somehow rewrite the element's value to `true` in standalone.xml.

PR #235, opened 2019-04-09
PLANNER-1442: Add Magic Property that makes MySQL goes faster

Christopher-Chianelli:
Reason: MySQL has a poor AUTO id generator, but it used to have a good one, so
setting the magic property to false increases the speed by a factor of at least
3.


rsynek:
@Fonifan would you please review this PR and check is solves the issue?

rsynek:
@Fonifan thanks for reviewing this!

@Christopher-Chianelli Ok to merge.

PR #238, opened 2019-05-09
RHDM-919: Improved error messages

Christopher-Chianelli:
Modified the backend to return a JSON containing information about the exception (including an i18n Translation Key) when an exception occurs.


employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/exception/RESTException.java
```diff
@@ -0,0 +1,87 @@
...
+
+    private static Exception getExceptionFrom(ServerSideException serverSideException) {
+        Exception out;
+        if (serverSideException.getExceptionCause() != null) {
+            out = new ServerException(getExceptionFrom(serverSideException.getExceptionCause()), serverSideException.getExceptionClass() + ": " + serverSideException.getExceptionMessage());
+        } else {
+            out = new ServerException(serverSideException.getExceptionClass() + ": " + serverSideException.getExceptionMessage());
+        }
+
+        StackTraceElement[] stackTrace = new StackTraceElement[serverSideException.getStackTrace().size()];
```
Fonifan: Probably just nitpicking, but this should be split into separate method for SRP purposes.

employee-rostering-gwtui/src/main/resources/org/optaweb/employeerostering/gwtui/client/resources/i18n/OptaWebUIConstants.properties
```diff
@@ -83,6 +83,11 @@ Rotation.dateHeader=Day {0}
 RotationTimeSelector.rotationDayNumber=Rotation Day #
 RotationTimeSelector.time=Time
 Save=Save
+ServerSideException.entityNotFound=The object you requested was not found: {0}
+ServerSideException.illegalArgument=The parameters you passed in were not valid: {0}
+ServerSideException.nullPointer=It appears the developer made a mistake somewhere. Please contact them.
```
Fonifan: Again minor issue - single developer made a mistake, but you ask user to contact "them". Should be either "developers" or "him".

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/gwtjackson/ServerSideExceptionDeserializer.java
```diff
@@ -0,0 +1,110 @@
...
+        JsonReader reader = new DefaultJsonReader(new StringReader(json));
+        try {
+            return deserialize(reader, json);
+        } finally {
+            reader.close();
+        }
+    }
+
+    public ServerSideException deserialize(JsonReader reader, String json) {
+        Set<String> fields = new HashSet<>(Arrays.asList("i18nKey", "messageParameters", "exceptionClass", "stackTrace", "exceptionMessage", "exceptionCause"));
```
rsynek: Do I get it correctly that this is a list of all fields of the `ServerSideException` class? If so, it's very fragile - there is no way to make sure every time you add a field to the `ServerSideException` you add a corresponding piece of code here.

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/gwtjackson/ServerSideExceptionDeserializer.java
```diff
@@ -0,0 +1,110 @@
...
+        reader.beginObject();
+        while (reader.hasNext()) {
+            String fieldName = reader.nextName();
+            if (!fields.contains(fieldName)) {
+                throw new IllegalArgumentException("Invalid Json: (" + json + ").");
+            }
+            fields.remove(fieldName);
+
+            switch (fieldName) {
+                case "i18nKey":
```
rsynek: The same issue about fragility - if you decide to define names of all `ServerSideException` fields, please do it in a single place and reuse them.

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/gwtjackson/ServerSideExceptionDeserializer.java
```diff
@@ -0,0 +1,110 @@
...
+        List<String> messageParameters = null;
+        String exceptionClass = null;
+        List<String> stackTrace = null;
+        ServerSideException exceptionCause = null;
+
+        reader.beginObject();
+        while (reader.hasNext()) {
+            String fieldName = reader.nextName();
+            if (!fields.contains(fieldName)) {
+                throw new IllegalArgumentException("Invalid Json: (" + json + ").");
```
rsynek: The exception with the same message has been repeated three times => DRY

employee-rostering-shared/src/main/java/org/optaweb/employeerostering/shared/exception/ServerSideException.java
```diff
@@ -0,0 +1,108 @@
...
+        this.exceptionClass = exceptionClass;
+        this.stackTrace = stackTrace;
+        this.exceptionCause = exceptionCause;
+    }
+
+    public ServerSideException(Throwable exception, String i18nKey, String... messageParameters) {
+        this.i18nKey = i18nKey;
+        this.exceptionMessage = exception.getMessage();
+        this.exceptionClass = exception.getClass().getName();
+        this.stackTrace = Arrays.asList(exception.getStackTrace()).stream()
```
rsynek: The list creation is unnecessary. Please see the `Arrays.stream()` method.

employee-rostering-gwtui/src/main/resources/org/optaweb/employeerostering/gwtui/client/resources/i18n/OptaWebUIConstants.properties
Christopher-Chianelli: Will change to "developers", but "them"/"they" can be used as a singular pronoun: https://en.wikipedia.org/wiki/Singular_they

employee-rostering-server/src/main/java/org/optaweb/employeerostering/server/common/jaxrs/OptaWebExceptionMapper.java
```diff
@@ -19,27 +19,60 @@
...
+        } else if (exception instanceof RollbackException) {
+            serverSideException = new ServerSideException(exception, "ServerSideException.rollback");
+        } else {
+            serverSideException = new ServerSideException(exception, "ServerSideException.unknown",
+                                                          exception.getMessage());
+        }
+        try {
+            return objectMapper.writeValueAsString(serverSideException);
+        } catch (JsonProcessingException e) {
+            return "{}";
```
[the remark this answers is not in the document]
Christopher-Chianelli: This is a difficult case: the client is expecting a ServerException JSON object. Trying to serialize the exception could cause another exception (example: objectMapper is null). So we either need to hard code a ServerException JSON object for this case (issue: fragile) or have some way to indicate the error to the client (so we don't try to deserialize the error message as a JSON object). Perhaps I can use different error codes to indicate different cases. WDYT?

employee-rostering-server/src/main/java/org/optaweb/employeerostering/server/common/jaxrs/OptaWebExceptionMapper.java
```diff
@@ -19,27 +19,60 @@
...
+                .type(MediaType.APPLICATION_JSON)
+                .entity(getEntity(exception))
                 .build();
     }
 
+    private String getEntity(final Exception exception) {
+        ObjectMapper objectMapper = objectMapperProvider.getContext(ServerSideException.class);
+        ServerSideException serverSideException;
+        if (exception instanceof IllegalArgumentException) {
+            serverSideException = new ServerSideException(exception, "ServerSideException.illegalArgument",
```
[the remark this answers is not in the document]
Christopher-Chianelli: Might move the different exception types/i18n keys to an enum, and make the enum extract the ServerSideException from the Exception.

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/gwtjackson/ServerSideExceptionDeserializer.java
Christopher-Chianelli: It is suppose to be fragile; having it like this make it fail fast; alerting me that I need to update this piece of code. We don't have reflection on the client, and we need to make sure this code navigate every field of the ServerSideException class, so we have few options. Technically, this class shouldn't be needed, but I am getting a compile error in GWT Devmode (https://github.com/nmorel/gwt-jackson/issues/79). A workaround is to add "dagger-compiler" as a dependency, but I rather not add otherwise unneeded dependencies.

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/gwtjackson/ServerSideExceptionDeserializer.java
Christopher-Chianelli: I can do it in an enum (but then the enum has the issue that it must be kept up to date with the ServerSideException class).

employee-rostering-server/src/main/java/org/optaweb/employeerostering/server/common/jaxrs/OptaWebExceptionMapper.java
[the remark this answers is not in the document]
rsynek: My suggestion was to communicate towards the client that there was an issue, but its serialization failed too - find more information in server log. Basically providing as much information as possible, in a nice format.
Right now if the serialization fails, client gets just an empty JSON object and there is nothing in the server log either.

employee-rostering-server/src/main/java/org/optaweb/employeerostering/server/common/jaxrs/OptaWebExceptionMapper.java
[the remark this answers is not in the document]
rsynek: Enum sounds good, not sure what you mean about the extraction of the ServerSideException. Would you maybe include some gist or update the PR to illustrate it?

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/gwtjackson/ServerSideExceptionDeserializer.java
rsynek: I've seen the issue and agree with you that bringing in another dependency as a workaround is far from ideal solution. Reflection is slightly better, still problematic solution in my opinion. But if you say it cannot be compiled to the client code, we don't have to even talk about this option.

Let's create the enum and reference it from the `ServerSideException` to minimize the maintenance cost. And wherever you need to work with the field, use the enum please. Are you ok with that?

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/gwtjackson/ServerSideExceptionDeserializer.java
rsynek: +1, see my answer in previous comment on this topic

employee-rostering-gwtui/src/main/resources/org/optaweb/employeerostering/gwtui/client/resources/i18n/OptaWebUIConstants.properties
rsynek: @Christopher-Chianelli your call - you're a native speaker

employee-rostering-server/src/main/java/org/optaweb/employeerostering/server/common/jaxrs/OptaWebExceptionMapper.java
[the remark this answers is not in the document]
Christopher-Chianelli: PLAIN_TEXT MediaType is now used when Exception serialization fails (as well as logging the error to the server), and a message telling the user to check the server logs for more details is now shown.

employee-rostering-server/src/main/java/org/optaweb/employeerostering/server/common/jaxrs/OptaWebExceptionMapper.java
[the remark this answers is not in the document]
Christopher-Chianelli: PR Updates; extraction refers to creating the ServerSideExceptionInfo for a given exception (see ExceptionMapping class).

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/pages/rotation/RotationPageViewportBuilder.java
```diff
@@ -143,30 +146,30 @@ private long getCurrentWorkerStartTime() {
...
+                                                                                                                                                                                                                                                                                                                                                                                                                                          ))
+                                                                                                                                                                                                                                                                       ))
+                                                                                                                                                            ))
+                                            ))
```
yurloc: Oops, we have a formatting problem here. I checked the code out and verified that this is actually formatted "correctly" so we need to be creative to reduce the line length. This is my proposal:
```java

        return promiseUtils.promise((res, rej) -> RosterRestServiceBuilder.getRosterState(
                tenantStore.getCurrentTenantId(),
                restCallbackFactory.onSuccess(rosterState -> SpotRestServiceBuilder.getSpotList(
                        tenantStore.getCurrentTenantId(),
                        restCallbackFactory.onSuccess(spotList -> EmployeeRestServiceBuilder.getEmployeeList(
                                tenantStore.getCurrentTenantId(),
                                restCallbackFactory.onSuccess(employeeList -> RotationRestServiceBuilder.getShiftTemplateList(
                                        tenantStore.getCurrentTenantId(),
                                        restCallbackFactory.onSuccess(shiftTemplateViewList -> {
                                                                          RotationView rotationView = new RotationView();
                                                                          rotationView.setTenantId(tenantStore.getCurrentTenantId());
                                                                          rotationView.setRotationLength(rosterState.getRotationLength());
                                                                          rotationView.setSpotList(spotList);
                                                                          rotationView.setEmployeeList(employeeList);
                                                                          Map<Long, List<ShiftTemplateView>> spotIdToShiftTemplateViewListMap = new HashMap<>();
                                                                          spotList.forEach(spot -> spotIdToShiftTemplateViewListMap.put(spot.getId(), new ArrayList<>()));

                                                                          shiftTemplateViewList.forEach(stv -> spotIdToShiftTemplateViewListMap.get(stv.getSpotId()).add(stv));
                                                                          rotationView.setSpotIdToShiftTemplateViewListMap(spotIdToShiftTemplateViewListMap);
                                                                          res.onInvoke(rotationView);
                                                                      }
                                        ))
                                ))
                        ))
                ))
        );
```

employee-rostering-server/src/main/java/org/optaweb/employeerostering/server/common/jaxrs/OptaWebExceptionMapper.java
```diff
@@ -40,39 +37,27 @@
...
-        }
         try {
-            return objectMapper.writeValueAsString(serverSideException);
+            return Response.status(resolveStatus(exception))
+                    .type(MediaType.APPLICATION_JSON)
+                    .entity(getEntity(exception))
+                    .build();
         } catch (JsonProcessingException e) {
-            return "{}";
+            e.printStackTrace();
```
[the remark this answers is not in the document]
Christopher-Chianelli: This was logging the exception to the server, but I changed it to use a Logger (using logger.error(...)) before you reviewed it.

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/pages/rotation/RotationPageViewportBuilder.java
Christopher-Chianelli: That looks pretty similar to what it was before I used the Formatter. Hopefully it doesn't change back again.

yurloc: The main difference is that the two parameters (id+callback) of each service call are always on a new line. That reduces the indentation accumulation. When I hit autoformat in my IDEA it stays unchanged. I try to follow KIE conventions as close as possible so I hope it will work for you as well.

Christopher-Chianelli: Fixed; thanks!

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/common/FailureShownRestCallbackFactory.java
```diff
@@ -0,0 +1,90 @@
...
+        private Consumer<Throwable> onFailure = throwable ->
+                GWT.getUncaughtExceptionHandler().onUncaughtException(throwable);
+
+        @Override
+        public void onError(final Response response) {
+            onError.accept(response);
+        }
+
+        @Override
+        public void onFailure(final Throwable throwable) {
```
rsynek: Would you please explain why is the behaviour of `onError` different than of `onFailure` ?

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/exception/RESTException.java
```diff
@@ -0,0 +1,90 @@
...
+        if (serverSideException.getExceptionCause() != null) {
+            out = new ServerException(getExceptionFrom(serverSideException.getExceptionCause()), serverSideException.getExceptionClass() + ": " + serverSideException.getExceptionMessage());
+        } else {
+            out = new ServerException(serverSideException.getExceptionClass() + ": " + serverSideException.getExceptionMessage());
+        }
+        out.setStackTrace(extractStackTrace(serverSideException));
+        return out;
+    }
+
+    private static StackTraceElement[] extractStackTrace(ServerSideExceptionInfo serverSideException) {
```
rsynek: This method has not been covered by any tests - please see the SonarCloud report.

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/gwtjackson/ServerSideExceptionDeserializer.java
```diff
@@ -0,0 +1,113 @@
...
+
+import com.github.nmorel.gwtjackson.client.stream.JsonReader;
+import com.github.nmorel.gwtjackson.client.stream.JsonToken;
+import com.github.nmorel.gwtjackson.client.stream.impl.DefaultJsonReader;
+import com.github.nmorel.gwtjackson.client.stream.impl.StringReader;
+import org.optaweb.employeerostering.shared.exception.ServerSideExceptionInfo;
+import org.optaweb.employeerostering.shared.exception.ServerSideExceptionInfo.ServerSideExceptionInfoFields;
+
+@Singleton
+public class ServerSideExceptionDeserializer {
```
rsynek: This class has not been covered by any tests - please see the SonarCloud report.

employee-rostering-server/src/main/java/org/optaweb/employeerostering/server/util/HierarchyTree.java
```diff
@@ -0,0 +1,192 @@
...
+
+/**
+ * A data structure representing a hierarchy tree. Each node in the tree represent
+ * a subclass of the Hierarchy. Can be used to find the most specific class
+ * an object is applicable to (example: in a Hierarchy tree containing Throwable and RuntimeException,
+ * IllegalStateException should map to RuntimeException, not Throwable).
+ * @param <K> The type of the key element in the Hierarchy
+ * @param <V> The type of the value element in the Hierarchy
+ */
+public class HierarchyTree<K, V> {
```
rsynek: @Christopher-Chianelli @ge0ffrey was it the plan to fully implement https://issues.jboss.org/browse/PLANNER-933 now? Given we want to rewrite most of the application, we should think twice about adding every such a complex piece of code.

rsynek:
@Christopher-Chianelli one more thing; does the dialog showing a full stacktrace have to display the error icon? Right now there is no space between the picture and text; I think the presence of the picture is questionable.

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/common/FailureShownRestCallbackFactory.java
Christopher-Chianelli: `onError` occurs when we have a Response (which may have a Json Body and if so contain a Server Exception); `onFailure` occurs when we only have a Throwable (and thus cannot extract a Server Exception from it).

employee-rostering-server/src/main/java/org/optaweb/employeerostering/server/util/HierarchyTree.java
Christopher-Chianelli: I think the rewrite apply mostly to the client code (i.e. most of the server code will remain the same). `HierarchyTree` is server code, so the probability of it changing in the rewrite is low.

Christopher-Chianelli:
@rsynek No, I think it was requested for there to be an error icon in the original Error Popup, which the Stack Trace uses. However, I can easily remove/change it.

employee-rostering-gwtui/src/main/java/org/optaweb/employeerostering/gwtui/client/common/FailureShownRestCallbackFactory.java
rsynek: Thank you

rsynek:
> @rsynek No, I think it was requested for there to be an error icon in the original Error Popup, which the Stack Trace uses. However, I can easily remove/change it.

I am not a UX expert, but I cannot see any added value by having the icon there. As the stack trace text is closely attached to the icon, it doesn't look really well.

employee-rostering-gwtui/src/test/java/org/optaweb/employeerostering/gwtui/client/gwtjackson/TestJsonReaderImpl.java
```diff
@@ -0,0 +1,255 @@
...
+
+/**
+ * GWT Jackson cannot compile in tests, so we delegate to Jackson.
+ * <p>
+ * Note: gwt-jackson uses nextToken, which Jackson does not have,
+ * thus we use the convention that the current token in the json parser
+ * is the "next token" (i.e. if the current token in the parser is 10,
+ * then nextInt() should return 10)
+ */
+public class TestJsonReaderImpl implements JsonReader {
```
rsynek: It's a pity there is no common interface (I haven't found any) between gwt-jackson and jackson-core, that would suit our needs. That way you could just switch the implementations.

employee-rostering-server/src/main/java/org/optaweb/employeerostering/server/common/jaxrs/OptaWebExceptionMapper.java
```diff
@@ -16,35 +16,53 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.optaweb.employeerostering.server.exception.ExceptionDataMapper;
+import org.optaweb.employeerostering.server.exception.ExceptionDataMapper.ExceptionData;
+import org.optaweb.employeerostering.shared.exception.ServerSideExceptionInfo;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 @Provider
 public class OptaWebExceptionMapper implements ExceptionMapper<Exception> {
```
rsynek: This class has no test coverage - please take a look at the SonarCloud coverate report. If you have any troubles seeing the report, let me know, I will be happy to help.

Christopher-Chianelli: It is partially covered in the integration test (although we don't inspect the JSON data in the integration test). I can add some JSON inspection tests to the integration tests (as well as create a unit test for it).

PR #245, opened 2019-06-07
DO NOT MERGE. Made a mistake in rebasing, please refer to updated PR in PLANNER-1499-skill

cuijulian:


employee-rostering-backend/src/main/java/org/optaweb/employeerostering/controller/SkillController.java
```diff
@@ -0,0 +1,47 @@
...
+package org.optaweb.employeerostering.controller;
+
+import org.springframework.web.bind.annotation.DeleteMapping;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class SkillController {
```
Christopher-Chianelli: It looks like you are doing controller correctly; however, the paths do no match the old paths, which is a problem for backward compatibility. The paths should be:

```
@GetMapping(/tenantId/{tenantId}/skill/{skillId})
public String getSkill(@PathVariable int tenantId, @PathVariable int skillId)
```

```
@DeleteMapping(/tenantId/{tenantId}/skill/{skillId})
public String removeSkill(@PathVariable int tenantId, @PathVariable int skillId)
```


```
@PostMapping(/tenantId/{tenantId}/skill/update)
public String updateSkill(@PathVariable int tenantId)
```

```
@PostMapping(/tenantId/{tenantId}/skill/add)
public String addSkill(@PathVariable int tenantId)
```

You are also missing the list getter method (the one that return all the skills for a given Tenant):

```
@GetMapping(/tenantId/{tenantId}/skill)
public String getSkillList(@PathVariable int tenantId)
```

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/Skill.java
```diff
@@ -0,0 +1,64 @@
...
+ * limitations under the License.
+ */
+
+package org.optaweb.employeerostering.domain;
+
+import java.util.Objects;
+
+public class Skill extends AbstractPersistable {
+
+    private final Integer tenantId;
```
Christopher-Chianelli: Not Needed; tenantId in `AbstractPersistable`

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/Skill.java
```diff
@@ -0,0 +1,64 @@
...
+    private final String name;
```
Christopher-Chianelli: Name is mutable (that is, a user can update the name of a skill). You may or may not need to change this depending on how updating persisted objects work.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/controller/SkillController.java
cuijulian: Good catch, though note that path names must be enclosed by parentheses. I'll implement the controller methods once I have Skill database persistence going.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/Skill.java
cuijulian: Yup, I'll fix that.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/Skill.java
cuijulian: Another good catch, fixed that in latest commit.

employee-rostering-backend/pom.xml
```diff
@@ -81,12 +81,6 @@
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-test</artifactId>
       <scope>test</scope>
-      <exclusions>
-        <exclusion>
-          <groupId>junit</groupId>
-          <artifactId>junit</artifactId>
-        </exclusion>
-      </exclusions>
```
yurloc: This is not enough. You need to also remove all `*jupiter*` dependencies if you want to use Junit 4. The single test that's currently in the project doesn't run. I think it's because of the jupiter dependencies on classpath.

+1 for keeping Junit 4 during migration. Let's keep the Spring Boot migration scope limited. We'll upgrade to Junit 5 after that.

yurloc: When doing this, please also remove the `pitest-junit5-plugin` dependency from `pitest-maven` plugin in this `pom.xml`. It's only there to make Pitest run with Junit 5 but makes Pitest fail when using Junit 4.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/controller/SkillController.java
```diff
@@ -0,0 +1,54 @@
...
+import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class SkillController {
+
+    @GetMapping("/tenant/{tenantId}/skill")
+    public String getSkillList(@PathVariable Integer tenantId) {
+        return "Get skill list";
```
rsynek: My understanding is that it's not yet implemented - so let's just not forget about it before merging.

Another option is to throw `UnsupportedOperationException` to make sure such an unfinished method won't slip through.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/AbstractPersistable.java
```diff
@@ -0,0 +1,118 @@
...
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.MappedSuperclass;
+import javax.persistence.Version;
+import javax.validation.constraints.NotNull;
+
+import org.optaplanner.core.api.domain.lookup.PlanningId;
+
+@MappedSuperclass
+public abstract class AbstractPersistable implements Serializable {
```
rsynek: May I ask what is the motivation for making the class `Serializable`?

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/controller/SkillController.java
cuijulian: That is correct, I am working on implementing SkillController methods and persistence. I will push those changes once I'm done.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/AbstractPersistable.java
cuijulian: This is for persistence. Java objects will be serialized into JSON format to implement CRUD functionality.

rsynek: Does JPA or Hibernate require making the classes serializable? There are very few reasons to use Java serialization nowadays. I found some interesting article putting some light on this topic: https://bvaisakh.wordpress.com/2015/03/04/do-jpa-entities-have-to-be-serializable

cuijulian: Hmm, it seems that JPA/Hibernate do not require entities to implement Serializable per se. I didn't know that, good to know! From the article: 

> An entity instance is to be passed by value as a detached object (e.g., through a remote interface), the entity class must implement the Serializable interface.

I don't plan on doing this, so it seems Serializable is not required. I'll fix this in my next commit. Good catch, thanks!

PR #246, opened 2019-06-12
PLANNER-1511: Add Spots Page

Christopher-Chianelli:
Depends on #244.

### DIFF
--- a/employee-rostering-frontend/src/store/tenant/tenant.test.ts
+++ b/employee-rostering-frontend/src/store/tenant/tenant.test.ts
@@ -18,6 +18,7 @@ import { mockStore } from '../mockStore';
 import { AppState } from '../types';
 import * as actions from './actions';
 import * as skillActions from '../skill/actions';
+import * as spotActions from '../spot/actions';
 import reducer, { tenantOperations } from './index';
 import {onGet} from 'store/rest/RestServiceClient';
 import Tenant from 'domain/Tenant';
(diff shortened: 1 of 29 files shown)


yurloc:
I think we should remove `DomainObject` and get rid of `tenantId` inherited by all domain objects. Current tenant ID should be stored in Redux store and updated whenever tenant selection changes. Domain objects needn't know about tenants (at least not on frontend). It would also make easier to use `DataTable` by avoiding to have an extra type for the display data structure (`SkillComponents`, `SpotComponents`).

The question is whether to do it now and lightly cross the scope of the rewrite (1:1) or do it after that, which could be a bit more difficult.

Christopher-Chianelli:
@yurloc Right now, the back end includes `tenantId`, `id` and `version` in all rest requests. `id` and `version` are needed; updates and deletes will fail without `id` and `version`. Since `id` and `version` are common to all Domain Objects, I think we should keep `DomainObject`  (it is also useful to have the `id` to update the list without making another rest call, and `ShiftView` references Spots and Employees by their `id`'s).

yurloc:
Yes, `id` is important for referencing. I'm not sure if inheriting it from a common ancestor is the right way to implement it but that's not the point now.

I just noticed that tenant ID is part of the rest call path so including it in the transferred entities is kind of redundant.

Christopher-Chianelli:
@yurloc Related JIRA issue: https://issues.jboss.org/browse/PLANNER-1431 (in which it is proposed to remove tenantId from the REST endpoint)

employee-rostering-frontend/src/store/tenant/tenant.test.ts
```diff
@@ -41,31 +42,69 @@ describe('Tenant operations', () => {
       name: "Tenant 3"
     }];
 
+    let expectedRefreshActions: any[] = [
```
yurloc: My advice: avoid using `any` unless necessary. It's something like `Object` in Java. Actual type of this variable is `(RefreshTenantListAction | RefreshSkillListAction | RefreshSpotListAction)[]` but that can be inferred from the value, so you don't have to specify it explicitly. In most simple cases it's really unnecessary to specify the variable type. For example in `const name: string = "Jiri";` you can safely omit `: string` type annotation because it's obvious that the type of the variable is `string`.

Christopher-Chianelli: I reuse `expectedRefreshActions` later where the first element is an `ChangeTenantAction`, hence the inferred type becomes an issue (and why I use `any`). So I would need to specify the type as `(ChangeTenantAction | RefreshTenantListAction | RefreshSkillListAction | RefreshSpotListAction)[]`.

yurloc: True. I missed the `ChangedTenantAction`. In that case I'd say it's better not to reuse the variable a declare a dedicated one for each set of expected actions.

employee-rostering-frontend/src/ui/components/DataTable.tsx
```diff
@@ -33,50 +33,45 @@ export interface DataTableProps<T> {
 }
 
 interface DataTableState<T> {
-  newRow: IRow | null;
+  newRowData: T|null;
```
yurloc: This is confusing. The state property is being used for two different purposes:

1. As a flag indicating that `Add` button has been clicked and a new entry row should be rendered (correct).

2. To hold the new data. This would be correct if the data were updated using `setState()` but you're allowing a child component (e.g. a `TextInput`) to change (mutate) the data directly without going through `setState()` (see [Using State Correctly](https://reactjs.org/docs/state-and-lifecycle.html#using-state-correctly). It works. I don't deny that. But mixing the responsibilities increases mental capacity required to understand what's going on. And this approach is also against [The Data Flows Down](https://reactjs.org/docs/state-and-lifecycle.html#the-data-flows-down) principle.

    Instead, DataTable should pass the TextInput an onChange() handler function that updates the state with the new value. It can also validate it and set `isValid` flag in the state, which might be a prop of the Save button and you'll get as-you-type validation.

employee-rostering-frontend/src/ui/pages/skill/SkillsPage.tsx
```diff
@@ -51,40 +47,34 @@ const mapDispatchToProps: DispatchProps = {
...
+  editDataRow(data: Skill): JSX.Element[] {
     return [<TextInput key={0} name="name"
       aria-label="Name"
       defaultValue={data.name}
-      onChange={(value) => dataStore.name = value}/>];
-  }
-  
-  isValid(editedValue: SkillComponents): boolean {
-    return editedValue.name.length > 0;
+      onChange={(value) => data.name = value}/>];
```
yurloc: This is the place where the TextInput mutates DataTables state. I think it would be better to call `this.setState(...)` here.

PR #247, opened 2019-06-13
PLANNER-1512: Add Contracts Page

Christopher-Chianelli:
Depends on #246

### DIFF
--- a/employee-rostering-frontend/src/store/contract/contract.test.ts
+++ b/employee-rostering-frontend/src/store/contract/contract.test.ts
@@ -18,8 +18,8 @@ import { mockStore } from '../mockStore';
 import { AppState } from '../types';
 import * as actions from './actions';
 import reducer, { contractOperations } from './index';
-import {withElement, withoutElement, withUpdatedElement} from 'util/ImmutableCollectionOperations';
-import {onGet, onPost, onDelete} from 'store/rest/RestServiceClient';
+import { withElement, withoutElement, withUpdatedElement } from 'util/ImmutableCollectionOperations';
+import { onGet, onPost, onDelete, resetRestClientMock } from 'store/rest/RestTestUtils';
 import Contract from 'domain/Contract';
 
 describe('Contract operations', () => {
@@ -44,7 +44,7 @@ describe('Contract operations', () => {
     expect(client.get).toHaveBeenCalledWith(`/tenant/${tenantId}/contract/`);
 
     store.clearActions();
-    client.get.mockClear();
+    resetRestClientMock(client);
 
     const contractToDelete = mockContractList[0];
     onDelete(`/tenant/${tenantId}/contract/${contractToDelete.id}`, true);
@@ -54,7 +54,7 @@ describe('Contract operations', () => {
     expect(client.delete).toHaveBeenCalledWith(`/tenant/${tenantId}/contract/${contractToDelete.id}`);
 
     store.clearActions();
-    client.delete.mockClear()
+    resetRestClientMock(client);
 
     const contractToAdd: Contract = {tenantId: tenantId,
       name: "Contract 1",
@@ -71,7 +71,7 @@ describe('Contract operations', () => {
     expect(client.post).toHaveBeenCalledWith(`/tenant/${tenantId}/contract/add`, contractToAdd);
 
     store.clearActions();
-    client.post.mockClear()
+    resetRestClientMock(client);
 
     const contractToUpdate: Contract = {tenantId: tenantId,
       name: "Contract 1",
@@ -112,7 +112,7 @@ describe('Contract reducers', () => {
     maximumMinutesPerMonth: 3,
     maximumMinutesPerYear: 4
   };
-  const deletedContract: Contract =       {
+  const deletedContract: Contract = {
     tenantId: 0,
     id: 2,
     version: 0,
--- a/employee-rostering-frontend/src/store/contract/contract.test.ts
+++ b/employee-rostering-frontend/src/store/contract/contract.test.ts
@@ -1,0 +1,138 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { mockStore } from '../mockStore';
+import { AppState } from '../types';
+import * as actions from './actions';
+import reducer, { contractOperations } from './index';
+import {withElement, withoutElement, withUpdatedElement} from 'util/ImmutableCollectionOperations';
+import {onGet, onPost, onDelete} from 'store/rest/RestServiceClient';
+import Contract from 'domain/Contract';
+
+describe('Contract operations', () => {
+  it('should dispatch actions and call client', async () => {
+    const { store, client } = mockStore(state);
+    const tenantId = store.getState().tenantData.currentTenantId;
+    const mockSkillList = [{
+      tenantId: tenantId,
+      id: 0,
+      version: 0,
+      name: "Skill 1"
+    },
+    {
+      tenantId: tenantId,
+      id: 1,
+      version: 0,
+      name: "Skill 2"
+    },
+    {
+      tenantId: tenantId,
+      id: 2,
+      version: 0,
+      name: "Skill 3"
+    }];
+
+    onGet(`/tenant/${tenantId}/skill/`, mockSkillList);
+    await store.dispatch(skillOperations.refreshSkillList());
+    expect(store.getActions()).toEqual([actions.refreshSkillList(mockSkillList)]);
+    expect(client.get).toHaveBeenCalledTimes(1);
+    expect(client.get).toHaveBeenCalledWith(`/tenant/${tenantId}/skill/`);
+
+    store.clearActions();
+    client.get.mockClear();
+
+    const skillToDelete: Skill = {tenantId: tenantId, id: 3214, name: "test"};
+    onDelete(`/tenant/${tenantId}/skill/${skillToDelete.id}`, true);
+    await store.dispatch(skillOperations.removeSkill(skillToDelete));
+    expect(store.getActions()).toEqual([actions.removeSkill(skillToDelete)]);
+    expect(client.delete).toHaveBeenCalledTimes(1);
+    expect(client.delete).toHaveBeenCalledWith(`/tenant/${tenantId}/skill/${skillToDelete.id}`);
+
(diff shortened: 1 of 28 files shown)


employee-rostering-frontend/src/store/contract/operations.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+import { ThunkCommandFactory } from '../types';
+import * as actions from './actions';
+import Contract from 'domain/Contract';
+import { AddContractAction, RemoveContractAction, UpdateContractAction, RefreshContractListAction } from './types';
+
+export const addContract: ThunkCommandFactory<Contract, AddContractAction> = contract =>
+  (dispatch, state, client) => {
+    let tenantId = contract.tenantId;
```
yurloc: Can be `const`.

employee-rostering-frontend/src/ui/components/OptionalInput.tsx
```diff
@@ -0,0 +1,76 @@
...
+  valueMapper: (value: string) => T;
+  valueToString: (value: T) => string;
+  onChange: (value: T|null) => void;
+  label: string;
+}
+
+interface OptionalInputState<T> {
+  inputValue: T|null;
+  currentValue: T|null;
+  isChecked: boolean;
```
yurloc: Can we do without the checkbox? I've tried the UI and seems to be unnecessary. I can imagine that if I delete the value from the field it's enough to make it empty. Why do I have to disable it? Or why do I have to enable it before I enter a value?

Christopher-Chianelli: From https://www.patternfly.org/pattern-library/forms-and-controls/data-input/: "Use an On/Off Switch when a user is presented with the binary choice to enable or disable some optional attribute or capability. " Since Maximum Daily Hours and the like are optional, it make sense to use an On/Off switch to make it clear to the user that they are, in fact, optional.

yurloc: Thanks, that's a valid argument. I'm not entirely convinced this should apply to the inline editable rows but I'm not a UX expert :slightly_smiling_face:. Let's stick with the guidelines unless UX advises otherwise.

employee-rostering-frontend/src/ui/pages/contract/ContractsPage.tsx
```diff
@@ -0,0 +1,140 @@
...
+        defaultValue={data.maximumMinutesPerYear}
+        onChange={value => data.maximumMinutesPerYear = value}
+        isValid={value => /^\d+$/.test(value)}
+        valueMapper={value => parseInt(value)}
+      />
+    ];
+  }
+  
+  isValid(editedValue: Contract): boolean {
+    return editedValue.name.length > 0;
```
yurloc: I think the frontend validation should prevent sending requests that will fail on the backend. This currently isn't true. For example contract name length is limited to 120 chars, hours must be positive. Is this `editedValue.name.length > 0` condition just a place holder and do you plan on implementing full validation later or should it be implemented now?

yurloc: I think the frontend validation should prevent sending requests that will fail on the backend. This currently isn't true. For example contract name length is limited to 120 chars, hours must be positive. Is this `editedValue.name.length > 0` condition just a place holder and do you plan on implementing full validation later or should it be implemented now?

employee-rostering-frontend/src/store/contract/contract.test.ts
```diff
@@ -0,0 +1,205 @@
...
+    tenantId: 0,
+    id: 1,
+    version: 0,
+    name: "Updated Contract 2",
+    maximumMinutesPerDay: 1,
+    maximumMinutesPerWeek: 2,
+    maximumMinutesPerMonth: 3,
+    maximumMinutesPerYear: 4
+  };
+  const deletedContract: Contract =       {
```
yurloc: Extra spaces.

employee-rostering-frontend/src/ui/pages/contract/ContractsPage.tsx
Christopher-Chianelli: Agreed; plan on implementing full validation later (probably in the Employees Page PR, which will also include a DataTable refactor adding pagination, filters and sorting)

employee-rostering-frontend/src/ui/components/OptionalInput.tsx
```diff
@@ -0,0 +1,76 @@
...
+  isValid: (value: string) => boolean;
+  valueMapper: (value: string) => T;
+  valueToString: (value: T) => string;
+  onChange: (value: T|null) => void;
+  label: string;
+}
+
+interface OptionalInputState<T> {
+  inputValue: T|null;
+  currentValue: T|null;
```
yurloc: The state property `currentValue` is unused. Reported by an ESLint rule [react/no-unused-state](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unused-state.md) that I'm trying out locally at the moment.

employee-rostering-frontend/src/ui/components/OptionalInput.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+    optionalInput.setState = jest.fn();
+
+    optionalInput.handleToggle(false);
+    expect(optionalInput.setState).toBeCalled();
+    expect(optionalInput.setState).toBeCalledWith({currentValue: null, isChecked: false});
+    expect(dataProps.onChange).toBeCalled();
+    expect(dataProps.onChange).toBeCalledWith(null);
+  });
+
+  it('should be load the old value after a toggle', () => {
```
yurloc: Extra word: "be".

employee-rostering-frontend/src/store/contract/contract.test.ts
```diff
@@ -0,0 +1,205 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { mockStore } from '../mockStore';
+import { AppState } from '../types';
+import * as actions from './actions';
+import reducer, { contractOperations } from './index';
+import {withElement, withoutElement, withUpdatedElement} from 'util/ImmutableCollectionOperations';
+import {onGet, onPost, onDelete} from 'store/rest/RestServiceClient';
```
yurloc: TypeScript compiler complains that
```
TS2305: Module '"../rest/RestServiceClient"' has no exported member 'onGet'.
```
Can you fix it? It seems that simply importing from `store/rest/__mocks__/RestServiceClient` fixes it. It's the easiest solution. Not sure if the correct and best one but could be.

employee-rostering-frontend/src/store/contract/contract.test.ts
```diff
@@ -0,0 +1,205 @@
...
+    ).toEqual({contractList: withUpdatedElement(state.contractList.contractList, updatedContract)})
+  });
+  it('refresh contract list', () => {
+    expect(
+      reducer(state.contractList, actions.refreshContractList([addedContract])),
+    ).toEqual({contractList: [addedContract]});
+  });
+});
+
+const state: AppState = {
```
yurloc: TypeScript compiler:
```
TS2741: Property 'contractList' is missing in type '{ tenantData: { currentTenantId: number; tenantList: never[]; }; spotList: { spotList: never[]; }; skillList: { skillList: { tenantId: number; id: number; version: number; name: string; }[]; }; }' but required in type 'AppState'.
```

employee-rostering-frontend/src/store/contract/contract.test.ts
```diff
@@ -0,0 +1,205 @@
...
+    }];
+
+    onGet(`/tenant/${tenantId}/contract/`, mockContractList);
+    await store.dispatch(contractOperations.refreshContractList());
+    expect(store.getActions()).toEqual([actions.refreshContractList(mockContractList)]);
+    expect(client.get).toHaveBeenCalledTimes(1);
+    expect(client.get).toHaveBeenCalledWith(`/tenant/${tenantId}/contract/`);
+
+    store.clearActions();
+    client.get.mockClear();
```
yurloc: This also causes a compiler error. Not sure what's the best way to fix.

employee-rostering-frontend/src/store/contract/contract.test.ts
Christopher-Chianelli: Changing it to that breaks the test. I am guessing the errors in the test are due to the TypeScript compiler in the IDE not being aware that the classes are being mocked. So we might need to change it in a compiler setting, or move the methods to a new file `RestTestUtils` (which would be strongly coupled with store/rest/__mocks__/RestServiceClient).

employee-rostering-frontend/src/store/contract/contract.test.ts
Christopher-Chianelli: Did you do it after updating the import to  `store/rest/__mocks__/RestServiceClient`? I can reproduce the error only after updating the import to `store/rest/__mocks__/RestServiceClient`; no errors without updating the import.

Christopher-Chianelli: Seems you commented on the wrong file; `tenant.test` is the one with the issue.

employee-rostering-frontend/src/store/contract/contract.test.ts
Christopher-Chianelli: `npm lint` does not catch that (nor does it update the spacing in packages).

employee-rostering-frontend/src/ui/components/OptionalInput.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  it('should be enabled initially if default value is not null', () => {
+    const optionalInput = new OptionalInput<MockData>(dataProps);
+    expect(optionalInput.state.isChecked).toEqual(true);
+    expect(optionalInput.state.inputValue).toEqual(dataProps.defaultValue);
+  });
+
+  it('should call onChange when input changes if valid', () => {
+    const optionalInput = mount(<OptionalInput {...dataProps}/>);
+    dataProps.isValid.mockReturnValue(true);
```
yurloc: On last compiler error.  Casting to `jest.Mock` solves it.

employee-rostering-frontend/src/ui/components/OptionalInput.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+    expect(dataProps.isValid).toBeCalledWith('A');
+    expect(dataProps.valueMapper).toBeCalled()
+    expect(dataProps.valueMapper).toBeCalledWith('A');
+    expect(dataProps.onChange).toBeCalled();
+    expect(dataProps.onChange).toBeCalledWith({name: 'A'});
+  });
+
+  it('should not call onChange when input changes if invalid', () => {
+    const optionalInput = mount(<OptionalInput {...dataProps}/>);
+    dataProps.isValid.mockReturnValue(false);
```
yurloc: On last compiler error.  Casting to `jest.Mock` solves it.

employee-rostering-frontend/src/ui/components/OptionalInput.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+    dataProps.onChange.mockClear();
```
yurloc: On last compiler error.  Casting to `jest.Mock` solves it.

PR #248, opened 2019-06-14
PLANNER-1513: Add Employees Page, Refactored DataTable

Christopher-Chianelli:
Depends on #247.


employee-rostering-frontend/src/ui/pages/employee/EmployeesPage.tsx
```diff
@@ -0,0 +1,123 @@
...
+        onChange={(value) => setProperty("name", value)}
+      />,
+      <TypeaheadSelectInput key={1}
+        emptyText={"Select contract"}
+        options={this.props.contractList}
+        optionToStringMap={contract => contract.name}
+        defaultValue={data.contract}
+        onChange={contract => setProperty("contract", contract)}/>,
+      <MultiTypeaheadSelectInput key={2}
+        emptyText={"Select required skills"}
```
yurloc: For employee it's not required skills, it's just skills, right?

yurloc: Or "skill proficiencies" as the column title says.

yurloc:
Adding skills/spots/etc. is broken on this PR.

employee-rostering-frontend/src/store/contract/operations.ts
```diff
@@ -0,0 +1,57 @@
...
+import * as actions from './actions';
+import Contract from 'domain/Contract';
+import { AddContractAction, RemoveContractAction, UpdateContractAction, RefreshContractListAction } from './types';
+import { employeeOperations } from 'store/employee';
+
+export const addContract: ThunkCommandFactory<Contract, AddContractAction> = contract =>
+  (dispatch, state, client) => {
+    let tenantId = contract.tenantId;
+    return client.post<Contract>(`/tenant/${tenantId}/contract/add`, contract).then(newContract => {
+      dispatch(actions.addContract(newContract))
```
yurloc: Actually, you can simply return the action here instead of dispatching it. You'd need to create an new type for the factory method (something like `ThunkActionFactory`) but that's expected. ThunkAction is a function that returns `Promise<SomeAction>`. I use ThunkCommand (which is void) in VRP because sending a message over WebSocket has no direct response, so there's no response data to create an action from.

employee-rostering-frontend/src/store/contract/reducers.ts
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ActionType, ContractList, ContractAction } from './types';
+import {withElement, withoutElement, withUpdatedElement} from 'util/ImmutableCollectionOperations';
```
yurloc: Would it be possible to use the spacing consistently? For destructuring (`const { name, age } = this.props.person;`), object literals (`const person = { name: "Joe", age: 5 }`) and ES6 imports, spaces around the curly braces are preferred.

employee-rostering-frontend/src/ui/components/DataTable.tsx
```diff
@@ -46,26 +49,25 @@ export abstract class DataTable<T, P extends DataTableProps<T>> extends React.Co
...
+  abstract isValid(editedValue: ReadonlyPartial<T>): editedValue is T;
 
   abstract updateData(data: T): void;
   abstract addData(data: T): void;
   abstract removeData(data: T): void;
 
   createNewRow() {
     if (this.state.newRowData === null) {
-      this.setState({...this.state, newRowData: this.createNewDataInstance()});
+      this.setState({...this.state, newRowData: {}});
```
yurloc: I think that spreading the state here is unnecessary as state updates are being merged https://reactjs.org/docs/state-and-lifecycle.html#state-updates-are-merged.

employee-rostering-frontend/src/store/contract/reducers.ts
Christopher-Chianelli: Didn't run linter yet as PR is a WIP; the linter might fix it

yurloc: True. Let's fix #247 first.

employee-rostering-frontend/src/store/contract/operations.ts
yurloc: I think we can take a look at this later after this PR.

yurloc:
As a user, I would really appreciate if filtering was case insensitive. So when searching for Amy Fox I can simply type `amy` without worrying about capital letters. Applies to all tables.

Christopher-Chianelli:
Side note: `npm run lint` was broken; needed to `npm install --save eslint-config-airbnb` to fix it.

employee-rostering-frontend/src/store/employee/employee.test.ts
```diff
@@ -0,0 +1,219 @@
...
+    const mockEmployeeList: Employee[] = [mockEmployee];
+
+    onGet(`/tenant/${tenantId}/employee/`, mockEmployeeList);
+    await store.dispatch(employeeOperations.refreshEmployeeList());
+    expect(store.getActions()).toEqual([actions.refreshEmployeeList(mockEmployeeList)]);
+    expect(client.get).toHaveBeenCalledTimes(1);
+    expect(client.get).toHaveBeenCalledWith(`/tenant/${tenantId}/employee/`);
+
+    store.clearActions();
+    resetRestClientMock(client);
```
yurloc: I suggest to split the test so that each use case (get/delete/post etc.) has its own test method. This will make the test more readable and isolated so it's not necessary to this kind of clean up.

employee-rostering-frontend/src/store/employee/employee.test.ts
```diff
@@ -0,0 +1,219 @@
...
+import {onGet, onPost, onDelete, resetRestClientMock} from 'store/rest/RestTestUtils';
+import Employee from 'domain/Employee';
+
+describe('Employee operations', () => {
+  it('should dispatch actions and call client', async () => {
+    const { store, client } = mockStore(state);
+    const tenantId = store.getState().tenantData.currentTenantId;
+    
+    const mockEmployee: Employee = {
+      tenantId: tenantId,
```
yurloc: What is the actual tenant ID? It would be better if it was a constant *defined* by the test, independent on some default values. It's probably zero but would the test work if we used a non-default value?

Christopher-Chianelli: I defined it in the test; the initial state is defined at the `const state = {...}` at the bottom of the test.

employee-rostering-frontend/src/store/employee/employee.test.ts
Christopher-Chianelli: I modelled it after the `optaweb-vehicle-routing` Routes test (which have all operations in a single test): https://github.com/kiegroup/optaweb-vehicle-routing/blob/master/optaweb-vehicle-routing-frontend/src/store/route/route.test.ts. Probably best to update that one too.

employee-rostering-frontend/src/ui/components/OptionalInput.tsx
```diff
@@ -67,6 +67,9 @@ export default class OptionalInput<T> extends React.Component<OptionalInputProps
                 this.props.onChange(mappedValue);
               }
             }
+            else {
+              this.props.onChange(undefined);
+            }
           }}
         />
         <Switch aria-label="Enabled" isChecked={isChecked} onChange={this.handleToggle} />
```
MusaTalluzi: This component isn't necessary, and it's use is extra work for users (they have to toggle the switch for every field they want to fill). Most (if not all) input components have `isRequired` property, see [patternfly's TextInput](https://www.patternfly.org/v4/documentation/react/components/textinput/#textinput-props).

employee-rostering-frontend/src/store/contract/contract.test.ts
```diff
@@ -17,9 +17,10 @@
 import { mockStore } from '../mockStore';
 import { AppState } from '../types';
 import * as actions from './actions';
+import * as employeeActions from 'store/employee/actions';
 import reducer, { contractOperations } from './index';
 import { withElement, withoutElement, withUpdatedElement } from 'util/ImmutableCollectionOperations';
-import { onGet, onPost, onDelete, resetRestClientMock } from 'store/rest/RestTestUtils';
+import {onGet, onPost, onDelete, resetRestClientMock} from 'store/rest/RestTestUtils';
```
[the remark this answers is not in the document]
Christopher-Chianelli: Oddly enough, `npm lint: fix` doesn't fix the bracket spacing inconsistency (it does fix nearly all other spacing inconsistencies though).

employee-rostering-frontend/src/ui/components/DataTable.tsx
```diff
@@ -141,55 +186,111 @@ export abstract class DataTable<T, P extends DataTableProps<T>> extends React.Co
...
+    const rows = additionalRows.concat(rowsOnPage);
+
+    const columns: ICell[] = this.props.columnTitles.map((t, index) => {
+      return { title: t, cellTransforms: [headerCol], props: {}, transforms: (sorters[index] !== null) ? [sortable] : undefined };
+    }).concat([{ title: '', cellTransforms: [headerCol], props: {}, transforms: undefined }]);
     return (
-      <div>
-        <Button isDisabled={this.state.newRowData !== null} onClick={this.createNewRow}>Add</Button>
-        <Table caption={this.props.title} cells={columns} rows={rows}>
+      <Card>
```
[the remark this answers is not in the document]
Christopher-Chianelli: It make the background of the search bar and pagination consistent consistent with the table (that is, it make it look like the search bar and pagination belong to the table). A plain `div` make it appears the pagination and search bar belong to the page, disjoint from the table and on a grayish background.

employee-rostering-frontend/src/ui/components/OptionalInput.tsx
Christopher-Chianelli: Patternfly Design Docs has two views on this: https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/forms. The first is to mark required fields with an "*" or put a message "All fields are optional" by the form. The second is to have a checkbox that enables an field.

employee-rostering-frontend/src/store/contract/contract.test.ts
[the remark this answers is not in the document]
MusaTalluzi: You need to add [object-curly-spacing](https://eslint.org/docs/rules/object-curly-spacing) rule to your `eslint.rc`

employee-rostering-frontend/src/ui/components/OptionalInput.tsx
MusaTalluzi: So in the contract case, only the name is required and others are optional, so you can add `isRequired` to the name TextInput which will display an asterisk on top of it and leave the rest as TextInput without `isRequired`, they'll be optional by default.
I didn't quite get the second view, a required input isn't up to the user to specify, developers decide whether a field is required or not and force users to fill it if so.

employee-rostering-frontend/src/ui/components/DataTable.tsx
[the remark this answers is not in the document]
MusaTalluzi: Makes sense, although I don't think Card should be used to solve this, but I can't think of an alternative either.

employee-rostering-frontend/src/store/employee/employee.test.ts
yurloc: Thanks. I missed that `mockStore(state)` accepts the initial state.

employee-rostering-frontend/src/store/contract/contract.test.ts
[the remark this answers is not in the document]
yurloc: @MusaTalluzi I'd recommend not to introduce new rule configurations as long as we can reuse ones that are part of Airbnb style. So the idea is to use the complete Airbnb style and only override minimal set of rules that are inconvenient for some reason. The groundwork has been laid in #251. Currently, `airbnb` is first and other configs override it. I chose this approach not to introduce too many rules at a time since Christopher has a lot of work in progress.

employee-rostering-frontend/src/store/employee/employee.test.ts
yurloc: Done: kiegroup/optaweb-vehicle-routing/pull/126.

employee-rostering-frontend/src/ui/components/DataTable.tsx
```diff
@@ -27,39 +39,76 @@ export interface DataTableProps<T> {
...
   currentFilter: (rowData: T) => boolean;
+  sortBy: ISortBy;
 }
 
+export type ReadonlyPartial<T> = { readonly [P in keyof T]?: T[P] };
+export type PropertySetter<T> = (propertyName: keyof T, value: T[keyof T] | undefined) => void;
+export type Sorter<T> = (a: T, b: T) => number;
+
 export abstract class DataTable<T, P extends DataTableProps<T>> extends React.Component<P, DataTableState<T>> {
+  [x: string]: any;
```
yurloc: What's this?

employee-rostering-frontend/src/store/employee/operations.ts
```diff
@@ -0,0 +1,55 @@
...
+      dispatch(actions.addEmployee(newEmployee))
+    });
+  };
+
+export const removeEmployee: ThunkCommandFactory<Employee, RemoveEmployeeAction> = employee =>
+  (dispatch, state, client) => {
+    let tenantId = employee.tenantId;
+    let employeeId = employee.id;
+    return client.delete<boolean>(`/tenant/${tenantId}/employee/${employeeId}`).then(isSuccess => {
+      if (isSuccess) {
```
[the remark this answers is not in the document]
Christopher-Chianelli: The unsuccessfully delete will probably be caught by the axios client (as in, it won't enter this method since the rest call didn't return 200). +1 for adding TODO to all the places that need error handling.

employee-rostering-frontend/src/ui/components/DataTable.tsx
Christopher-Chianelli: I have no idea; will remove.

employee-rostering-frontend/src/ui/components/DataTable.test.tsx
```diff
@@ -17,16 +17,20 @@ import { mount, shallow } from 'enzyme';
...
 
 interface MockData {name: string; number: number}
 class MockDataTable extends DataTable<MockData, DataTableProps<MockData>> {
-  createNewDataInstance = jest.fn(() => ({name: '', number: 0}));
   displayDataRow = jest.fn((data) => [<span key={0} id="viewer">{data.name}</span>,
     <span key={1}>{data.number}</span>]);
+  getInitialStateForNewRow = jest.fn(() => ({}));
   editDataRow = jest.fn((data) => [<input key={0} id="editor" />,
     <input key={1} />]);
+  // @ts-ignore
```
yurloc: It seems this a result of `isValid()` mixing responsibilities. It does two different things:
1. it checks that edited value contains valid data and
2. it is a type predicate asserting that `editedValue: ReadonlyPartial<T>` is in fact a complete `T`.

Could you split these two functions into:
- `isT(editedValue: ReadonlyPartial<T>): editedValue is T;` and
- `isValid(data: T): boolean`?

Christopher-Chianelli: I was thinking of refactoring the Validation logic where each column has its own list of validators (which allow us to display feedback on the input component causing the error). So I will probably add a `isDataComplete(editedValue: ReadonlyPartial<T>): editedValue is T;` and a `getValidatorLists(): Validator[][]` (two-dimensional since each column can have multiple validators with different error messages for each one).

employee-rostering-frontend/src/store/contract/contract.test.ts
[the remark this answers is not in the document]
MusaTalluzi: @yurloc `object-curly-spacing` is required in [airbnb style](https://github.com/airbnb/javascript#whitespace--in-braces) but it's probably disabled by some subsequent rule/extension.

employee-rostering-frontend/src/ui/components/DataTable.test.tsx
Christopher-Chianelli: Did a refactoring of validation logic (not pushed yet). I used `formsy-react` (https://github.com/formsy/formsy-react) (2.0 beta) and `validator` (https://github.com/validatorjs/validator.js) to validate the row and provide feedback. This is done by wrapping validated inputs into a `Validated` element, which attaches the validators.

employee-rostering-frontend/src/store/contract/contract.test.ts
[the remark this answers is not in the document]
yurloc: Yes. That's what I'm saying :slightly_smiling_face:

employee-rostering-frontend/src/ui/components/DataTable.test.tsx
yurloc: Yes. Per-column validation sounds like a good idea, although I'd says it's the next level. Could you do this in a separate PR (unless it actually reduces the PR size :slightly_smiling_face:)?

employee-rostering-frontend/src/ui/components/DataTable.tsx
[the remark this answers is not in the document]
Christopher-Chianelli: Used a `Level` component and set the background color to a PattenFly global color

Christopher-Chianelli:
@MusaTalluzi your concerns about validations/user feedback will be addressed in another PR to keep this PR size manageable. Your other concerns were addressed by my latest commit.

employee-rostering-frontend/src/ui/components/DataTable.tsx
```diff
@@ -73,8 +71,10 @@ export abstract class DataTable<T, P extends DataTableProps<T>> extends React.Co
 
   abstract getInitialStateForNewRow(): Partial<T>;
   abstract displayDataRow(data: T): JSX.Element[];
-  abstract editDataRow(data: ReadonlyPartial<T>, setProperty: (propertyName: keyof T, value: T[keyof T] | undefined) => void): JSX.Element[];
-  abstract isValid(editedValue: ReadonlyPartial<T>): editedValue is T;
+  abstract editDataRow(data: ReadonlyPartial<T>,
+    setProperty: (propertyName: keyof T, value: T[keyof T] | undefined) => void): JSX.Element[];
```
yurloc: Can now be `setProperty: PropertySetter<T>`.

employee-rostering-frontend/src/types.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+export type Predicate<T> = (value: T) => boolean;
+export type ReadonlyPartial<T> = { readonly [P in keyof T]?: T[P] };
+export type Sorter<T> = (a: T, b: T) => number;
```
yurloc: Good idea! Just please fix imports in tests. Some of them still import the types from the TSX components where they where originally introduced. Running `tsc` from command line can help identify places with TypeScript compilation errors. I have no idea why `tsc` gives different results (fails) than running `npm run build` or `npm test` (so basically react-scripts). There must be some kind of discrepancy between plain `tsconfig.json` and `react-scripts`.

yurloc:
The failing build - I'm convinced it is a bug in `npm`. Travis uses npm 6.4.1 while we all probably have installed the latest version 6.9.0. The easiest solution seems to be to force the same, latest version on Travis too. It can be done by adding:
```yaml
before_install:
  - npm i -g npm@6.9.0
```
to `.travis.yml`. Found this in [Travis docs](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#using-npm).

employee-rostering-frontend/src/types.ts
Christopher-Chianelli: Running `tsc` from the command line complains about the `node_modules` folder, and doesn't complain about the wrong imports. I found the locations via a find operation.

employee-rostering-frontend/src/ui/components/TypeaheadSelectInput.tsx
```diff
@@ -0,0 +1,113 @@
...
+          aria-label={emptyText}
+          onToggle={this.onToggle}
+          onSelect={this.onSelect}
+          onClear={this.clearSelection}
+          selections={selection}
+          isExpanded={isExpanded}
+          ariaLabelledBy={titleId}
+          placeholderText={emptyText}
+        >
+          <SelectOption isDisabled={false} key={-1} value="" />
```
yurloc: @Christopher-Chianelli What's the purpose of this option?

Christopher-Chianelli: Will probably remove in #263; I believe I used it as a placeholder element.

PR #249, opened 2019-06-14
[PLANNER-1499] Move Skill domain object to SpringBoot

cuijulian:


employee-rostering-backend/pom.xml
```diff
@@ -55,90 +54,45 @@
...
-    </dependency>
-    <dependency>
-      <groupId>com.fasterxml.jackson.core</groupId>
-      <artifactId>jackson-databind</artifactId>
-      <version>2.9.9</version>
-    </dependency>
-    <!-- Enable Spring Boot Automatic Restart, see Development Guide in Readme to learn how to use it. -->
-    <dependency>
-      <groupId>org.springframework.boot</groupId>
-      <artifactId>spring-boot-devtools</artifactId>
```
yurloc: You'll probably want to keep this. Provides nice developer experience https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#using-boot-devtools-restart.

yurloc: But feel free to remove it and add it later when you really want to start using it.

employee-rostering-backend/pom.xml
```diff
@@ -55,90 +54,45 @@
...
-      <optional>true</optional>
-    </dependency>
-    <!-- Include annotation processor that generates metadata for @ConfigurationProperties -->
-    <dependency>
-      <groupId>org.springframework.boot</groupId>
-      <artifactId>spring-boot-configuration-processor</artifactId>
```
yurloc: This will be helpful when you start using configuration properties. Again, feel free to remove it until it's really needed.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/controller/SkillController.java
```diff
@@ -0,0 +1,67 @@
...
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class SkillController {
+
+    @Autowired
```
yurloc: Spring Team recommends: "Always use constructor based dependency injection in your beans. Always use assertions for mandatory dependencies".

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/controller/SkillController.java
```diff
@@ -0,0 +1,67 @@
...
+    private SkillService skillService;
```
yurloc: Can be `final` if you use constructor dependency injection.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/skill/SkillService.java
```diff
@@ -0,0 +1,62 @@
...
+import org.optaweb.employeerostering.domain.Skill;
+import org.optaweb.employeerostering.persistence.SkillRepository;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service
+public class SkillService {
+
+    @Autowired
+    private SkillRepository skillRepository;
```
yurloc: Constructor injection.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/skill/SkillService.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public List<Skill> getSkillList(Integer tenantId) {
+        return skillRepository.findAll();
+    }
+
+    public Skill getSkill(Integer tenantId, Long id) {
+        if (skillRepository.findById(id).isPresent())
```
yurloc: Missing braces around `if` body (checkstyle fails).

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/skill/SkillService.java
```diff
@@ -0,0 +1,62 @@
...
+            skillRepository.deleteById(id);
+    }
+
+    public void createSkill(Integer tenantId, Skill skill) {
+        skillRepository.save(skill);
+    }
+
+    public void updateSkill(Integer tenantId, Skill skill) {
+        if (skill.getId() != null) {
```
yurloc: Soft suggestion: `getId()`'s contract should be defined so that it never returns `null`. Then this check can be removed. If we can't guarantee that it never returns null, it's better to return `Optional<Long>` so that it's harder to avoid the check for presence of the value.

If the `id` is null or empty `Optional` the method should throw `IllegalArgumentException`.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/skill/SkillService.java
```diff
@@ -0,0 +1,62 @@
...
+            if (skillRepository.findById(skill.getId()).isPresent()) {
```
yurloc: What happens if a skill with that id is not found? This case should be covered by a test.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/skill/SkillService.java
```diff
@@ -0,0 +1,62 @@
...
+                Skill skill1 = skillRepository.findById(skill.getId()).get();
+
+                skill1.setTenantId(skill.getTenantId());
```
yurloc: Is this a legal operation? Does the domain allow any entity to move from one tenant to another? Please discuss this with @Christopher-Chianelli. The answer strongly depends on the definition of tenant. If the domain treats tenants as completely separated worlds, it's likely not possible to move entities between tenants. In that case making a call to the service with `tenantId` different from `skill.tenantId` would be an invalid request. In such case, write a defensive code and fail immediately. And don't forget about tests :slightly_smiling_face:

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/skill/SkillService.java
```diff
@@ -0,0 +1,62 @@
...
+                skill1.setName(skill.getName());
```
yurloc: Don't forget to save the change :slightly_smiling_face: (sorry if I'm stating the obvious). And don't forget to write a test that verifies it.

yurloc: One more thing, ideally when writing such test, don't go through the controller which needs the whole web layer. Look for this https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/test/autoconfigure/orm/jpa/DataJpaTest.html.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/skill/SkillService.java
cuijulian: From my understanding, all domain objects that inherit from `AbstractPersistable` (including Skill) will have `id = null` until it's persisted in a database. I can't see why `updateSkill()` would be called on a Skill object that's not persisted. I was planning to throw an `EntityNotFoundException` for the null case, but returning `Optional<Long>` may be better. What are your thoughts on this?

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/skill/SkillService.java
cuijulian: I plan to throw an `EntityNotFoundException` if the skill was not found. Would that be sufficient?

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/skill/SkillService.java
cuijulian: I did not consider this, thank you for the reminder. I will discuss this with Chris.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/skill/SkillService.java
rsynek: I would say it depends on whether you consider calling `getId()` on an object that has not yet been persisted (and flushed) an exceptional - or rather illegal - case.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/skill/SkillService.java
rsynek: @cuijulian you can get rid of such issues by importing the KIE code style in your IDE https://github.com/kiegroup/droolsjbpm-build-bootstrap/tree/master/ide-configuration

employee-rostering-backend/pom.xml
```diff
@@ -55,90 +54,45 @@
...
     <java.version>1.8</java.version>
-    <version.org.jacoco>0.8.3</version.org.jacoco>
   </properties>
 
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.optaplanner</groupId>
+        <artifactId>optaplanner-bom</artifactId>
+        <version>7.22.0.Final</version>
```
rsynek: Is there a reason why this particular version has been used? I would expect 7.24.0-SNAPSHOT, or if we want to use a final version, then use the latest one - 7.23.0.Final.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/AbstractPersistable.java
```diff
@@ -0,0 +1,116 @@
...
+
+import org.optaplanner.core.api.domain.lookup.PlanningId;
+
+@MappedSuperclass
+public abstract class AbstractPersistable {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO)
+    @PlanningId
+    protected Long id;
```
rsynek: For the same of encapsulation, this one and another fields below could be declared as private and each descendent class would use getters to access them.

employee-rostering-backend/src/test/java/org/optaweb/employeerostering/controller/SkillControllerTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Autowired
+    private MockMvc mvc;
+
+    @Test
+    public void getSkillListTest() throws Exception {
+        mvc.perform(MockMvcRequestBuilders
+                            .get("/tenant/{tenantId}/skill", 0)
+                            .accept(MediaType.APPLICATION_JSON))
+                .andDo(print())
```
rsynek: With a growing number of tests, increasing console output will become less and less comprehensible. Generally, you want each test to either pass or fail (with a self-explanatory failure message). Such an output should be used only for debugging purposes, so I would recommend using `DEBUG` level logging instead of printing to console output.

employee-rostering-backend/src/test/java/org/optaweb/employeerostering/controller/SkillControllerTest.java
```diff
@@ -0,0 +1,94 @@
...
+                .andExpect(status().isOk());
```
rsynek: Checking status is necessary part of every integration test focused on REST API/HTTP communication. However, there is no verification of the `Skill` list that has been retrieved.

Similarly, other tests in this class do only status checking too.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/skill/SkillService.java
yurloc: @cuijulian Ahh, I see. Good point. So we can't have `final id` at the moment. This is a result of mixing domain model with persistence infrastructure. But this problem exists in the original code so I don't think fixing this is in scope of the migration to Spring Boot. Let's address this later.

If you're interested in seeing how it works when domain model is independent and unaffected by persistence layer, see
- [Location](https://github.com/kiegroup/optaweb-vehicle-routing/blob/master/optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/domain/Location.java) (the domain model class, an entity object that has an identity represented by an unmodifiable id). It's true that it doesn't have an id before it's persisted. And it's because the *entity* doesn't *exist* before being persisted. In other words, the entity object's life cycle starts with being persisted and obtaining a unique id. Before that, everything we have is the user's request to create a new Location entity with the data necessary to create it (an id is not among them).
- [LocationRepository](https://github.com/kiegroup/optaweb-vehicle-routing/blob/master/optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/persistence/LocationRepositoryImpl.java#L44)'s `createLocation()` method. Notice that before the Location is created, we only have the necessary *data* from the client (coordinates and description). Also notice that `LocationEntity` is not *the entity* from domain model perspective (I named it so to honor JPA conventions but I might change that since it's confusing). The `LocationEntity` is a plumbing object that only servers to communicate with the CRUD repository. It's mutable but we don't mind since it never leaks outside the repository API so it doesn't affect the rest of the system, where we mostly work with the domain model where Location is a safe, immutable object that never returns `nulls`.

The vehicle routing code is inspired by [Clean Architecture](https://blog.cleancoder.com/uncle-bob/2012/08/13/the-clean-architecture.html) ([book](https://www.oreilly.com/library/view/clean-architecture-a/9780134494272/) mildly recommended) and [Domain Driven Design](https://www.oreilly.com/library/view/domain-driven-design-tackling/0321125215/) (recommended).

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/skill/SkillService.java
yurloc: Yes that's a good, defensive approach.

employee-rostering-backend/pom.xml
yurloc: I think the non-latest version is not a problem at the moment. Master is still on snapshot so if there's a change in optaplanner that breaks employee rostering, we'll know about it.

During the migration it will be convenient to use a final version so that you don't wait for artifacts download each morning.

employee-rostering-backend/pom.xml
```diff
@@ -364,96 +243,6 @@
...
-        <sonar.pullrequest.key>${env.TRAVIS_PULL_REQUEST}</sonar.pullrequest.key>
-      </properties>
-    </profile>
   </profiles>
 
-  <repositories>
-    <repository>
-      <id>jboss-public-repository-group</id>
-      <name>JBoss Public Repository Group</name>
-      <url>https://repository.jboss.org/nexus/content/groups/public/</url>
```
yurloc: FYI, you'll need this repository once you change your optaplanner dependency to a snapshot version. JBoss snapshot artifacts are not available in Maven Central repository.

You don't need to declare the repository here if you have `optaweb-employee-rostering` as the parent of this POM (which you currently don't). Its parent is `kie-parent` where this repository is declared, so you'll inherit it from there. If you want to switch the parent from `spring-boot-starter-parent` to `kie-parent` (which is a good idea), follow https://www.baeldung.com/spring-boot-dependency-management-custom-parent.

employee-rostering-backend/pom.xml
rsynek: Makes sense, still I would rather use the latest final version, 7.23.0.Final. But it's not a big deal.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/skill/SkillService.java
cuijulian: I agree, we should keep the domain model and architecture consistent with original code. `LocationEntity` is an interesting example; we can consider domain model and persistence layer independence at a later step. I'll keep the `id = null` check for now.

employee-rostering-backend/src/test/java/org/optaweb/employeerostering/controller/SkillControllerTest.java
cuijulian: Correct, this is just the integration test. I will create a SkillService unit test file that will check for individual CRUD functionality.

employee-rostering-backend/pom.xml
```diff
@@ -85,7 +85,7 @@
     <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-configuration-processor</artifactId>
-      <version>2.1.5.RELEASE</version>
+      <version>2.1.6.RELEASE</version>
```
yurloc: Is there a reason to use different version of `spring-boot-configuration-processor` than the rest of Spring Boot artifacts? Doesn't look like a good idea. All the artifact versions (from one group) should be aligned by using a dependency management from imported BOM or from parent and not specifying the version explicitly.

employee-rostering-backend/pom.xml
```diff
@@ -74,71 +84,31 @@
...
-      <optional>true</optional>
-    </dependency>
-    <!-- Include annotation processor that generates metadata for @ConfigurationProperties -->
-    <dependency>
-      <groupId>org.springframework.boot</groupId>
-      <artifactId>spring-boot-configuration-processor</artifactId>
-      <optional>true</optional>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-core</artifactId>
+      <version>1.2.3</version>
```
yurloc: This seems really odd. Why would you need `logback-core` as a compile scoped dependency?

employee-rostering-backend/pom.xml
```diff
@@ -74,71 +84,31 @@
...
-          <artifactId>junit</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
     <dependency>
-      <groupId>org.junit.jupiter</groupId>
-      <artifactId>junit-jupiter-api</artifactId>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
+      <version>1.2.3</version>
```
yurloc: This is unnecessary too. Logback is provided by Spring Boot Starters. See [Logging](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-logging.html). You can verify it by running `mvn dependency:tree`, you'll see the artifact in the dependency tree.

Although it *is* your runtime dependency and it is correct to declare it under normal condition (we ban transitive dependencies in every other kiegroup project), it's against the "philosophy" of Spring Boot [Starters](https://docs.spring.io/spring-boot/docs/current/reference/html/using-boot-build-systems.html#using-boot-starter) which are purposefully used to drag in a bunch of dependencies in a consistent ways so that you don't have to declare them individually.

As an example Spring Boot app, I think we should follow this convention and not override the dependencies provided by Starters unless necessary.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/SkillService.java
```diff
@@ -0,0 +1,79 @@
...
+import org.springframework.util.Assert;
+
+@Service
+public class SkillService {
+
+    private final SkillRepository skillRepository;
+
+    public SkillService(SkillRepository skillRepository) {
+        this.skillRepository = skillRepository;
+        Assert.notNull(skillRepository, "skillRepository must not be null.");
```
yurloc: This null check is kind of redundant. If the Spring IoC cannot find a bean satisfying the `skillRepository` dependency it will [fail](https://www.baeldung.com/spring-nosuchbeandefinitionexception) earlier, so this is an unreachable statement.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/SkillService.java
```diff
@@ -0,0 +1,79 @@
...
+            skillRepository.deleteById(id);
+        }
+        else {
+            throw new EntityNotFoundException("No Skill entity found with ID (" + id + ").");
+        }
+    }
+
+    public Skill createSkill(Integer tenantId, Skill skill) {
+        skillRepository.save(skill);
+        return skill;
```
yurloc: This is a potential source of bugs because `save()` returns a new Skill that may be different from the skill you provided (think of a generated unique ID). You should use the new, returned instance and discard the old one.
> Saves a given entity. Use the returned instance for further operations as the save operation might have changed the entity instance completely.

https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/repository/CrudRepository.html#save-S-

employee-rostering-backend/pom.xml
cuijulian: My mistake, neither of the `logback` dependencies are needed. Spring boot starters dragged in more than I expected. Will keep `mvn dependency:tree` in mind for the future.

yurloc: Perfect! Thanks for confirmation. You don't need to use `mvn dependency:tree` often but when you wonder what's exactly on your classpath it's very handy. I just want to make sure you know it exists.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/SkillService.java
cuijulian: Good catch, just noticed this yesterday. Haven't gotten around to push the changes yet.

PR #251, opened 2019-06-18
Configure ESLint with Airbnb style

yurloc:
Followed installation instructions in https://www.npmjs.com/package/@typescript-eslint/eslint-plugin#usage-with-airbnb.

Airbnb JavaScript Style Guide: https://github.com/airbnb/javascript. Contains explanation and examples for individual rules. Contains useful links.

We're not following it completely because:
- We turn off some rules because they don't work with TypeScript and give false positives.
- We turn off some rules that are too strict in the beginning. We might decide to adopt them later.
- `airbnb` config is extended first. Other configs that are extended after `airbnb` override or turn off some of its rules. This is intentional. We don't want to introduce to many strict rules in the beginning.

https://github.com/airbnb/javascript/tree/master/react contains React and JSX related rules explanation and examples.

More React rules: https://github.com/yannickcr/eslint-plugin-react.

@Christopher-Chianelli I recommend to merge early even though there's other work in progress. Conflict resolution will be easy. You can drop my changes completely because they don't introduce anything original. They're only here to make the build pass. If you drop them when resolving conflicts it's enough to just make your code pass linting.

I highly recommend to keep the code formatted properly at all times by hitting auto-formatting frequently. ESLint can be integrated into many IDEs including [Eclipse](https://github.com/angelozerr/tern.java/wiki/Tern-Linter-ESLint). This will remind you in real-time in case the code is not formatted properly.

The Airbnb style rules really make the code easier to read and understand. This is important for code reviews and future maintenance and expansion.


Christopher-Chianelli:
I agree (after all, we have an issue for adding a formatter: https://issues.jboss.org/browse/PLANNER-936). Will merge after I verify the build.

PR #253, opened 2019-06-21
PLANNER-1539: Normalized the Data Store

Christopher-Chianelli:
Depends on #248


Christopher-Chianelli:
Note: there might be a bug in either the TypeScript compiler or Redux: replacing `Boolean` with `boolean` cause `type` to not be recognized in `Set...ListLoading` and calls to the actions have the error `true` cannot be cast to `false`.

PR #255, opened 2019-06-28
PLANNER-1539: Normalized the Data Store

Christopher-Chianelli:


employee-rostering-frontend/src/domain/DomainObjectView.ts
```diff
@@ -0,0 +1,29 @@
...
+   [K in keyof T]: 
+   T[K] extends DomainObject[]? number[] :
+     T[K] extends DomainObject? number :
+       T[K] extends (DomainObject | null)? number | null :
+         T[K];
+ }
+ 
+/* eslint-disable no-undef */
+export default DomainObjectView;
+/* eslint-enable no-undef */
```
MusaTalluzi: If you want to disable a rule on only one line you can use `// eslint-disable-next-line no-undef` before that line or `// eslint-disable-line no-undef` on the same line.

employee-rostering-frontend/src/store/employee/selectors.ts
```diff
@@ -0,0 +1,41 @@
...
+import { contractSelectors } from 'store/contract';
+import { skillSelectors } from 'store/skill';
+import Employee from 'domain/Employee';
+import DomainObjectView from 'domain/DomainObjectView';
+
+export const getEmployeeById = (state: AppState, id: number): Employee => {
+  if (state.employeeList.isLoading || state.skillList.isLoading || state.contractList.isLoading) {
+    throw Error("Employee list is loading");
+  }
+  const employeeView = state.employeeList.employeeMapById.get(id) as DomainObjectView<Employee>;
```
yurloc: Are you sure that `Map.get()` here never returns `undefined`? Is it guaranteed that an employee with the given `id` is always there in the map? Maybe it is guaranteed but it's not obvious from here and the Map API doesn't guarantee it so I'd probably rather see the `undefined`-check here instead of the cast. Wdyt?

employee-rostering-frontend/src/store/contract/operations.ts
```diff
@@ -45,14 +44,15 @@ export const updateContract: ThunkCommandFactory<Contract, UpdateContractAction>
     const tenantId = contract.tenantId;
     return client.post<Contract>(`/tenant/${tenantId}/contract/update`, contract).then(updatedContract => {
       dispatch(actions.updateContract(updatedContract));
-      dispatch(employeeOperations.refreshEmployeeList());
```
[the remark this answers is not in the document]
Christopher-Chianelli: It is redundant now; since `EmployeeList` is now a `id -> DomainObjectView<Employee>` map, an `Employee` in the map no longer have the `contract` hard coded into the object. Rather, they hold an `id`, and you need to use an `employee` selector to get the `Employee` list with the Contract.

employee-rostering-frontend/src/store/employee/selectors.ts
Christopher-Chianelli: Since the `employeeList`, `skillList`, and `contractList` are finished loading (see the if statement above), all id's should be in the map. Moreover, since the id's come from other `DomainObjects` (with the same `tenantId`), the method would not be called with a non-existent employee id (unless someone modify the JavaScript variables in the console). Then we have the question of what to do with an id that not in the map; I would throw an Error (which is also the implicit current behaviour).

yurloc: Sounds good.

employee-rostering-frontend/src/store/contract/operations.ts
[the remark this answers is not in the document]
yurloc: Yes, of course! I forgot this is one of the reasons we're normalizing the state :slightly_smiling_face:.

PR #257, opened 2019-06-28
[PLANNER-1499] Move Skill domain object to SpringBoot

cuijulian:
Continues from #249, which was closed because of deleted branch from rebasing onto master.


employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/SkillService.java
```diff
@@ -0,0 +1,116 @@
...
+        if (name == null) {
+            throw new IllegalStateException("Skill entity name with tenantId (" + tenantId + ") cannot be null.");
+        }
+
+        List<Skill> skillList = skillRepository.findAllByTenantId(tenantId);
+
+        for (Skill s : skillList) {
+            //If skill already exists
+            if (s.getName().equals(name)) {
+                throw new EntityExistsException("Skill entity with tenantId (" + tenantId + ") and name (" + name
```
Christopher-Chianelli: Is there no way to set constraints directly on entities? In the old backend, we had
```
@Table(uniqueConstraints = @UniqueConstraint(columnNames = {"tenantId", "name"}))
public class Skill extends AbstractPersistable {
// ...
```
which creates a unique constraint for the pair `(tenantId, name)`, and will throw an exception automatically if you try to persist/update name to an existing one for the tenant.

We also have an `@AssertTrue` for different kind of constraints (ex: start time before end time in `EmployeeAvaliability`)

MusaTalluzi: @Christopher-Chianelli +1. This applies for all kinds of validation (nonNull ...). Check https://www.baeldung.com/javax-validation.

cuijulian: You're right, @UniqueConstraint is a better way to handle duplicates.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/SkillService.java
```diff
@@ -0,0 +1,116 @@
...
+            throw new EntityNotFoundException("Skill entity with ID (" + skill.getId() + ") not found.");
+        }
+        //Skill changes tenantId
+        else if (!skillOptional.get().getTenantId().equals(skill.getTenantId())) {
+            throw new IllegalStateException("Skill entity with tenantId (" + skillOptional.get().getTenantId()
+                    + ") cannot change tenants.");
+        }
+
+        skillRepository.deleteById(skill.getId());
+        return skillRepository.save(skill);
```
MusaTalluzi: Don't delete the existing row when you want to update it. `reposiotry.save()` creates a new row for non-existing entity (doesn't have an id) and updates the row if the entity has an id (i.e. exists).

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/controller/SkillController.java
```diff
@@ -0,0 +1,68 @@
...
+public class SkillController {
+
+    private final SkillService skillService;
+
+    public SkillController(SkillService skillService) {
+        this.skillService = skillService;
+        Assert.notNull(skillService, "skillService must not be null.");
+    }
+
+    @GetMapping("/tenant/{tenantId}/skill")
```
MusaTalluzi: Since all endpoints share a common path `/tenant/{tenantId}/skill` add it once for the class by adding `@RequestMapping("/tenant/{tenantId}/skill")`. See https://github.com/MusaTalluzi/spring-boot-task-assigning/blob/master/src/main/java/org/optaplanner/springboottaskassigning/TaskAssigningSolverManagerController.java#L33

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/SkillService.java
```diff
@@ -0,0 +1,116 @@
...
+        validateTenantIdParameter(tenantId, skill);
+
+        Optional<Skill> skillOptional = skillRepository.findById(skill.getId());
+
+        //If name is null
+        if (skill.getName() == null) {
+            throw new IllegalStateException("Skill entity name with tenantId (" + tenantId + ") cannot be null.");
+        }
+        //Skill not found
+        else if (!skillOptional.isPresent()) {
```
MusaTalluzi: Follow the `KIE Java Convention` code style, `else if` should be on the same line as the closing curly brace for preceding if block. (You can use the shortcut `ctl + alt + L` to auto format your file`.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/SkillService.java
cuijulian: Ah, so if `skill` has an id (ie. has been persisted) then `repository.save()` won't create a new row and will simply update? Good to know.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/SkillService.java
```diff
@@ -85,21 +85,15 @@ public Skill updateSkill(Integer tenantId, Skill skill) {
...
-        }
-        //Skill changes tenantId
-        else if (!skillOptional.get().getTenantId().equals(skill.getTenantId())) {
+        } else if (!skillOptional.get().getTenantId().equals(skill.getTenantId())) {
             throw new IllegalStateException("Skill entity with tenantId (" + skillOptional.get().getTenantId()
                     + ") cannot change tenants.");
         }
 
-        skillRepository.deleteById(skill.getId());
         return skillRepository.save(skill);
```
MusaTalluzi: `skill` instance here is the one coming from the request, it might have an old version or lacking an id. A better approach would be:
```java
skillEntity = reposiotry.findById(id); // or skillOptional.get() after ensuring it exists
skillEntity.setName(skill.getName());
... modify other fields
```
Pay closer attention to the memory model and lifecycle of jpa entities (attached/detached ...). It'll be more important when you deal with more complex classes, check how this is handled in [shift roster](https://github.com/kiegroup/optaweb-employee-rostering/blob/3df530af050b235f7ed8614e44db833c788b26b8/employee-rostering-server/src/main/java/org/optaweb/employeerostering/server/roster/RosterRestServiceImpl.java#L334).

cuijulian: Yes, using `Skill` instance from request parameter will lead to optimistic locking exceptions when version is outdated. Is it necessary to save the repository `skillEntity` after modifying its fields? Point 5 in this tutorial goes over updating entities, and it uses `repository.save()` after modifying fields: https://www.baeldung.com/spring-data-crud-repository-save. Otherwise it's consistent with your approach.

I'll do more research into the JPA entity lifecycle and their states. Good catch!

employee-rostering-backend/src/test/java/org/optaweb/employeerostering/controller/SkillControllerTest.java
```diff
@@ -0,0 +1,262 @@
...
+        String name = "name";
+
+        Skill skill = new Skill(tenantId, name);
+
+        skillService.createSkill(tenantId, skill);
+
+        try {
+            skillService.getSkill(1, -1L);
+        }
+        catch (EntityNotFoundException e) {
```
yurloc: `catch` should be on the same line as the closing bracket. Same as if/else:
```java
if (expr) {
    // something1
} else {
    // something2
}
```
```java
try {
    // unsafe stuff
} catch (MyException e) {
    // cleanup
}
```

employee-rostering-backend/src/test/java/org/optaweb/employeerostering/controller/SkillControllerTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+        Skill skill = new Skill(tenantId, name);
+
+        skillService.createSkill(tenantId, skill);
+
+        try {
+            skillService.getSkill(1, -1L);
+        } catch (EntityNotFoundException e) {
+            String expectedMessage = "No Skill entity found with ID (-1).";
+            assertEquals(expectedMessage, e.getMessage());
```
Christopher-Chianelli: Non Blocker; you can use AssertJ for testing exceptions (see https://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#exception-assertion). In this case, it be

```
assertThatExceptionOfType(EntityNotFoundException.class).isThrownBy(() => skillService.getSkill(1, -1L))
    .withMessage("No Skill entity found with ID (-1).");
```

PR #260, opened 2019-07-11
 Enable SonarCloud on Employee Rostering Backend 

Christopher-Chianelli:


employee-rostering-frontend/pom.xml
```diff
@@ -79,5 +75,16 @@
       </plugin>
     </plugins>
   </build>
+  
+  <profiles>
+    <profile>
+      <id>sonar</id>
+      <properties>
+        <sonar.sources>src</sonar.sources>
+        <sonar.test.inclusions>optaweb-employee-rostering-frontend/**/*test.ts,optaweb-employee-rostering-frontend/**/*test.tsx</sonar.test.inclusions>
```
yurloc: @Christopher-Chianelli Simply use `**/*test.ts,**/*test.tsx`. Sonar will complain that patterns should be relative to module root but it's currently the only way to get correct coverage report.

PR #261, opened 2019-07-11
[PLANNER-1533] Move Spot domain object to SpringBoot

cuijulian:
- Create Spot domain entity
- Expose Spot CRUD endpoints using service and controller
- Add integration tests for Spot persistence


employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/skill/Skill.java
```diff
@@ -23,14 +23,16 @@
...
+
 @Entity
 @Table(uniqueConstraints = @UniqueConstraint(columnNames = {"tenantId", "name"}))
 public class Skill extends AbstractPersistable {
 
     @NotNull
     @Size(min = 1, max = 120)
     @Pattern(regexp = "^(?!\\s).*(?<!\\s)$", message = "Name should not contain any leading or trailing whitespaces")
-    private String name;
+    protected String name;
```
yurloc: Seems to be unnecessary.

employee-rostering-backend/src/test/java/org/optaweb/employeerostering/controller/SpotControllerTest.java
```diff
@@ -0,0 +1,301 @@
...
+
+        Spot returnSpot = spotService.createSpot(tenantId, spot);
+
+        assertEquals((Integer) 1, returnSpot.getTenantId());
+        assertEquals("name", returnSpot.getName());
+        assertEquals(spot.getRequiredSkillSet(), returnSpot.getRequiredSkillSet());
+    }
+
+    @Test
+    public void createNonMatchingSpotTest(){
```
yurloc: Missing space, should be `...Test() {`.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/persistence/SpotRepository.java
```diff
@@ -0,0 +1,28 @@
...
+package org.optaweb.employeerostering.persistence;
+
+import java.util.List;
+
+import org.optaweb.employeerostering.domain.spot.Spot;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface SpotRepository extends JpaRepository <Spot, Long> {
```
yurloc: Extra space, should be `JpaRepository<Spot, Long>`.

employee-rostering-backend/src/test/java/org/optaweb/employeerostering/controller/SpotControllerTest.java
```diff
@@ -0,0 +1,301 @@
...
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.junit4.SpringRunner;
+
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.junit.Assert.assertEquals;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@Transactional
+public class SpotControllerTest {
```
yurloc: I think this is actually a `SpotServiceTest` because it tests the service and doesn't even touch the `SpotController`.

yurloc:
Wrt. controller coverage, unless we decide to ignore it, I can see 2 options:
1. Unit test with mocked out dependencies (services that the controller invokes).
2. Integration test (`@RunWith(SpringBootTest.class`), using something like https://www.baeldung.com/spring-boot-testresttemplate or https://www.baeldung.com/restclienttest-in-spring-boot.

Option 1 is fast and focused.

Option 2 is end-to-end but slow and we already have an integration test for the service so it would be redundant I guess.

Question: do we prefer that our integration tests start with the service layer and circumvent the controller?

Note: Keep in mind the [test pyramid](https://martinfowler.com/bliki/TestPyramid.html). I personally try to achieve close to 100% coverage with unit tests and then add a few integration tests on top of that.

employee-rostering-backend/src/test/java/org/optaweb/employeerostering/skill/SkillRestControllerIT.java
```diff
@@ -0,0 +1,164 @@
...
+                , tenantId);
+        ResponseEntity<Skill> postResponse3 = skillRestTemplate.postForEntity(skillPathURI + "add", skill3, Skill.class
+                , tenantId2);
+
+        ResponseEntity<List<Skill>> response = skillRestTemplate.exchange(skillPathURI, HttpMethod.GET, null,
+                                                                          new ParameterizedTypeReference
+                                                                                  <List<Skill>>() {}, tenantId);
+        ResponseEntity<List<Skill>> response2 = skillRestTemplate.exchange(skillPathURI, HttpMethod.GET, null,
+                                                                           new ParameterizedTypeReference
+                                                                                   <List<Skill>>() {}, tenantId2);
```
yurloc: Could you please isolate the `skillRestTemplate` calls into reusable method. The tests will become much simpler to understand if we can read something like:
```java
Skill skill = new Skill(tenantId, name);
ResponseEntity<Skill> postResponse = addSkill(skill, tenantId);
ResponseEntity<List<Skill>> response = getSkills(tenantId);
```

employee-rostering-backend/src/test/java/org/optaweb/employeerostering/skill/SkillServiceUnitTest.java
```diff
@@ -0,0 +1,270 @@
...
+import org.springframework.web.util.NestedServletException;
+
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@AutoConfigureMockMvc
+@Transactional
+public class SkillServiceUnitTest {
```
yurloc: The "Unit" in the test name is superfluous. "IT" as a class name suffix is required by Maven Failsafe plugin. It's used to exclude integration test (usually long-running) from the `test` phase and defer them to `integration-test` phase which is executed less often during development.

As I'm writing this, I realize that your `*ControllerIT` tests are actually not executed at all because the project doesn't use Failsafe plugin and Surefire's default include pattern doesn't match ITs. So unless you want to execute them in `integration-test` phase (I don't see a reason for that at the moment), rename them to `*ControllerTest`, please.

And please remove "Unit" from test names as well. There is no such convention yet and for consistency reasons, let's not introduce it. It would be hard to follow such convention consistently. We'd have to think about classification of tests as unit, component, integration, something in between? and reflect it in class names. I don't think we want to do that.

cuijulian: Wow, I didn't know something as subtle as a test class name could affect test phases. Will fix this and test name convention in next push. Thanks Jiri!

employee-rostering-backend/src/test/java/org/optaweb/employeerostering/skill/SkillRestControllerTest.java
```diff
@@ -0,0 +1,168 @@
...
+public class SkillRestControllerTest {
+
+    @Autowired
+    private TestRestTemplate skillRestTemplate;
+
+    private String skillPathURI = "http://localhost:8080/rest/tenant/{tenantId}/skill/";
+
+    private ResponseEntity<List<Skill>> getSkills(Integer tenantId) {
+        return skillRestTemplate.exchange(skillPathURI, HttpMethod.GET, null, new ParameterizedTypeReference
+                <List<Skill>>() {}, tenantId);
```
yurloc: Code style: The point between class name and its type parameters is not the best place for a line break. Here, the whole expression creating an instance of the anonymous class functions as a single argument. Except for the type param, it's pretty atomic so it's a bit confusing that it's interrupted in the middle and continues on the next line.

Space between parameters is a better place for a line break.

Another option is to create the instance and store it in a variable as a separate statement:
```java
ParameterizedTypeReference<List<Spot>> typeReference = new ParameterizedTypeReference<List<Spot>>() {};
return spotRestTemplate.exchange(spotPathURI, HttpMethod.GET, null, typeReference, tenantId);
```

employee-rostering-backend/src/test/java/org/optaweb/employeerostering/spot/SpotServiceTest.java
```diff
@@ -0,0 +1,341 @@
...
+
+    @Autowired
+    private SpotService spotService;
+
+    @Autowired
+    private SkillService skillService;
+
+    private Skill createSkill(Integer tenantId, String name) {
+        Skill skill = new Skill(tenantId, name);
+        Skill out = skillService.createSkill(tenantId, skill);
```
yurloc: Code style: redundant local variable. "Adds nothing to comprehensibility of the method", as IDEA reports :slightly_smiling_face:. You can return the value immediately.

yurloc:
@Christopher-Chianelli please merge when you get to this.

PR #264, opened 2019-07-22
PLANNER-1550: Add mock Tenant entity

cuijulian:
- Purpose: test new React frontend and new SpringBoot backend integration by creating mock Tenant entity to test entity persistence
- Move Tenant domain entity to Spring Boot
- Add TenantMockController to create and return mock Tenant entity to test web application startup

### DIFF
--- a/employee-rostering-backend/src/main/java/org/optaweb/employeerostering/controller/TenantMockController.java
+++ b/employee-rostering-backend/src/main/java/org/optaweb/employeerostering/controller/TenantMockController.java
@@ -34,7 +34,7 @@ public class TenantMockController {
     final Tenant TENANT = createTenant();
 
     @GetMapping
-    public ResponseEntity<List> getTenantList() {
+    public ResponseEntity<List<Tenant>> getTenantList() {
         return new ResponseEntity<>(Arrays.asList(TENANT), HttpStatus.OK);
     }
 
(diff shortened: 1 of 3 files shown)


employee-rostering-backend/src/main/java/org/optaweb/employeerostering/controller/TenantMockController.java
```diff
@@ -0,0 +1,52 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping("/rest/tenant")
+public class TenantMockController {
```
rsynek: Why not using a real tenant controller for integration testing? Is this a temporary solution before the controller is implemented?

Christopher-Chianelli: IIRC, the main reason for this class is so @cuijulian can see how the frontend handles the new backend. It is a temporary solution before the controller is implemented.

cuijulian: Yes, as @Christopher-Chianelli has stated this is a temporary solution so we could test the new frontend with the new backend. The Tenant controller and service (ie. CRUD functionality) will be one of the last entities to be moved to Spring Boot, so this mock controller will be replaced by that once it's implemented.

I will be creating a PR for the entities I have moved to Spring Boot once Christopher merges his current PR. That PR will include the changes made in this one.

PR #265, opened 2019-07-30
PLANNER-1552-ShiftTemplate

cuijulian:
- Create domain entities: Tenant, Contract, Employee, EmployeeAvailability, Shift, RosterState, RosterParametrization, Roster, ShiftTemplate
- Add controller and service CRUD implementation to domain entities
- Add integration and unit tests for persistence


employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/contract/ContractController.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @DeleteMapping("/{id}")
+    public ResponseEntity<Boolean> deleteContract(@PathVariable Integer tenantId, @PathVariable Long id) {
+        return new ResponseEntity<>(contractService.deleteContract(tenantId, id), HttpStatus.OK);
+    }
+
+    @PostMapping("/add")
+    public ResponseEntity<Contract> createContract(@PathVariable Integer tenantId,
+                                                   @RequestBody ContractDTO contractDTO) {
+        Contract contract = new Contract(contractDTO.getTenantId(), contractDTO.getName(),
```
Christopher-Chianelli: It might be useful to create a `Contract` constructor which takes a `ContractDTO` and populate its field from it.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/spot/SpotService.java
```diff
@@ -57,7 +57,7 @@ public Boolean deleteSpot(Integer tenantId, Long id) {
         Optional<Spot> spotOptional = spotRepository.findById(id);
 
         if (!spotOptional.isPresent()) {
-            throw new EntityNotFoundException("No Spot entity found with ID (" + id + ").");
```
Christopher-Chianelli: Why don't you throw an Exception?

cuijulian: I initially thought that `false` should be returned when entity is not deleted (which includes the case when it's not found), but throwing an Exception for that does make more sense in hindsight. Will fix this to keep consistent with old code.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/contract/ContractDTO.java
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaweb.employeerostering.domain.contract;
+
+public class ContractDTO extends Contract {
```
rsynek: This class is supposed to be used for json serialization - it's used in the controller methods. But it extends the `Contract` class, which contains all the fields and methods, thus this DTO class actually adds no value. It's closely coupled with the `Contract` class by extending it.
That's the opposite to a reason for introducing such a DTO class - to loose the coupling between core domain and json serialization.
The `Contract` class also depends on `ContractDTO`, as it accepts it in one of the constructors. Together with the "extends" relationship, this creates a circular dependency.

Suggestion:
- don't extend `Contract` by the DTO class, rather duplicate the fields of the core domain class and create a standalone DTO, which can be safely used for json serialization
- never create a dependency from core domain class to such a DTO class. The core domain classes represent business rules of your application, while the DTO class is just an implementation detail on how to serialize some data to json. Rather create a constructor here, taking the `Contract` class as a parameter.
- the name `ContractDTO` does not tell much. Neither the location (package) of the class. Does it deal with JPA? Json? Xml? Who knows - before opening the file and reading through it. If you e.g. group a "copy" of the domain classes that is supposed to serve for json serialization into a package that clearly reflects it, you will make it obvious.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/employee/EmployeeAvailability.java
```diff
@@ -0,0 +1,127 @@
...
+        this.startDateTime = OffsetDateTime.of(employeeAvailabilityView.getStartDateTime(),
+                                             zoneId.getRules().getOffset(employeeAvailabilityView.getStartDateTime()));
+        this.endDateTime = OffsetDateTime.of(employeeAvailabilityView.getEndDateTime(),
+                                             zoneId.getRules().getOffset(employeeAvailabilityView.getEndDateTime()));
+        this.state = employeeAvailabilityView.getState();
+    }
+
+    @AssertTrue
+    @JsonIgnore
+    public boolean isValid() {
```
rsynek: The same architectural issue as with the `Contract` class - violating the SRP.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/rotation/view/ShiftTemplateView.java
```diff
@@ -0,0 +1,102 @@
...
+package org.optaweb.employeerostering.domain.rotation.view;
+
+import java.time.Duration;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import org.optaweb.employeerostering.domain.common.AbstractPersistable;
+import org.optaweb.employeerostering.domain.common.HasTimeslot;
+import org.optaweb.employeerostering.domain.rotation.ShiftTemplate;
+
+public class ShiftTemplateView extends AbstractPersistable implements HasTimeslot {
```
rsynek: Do I understand it correctly that *View classes are equivalent to the *DTO classes in terms of their purpose? If so, it would be good to unify the naming and place them to a similar package.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/common/DateTimeUtil.java
```diff
@@ -0,0 +1,29 @@
...
+
+package org.optaweb.employeerostering.domain.common;
+
+import java.time.LocalDateTime;
+import java.time.OffsetDateTime;
+import java.time.ZoneId;
+
+public class DateTimeUtil {
+
+    public static LocalDateTime toLocalDateTimeInZone(OffsetDateTime dateTime, ZoneId zoneId) {
```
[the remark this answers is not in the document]
Christopher-Chianelli: This code is from the old `shared` directory, which in turn use a different `java.time` implementation that didn't have all the methods. The purpose of this method is to convert an OffsetDateTime into a LocalDateTime which will be shown to the client.  The client is more interested in the LocalDateTime in the Zone than the actual entry, so this basically computes `dateTime.atZoneSameInstant(zoneId).toLocalDateTime()`. However, the OffsetDateTime is generated from the ZoneId, so `atZoneSameInstant(zoneId)` may not be needed (unless the `ZoneId` has its rules change, which is completely possible).

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/common/HasTimeslot.java
```diff
@@ -0,0 +1,33 @@
...
+
+package org.optaweb.employeerostering.domain.common;
+
+import java.time.Duration;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+
+public interface HasTimeslot {
+
+    public static LocalDateTime EPOCH = LocalDateTime.ofEpochSecond(0, 0, ZoneOffset.UTC);
```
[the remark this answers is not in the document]
cuijulian: Noted. Since there is only one shared constant, I don't think it makes sense to create a whole new class for it. For good convention, I'll declare the constant in the separate classes. This should also improve readability for the classes that use `EPOCH`.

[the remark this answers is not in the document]
rsynek: Declaring the constant in each class that use it leads to code duplication. 

What is the constant used for? Most likely for a comparison or calculation of a duration since the epoch, right? How about encapsulating the constant into a class that has methods providing exactly the described capabilities? The code that now directly use the method can call the methods instead.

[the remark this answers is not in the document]
Christopher-Chianelli: It is a hack to allow Shift Template to use the same viewer as Shifts and Employee Availabilities. Rotations don't really have/use "Dates", but in order to use the old viewers, a reference to a date was needed to place entities onto the viewer. The reference is the Epoch. Since the frontend and backend code are decoupled now, I think the constant and the `HasTimeslot` interface can be safely removed.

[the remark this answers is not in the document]
cuijulian: Looking through the old code, it seems that with the new frontend and backend, this hack is not needed. I agree with removing it.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/contract/Contract.java
```diff
@@ -0,0 +1,157 @@
...
+import javax.validation.constraints.Pattern;
+import javax.validation.constraints.Size;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import org.optaweb.employeerostering.domain.common.AbstractPersistable;
+
+@Entity
+@Table(uniqueConstraints = {@UniqueConstraint(columnNames = {"tenantId", "name"}),
+        @UniqueConstraint(columnNames = {"id"})})
+public class Contract extends AbstractPersistable {
```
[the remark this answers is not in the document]
cuijulian: This implementation is found in the old code, and should be addressed in a future PR.  I can add a TODO for the SRP issue.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/contract/ContractDTO.java
cuijulian: Very good points, especially on the circular dependency. I plan to keep the DTO objects for reasons I explain below. To fix the circular dependency, I will duplicate the fields from the core domain class and remove the dependent constructor in the `Contract` class. For clarity, I plan to move all the DTO classes to a package called `jsonserialization` so their purpose is clear. Do you agree with this approach?

The reason DTOs were introduced is due to the Spring framework convention. Because Spring MVC automatically binds request parameters to beans declared as arguments of methods annotated with any form of `@RequestMapping` (which is indeed the case for `updatecontract()` and `createContract()`). Passing objects annotated with `@Entity` (`Contract` is an example) directly into controller methods introduces a critical security vulnerability, as users can pass malicious data into the method and directly change unexpected fields in the database. I believe having a DTO class for serialization is necessary due to this reason. It is true that the DTO object adds no business logic, it is purely to mitigate the security flaw, but logic may be added in the future (eg. checking user input).

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/employee/EmployeeAvailability.java
cuijulian: Similar to `Contract`, will add TODO.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/rotation/view/ShiftTemplateView.java
cuijulian: The *View classes and *DTO classes don't have the same purpose. The DTO classes were newly added to avoid the security flaw mentioned above and conform with Spring convention. From what I understand, the View objects were used to store data useful for the _old_ frontend to display the information.

@Christopher-Chianelli feel free to correct me if my comments above are wrong. Does the new frontend require the *View objects?

Christopher-Chianelli: The `*DTO` and `*View` classes share the same purpose (which is to remove backend specific information such as time zone that is not needed in the frontend). The new frontend require the *View objects. (Although the `*DTO` were introduced to circumvent a security flaw, that security flaw is related to the purpose of the `*View`)

cuijulian: I see. Would it make sense to rename the `*DTO` classes to `*View` for consistency?

Christopher-Chianelli: +1 to renaming `*DTO` classes to `*View`

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/contract/ContractDTO.java
rsynek: The usage of DTOs is completely fine - not only because of some security flaws, but due to the separation of concerns as I tried to explain => good architecture can actually reduce the risk of security vulnerabilities.
The issue I had with DTOs is that the name does not tell you anything about the purpose - it's very general. Placing them into a package saying the purpose helps. Imagine now, that you have a DTO for service->REST and DTO for database->service. If both classes are named *DTO, it will be hard to distinguish them just by the package. Thus I would actually move them to the right package and also rename them to better reflect their purpose.

rsynek: For inspiration: http://coders-kitchen.com/2013/07/22/serie-choosing-names-1-part-dtos/

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/rotation/view/ShiftTemplateView.java
rsynek: If the purpose is to get the data to frontend, `*View` sounds good.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/common/DateTimeUtil.java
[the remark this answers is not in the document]
rsynek: Thanks

PR #267, opened 2019-08-07
PLANNER-1514: Add Availability Page

Christopher-Chianelli:


Christopher-Chianelli:
FYI, the two bugs in SonarCloud are false positives; the font-family `pficon` should be provided by PatternFly (`pficon` is the font-family for PatternFly icons).

employee-rostering-frontend/src/store/roster/operations.ts
```diff
@@ -168,4 +206,50 @@ export const getShiftRosterFor: ThunkCommandFactory<RosterSliceInfo & { spotList
...
+    });
+  };
+
+export const getAvailabilityRosterFor: ThunkCommandFactory<RosterSliceInfo & { employeeList: Employee[] }, SetAvailabilityRosterIsLoadingAction | SetAvailabilityRosterViewAction> = params =>
+  (dispatch, state, client) => {
+    const tenantId = state().tenantData.currentTenantId;
+    const fromDateAsString = moment(params.fromDate).format("YYYY-MM-DD");
+    const toDateAsString = moment(params.toDate).add(1, "day").format("YYYY-MM-DD");
+    dispatch(actions.setAvailabilityRosterIsLoading(true));
+    return client.post<KindaAvailabilityRosterView>(`/tenant/${tenantId}/roster/availabilityRosterView/for?` +
```
yurloc: Mixing different levels of abstraction. HTTP level (method, URL, encoding parameters) and application logic (I can't easily understand what's actually posted here and how it's related to getting an employee's availability).

Could you create a wrapper around the rest client that would have methods in domain language? For example:
```ts
client.get<RosterState>(`/tenant/${tenantId}/roster/state`);
```
change to
```ts
client.getRosterState(tenantId);
```
-> less verbose, less distraction with lower level stuff.

employee-rostering-frontend/src/store/roster/operations.ts
```diff
@@ -39,9 +42,18 @@ interface KindaShiftRosterView extends Omit<ShiftRosterView, "spotIdToShiftViewL
...
+  employeeIdToShiftViewListMap: ObjectNumberMap<KindaShiftView[]>;
+  employeeIdToAvailabilityViewListMap: ObjectNumberMap<KindaEmployeeAvailabilityView[]>;
+  unassignedShiftViewList: KindaShiftView[];
+}
+
 let lastCalledShiftRosterArgs: any | null;
 let lastCalledShiftRoster: ThunkCommandFactory<any, SetShiftRosterIsLoadingAction | SetShiftRosterViewAction> | null = null;
 
+let lastCalledAvailabilityRosterArgs: any | null;
+let lastCalledAvailabilityRoster: ThunkCommandFactory<any, SetAvailabilityRosterIsLoadingAction | SetAvailabilityRosterViewAction> | null = null;
```
yurloc: Is this the best way to achieve auto-refresh? Maybe if the TSX component set up the timeout it could use the same arguments and you wouldn't have to remember it here? Or you could keep it in the AppState. Any example code, blog etc. out there that uses similar approach?

employee-rostering-frontend/src/store/roster/operations.ts
Christopher-Chianelli: The REST client would get extremely large if I were to wrap every single REST method in it. As I see it, the "operations" class serves as the wrapper (with the exception it writes to AppState instead of returning a value).

employee-rostering-frontend/src/store/roster/operations.ts
Christopher-Chianelli: The primary issue of putting auto-refresh directly into a TSX component is that `Shift` and `EmployeeAvailability` operations modify both the Shift Roster and the Availability roster in non-trivial ways (i.e. can generate new indictments), meaning the Shift Roster and Availability roster need to be refreshed. There are quite a few ways to call `Shift` and `EmployeeAvailability` operations in the UI (using the `Modal` or click and drag a new Shift), so it cannot be tied to a single UI component.

Christopher-Chianelli:
@yurloc The "flickering" also occurs on the Shift Roster page, but only on larger datasets. The `Shift Roster` and `Availability Roster` set an `isLoading` flag when they are being loaded. This is needed since if you were to change `Tenants`, the `Employees` and the like referenced in the old `Shift Roster` are no longer valid, and cause a crash. When the `isLoading` flag is set, the page render a blank page (hence cause the flickering). I did a trick in Redux to keep the old page when `isLoading` is set, but when solving, props change before `isLoading` is set causing the blank screen. I'll look for a work around.

PR #268, opened 2019-08-12
PLANNER-1514: Add Availability Roster in React

Christopher-Chianelli:
Continues #267 (which was closed due to rebasing `react-spring-boot-refactor` onto `master`).

### DIFF
--- a/employee-rostering-frontend/src/store/availability/operations.ts
+++ b/employee-rostering-frontend/src/store/availability/operations.ts
@@ -35,7 +35,8 @@ export function availabilityAdapter(employeeAvailability: EmployeeAvailability):
   };
 }
 
-export function kindaAvailabilityViewAdapter(kindaAvailabilityView: KindaEmployeeAvailabilityView): EmployeeAvailabilityView {
+export function kindaAvailabilityViewAdapter(kindaAvailabilityView: KindaEmployeeAvailabilityView):
+EmployeeAvailabilityView {
   return {
     ...kindaAvailabilityView,
     startDateTime: moment(kindaAvailabilityView.startDateTime).toDate(),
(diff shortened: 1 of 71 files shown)


Christopher-Chianelli:
@yurloc The Screen Flickering problem is fixed. As for your other issues:

- Is there a rule in `npm lint` to break lines after `120` characters?
- I can move about `200` lines out of `AvailabilityRosterPage`.
- `Dropdown Bug 1` probably need to be fixed upstream at `patternfly` (I don't do anything with auto-complete; I needed to do some CSS hacks just to allow users to input text).
- I will see if there anything I can do on `Dropdown Bug 2`; it depends on how `patternfly` implemented the dropdown.
- I disagree with creating wrappers for every single method in the `RestServiceClient` class, since that will make it massive and I believe the `operations.ts` file serve as the wrappers (`operation.ts` contain very little application logic, they literally just past the result down into `AppState` (with the exception of `solveRoster`, which control intervals)).
- The primary issue of putting auto-refresh directly into a TSX component is that Shift and EmployeeAvailability operations modify both the Shift Roster and the Availability roster in non-trivial ways (i.e. can generate new indictments), meaning the Shift Roster and Availability roster need to be refreshed. There are quite a few ways to call Shift and EmployeeAvailability operations in the UI (using the Modal or click and drag a new Shift), so it cannot be tied to a single UI component.

Christopher-Chianelli:
Reduced size of `AvailabilityRosterPage` by about `100` lines + enforced 120 char limit per line in the latest commit.

Christopher-Chianelli:
Cannot fix `Dropdown Bug 2` as the Dropdown is a child of `Select` which is a child of the modal; cannot be fixed without changing the upstream `Select` component (https://stackoverflow.com/questions/16057361/how-to-make-child-element-higher-z-index-than-parent) (none of the CSS tricks in there worked either)

employee-rostering-frontend/src/ui/pages/availability/AvailabilityEvent.tsx
```diff
@@ -0,0 +1,96 @@
...
+            onClick={() => props.removeEmployeeAvailability(props.availability)}
+            variant="danger"
+          >
+            <TrashIcon />
+          </Button>
+        </SplitItem>
+      </Split>
+      <Level gutter="sm">
+        <LevelItem>
+          <Button
```
MusaTalluzi: Adding attribute `title="Desired"` is better to users know what's the button for by hovering over it. Similarly for the other two buttons. It's not too obvious from the icons and colours, although it makes sense after knowing what they're for.

employee-rostering-frontend/src/ui/pages/availability/AvailabilityEvent.tsx
```diff
@@ -0,0 +1,96 @@
...
+            onClick={() => props.updateEmployeeAvailability({
+              ...props.availability,
+              state: "DESIRED"
+            })}
+            style={{
```
MusaTalluzi: Buttons are too big and in most displays the unavailable one don't appear. Maybe them 10px and the margin 1px?
In addition, for smaller unavailable slots (e.g. 1-hour) the buttons don't appear at all which makes it impossible for users to modify/delete this unavailability.
Add an [onSelectEvent](http://intljusticemission.github.io/react-big-calendar/examples/index.html#prop-onSelectEvent) that shows details of the event and these buttons or the EditAvailabilityModal (see how it works on google calendar).

employee-rostering-frontend/src/store/availability/operations.ts
```diff
@@ -0,0 +1,96 @@
...
+    const shiftId = employeeAvailability.id;
+    return client.delete<boolean>(`/tenant/${tenantId}/employee/availability/${shiftId}`).then(isSuccess => {
+      if (isSuccess) {
+        dispatch(alert.showSuccessMessage("removeAvailability", { 
+          employeeName: employeeAvailability.employee.name,
+          startDateTime: moment(employeeAvailability.startDateTime).format("LLL"),
+          endDateTime: moment(employeeAvailability.endDateTime).format("LLL") 
+        }));
+        dispatch(refreshShiftRoster());
+        dispatch(refreshAvailabilityRoster());
```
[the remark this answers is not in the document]
Christopher-Chianelli: Not possible, since we include indictments inside the shift and availability objects. If you add an availability, it can change the indictments of shifts (likewise for adding/updating a shift). If there were no indictments, then it would be possible.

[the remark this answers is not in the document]
Christopher-Chianelli: (note this is only for when a client changes/update a shift/availability)

Christopher-Chianelli:
For your first issue, I don't think it is the client fault, but probably something in the back end (I don't think the planning facts/objects are updated when solving). Second issue, probably cannot do (related to react-big-calendar). I would need to hide either the time or the employee for it to fully render (it occurs as there is too little height for both). Refresh is due to a missing i18n translation; will add; Will fix typos in Contracts.

employee-rostering-frontend/src/ui/pages/availability/AvailabilityEvent.tsx
Christopher-Chianelli: Cannot shrink button much; already at min-content. Add popup (a bit wonky, but I will fix it in a later PR where I refactor the how the Roster Pages work).

MusaTalluzi: You can add `fontSize: "10px"` for example in the `style` to make the buttons smaller (min size of button depends on the size of the text inside it). See what font size looks ideal and displays all buttons in medium screen sizes.

MusaTalluzi: And yeah the pop-up only shows for larger events because it's a nested component inside the event so it's after the text, when event is small you can't even click on it (same problem). But it's fine to be fixed in a later PR.

MusaTalluzi:
LGTM

employee-rostering-frontend/src/ui/pages/shift/ReactBigCalendarOverrides.css
```diff
@@ -32,22 +35,152 @@
...
+    position: absolute !important;
+    top: 0 !important;
+    left: 0 !important;
+    width: 100% !important;
+    height: 100% !important;
+    opacity: 0.75;
+}
+
+.availability-event {
+    display: "flex";
```
yurloc: @Christopher-Chianelli This is not a valid value for the `display` property. If you want to use `flex` use `display: flex` (https://developer.mozilla.org/en-US/docs/Web/CSS/flex#Example).

PR #269, opened 2019-08-12
PLANNER-1551: Add RosterGenerator and BenchmarkApp

cuijulian:
- Add StringDataGenerator and RosterGenerator to generate sample data on startup with ApplicationRunner and integration tests
- Add SystemPropertiesRetriever
- Add solver config file and business constraints in Drools format 
- Add BenchmarkApp


employee-rostering-backend/src/main/resources/application.properties
```diff
@@ -1,2 +1,8 @@
+# Setting datasource
+spring.datasource.driverClassName=org.h2.Driver
+spring.datasource.url=jdbc:h2:mem:testdb
```
Christopher-Chianelli: Why do we need this? Is it for the Benchmark>

employee-rostering-backend/src/test/java/org/optaweb/employeerostering/contract/ContractServiceTest.java
```diff
@@ -103,23 +106,23 @@ public void getNonExistentContractTest() {
 
     @Test
     public void getNonMatchingContractTest() {
-        Integer tenantId = 1;
+        Integer tenantId = 2;
```
Christopher-Chianelli: Why did you need to change the tenantId?

employee-rostering-backend/src/main/resources/application.properties
cuijulian: This is not needed, Spring Boot autoconfigures the DataSource. I added it to make sure h2 is used for testing purposes, but I can remove it.

employee-rostering-backend/src/test/java/org/optaweb/employeerostering/contract/ContractServiceTest.java
cuijulian: This is a temporary hack to get around the fact that `RosterGenerator` generates entities whose `id`'s interfere with the entities' `id`'s in the test classes (one of the Tenants created by `RosterGenerator` has `id = 1`, which is the same `id` used in this test). Once I implement Tenant CRUD I will add `createTestTenant()` methods to each test class so that each test class has its own persisted `Tenant` entity that will not interfere with `RosterGenerator`. This . Your `BaseTest` solution also gets around this problem, but with this alternative we don't have to "turn off" `RosterGenerator` for certain test classes. I'm open to discussing this, love to know what you think.

PR #275, opened 2019-09-03
Update Backend README doc

cuijulian:


employee-rostering-backend/README.adoc
```diff
@@ -1 +1,57 @@
...
+"Security Rating", link="https://sonarcloud.io/dashboard?id=optaweb-employee-rostering"]
+image:https://sonarcloud.io/api/project_badges/measure?project=optaweb-employee-rostering&metric=sqale_rating[
+"Maintainability Rating", link="https://sonarcloud.io/dashboard?id=optaweb-employee-rostering"]
+image:https://sonarcloud.io/api/project_badges/measure?project=optaweb-employee-rostering&metric=ncloc[
+"Lines of Code", link="https://sonarcloud.io/dashboard?id=optaweb-employee-rostering"]
+image:https://sonarcloud.io/api/project_badges/measure?project=optaweb-employee-rostering&metric=coverage[
+"Coverage", link="https://sonarcloud.io/dashboard?id=optaweb-employee-rostering"]
+
+image:https://david-dm.org/kiegroup/optaweb-employee-rostering/status.svg?path=optaweb-employee-rostering[
+"dependencies Status", link="https://david-dm.org/kiegroup/optaweb-employee-rostering?path=optaweb-employee-rostering"]
```
ge0ffrey: What is david-dm.org? Is it a reliable service (= do we expect it to keep running in 3 years from now - 90% confidence or more) @rsynek Do we use david-dm.org elsewhere?

employee-rostering-backend/README.adoc
```diff
@@ -1 +1,57 @@
...
+In the project backend directory, run the following command to build from source:
+
+----
+mvn clean install
+----
+
+To ignore tests during the build, run:
+
+----
+mvn clean install -DskipTests
```
ge0ffrey: I'd immediately give them the -DskipTests command.
People know what it does, and it's a good thing for people who try it out quickly to ignore the tests so they get it to run faster. Only a small portion of them become contributors and those will know to run the tests before submitting a PR (and if they don't, jenkins does it for them anyway).

employee-rostering-backend/README.adoc
```diff
@@ -1 +1,57 @@
...
+----
+
+To quickly compile and run the backend application with the Spring Boot Maven plugin, use the following command:
```
ge0ffrey: "To quickly run" is passive voice.
I'd also leave out the "quickly" as you're not mention to quicker than what. 5 seconds is not always quick (think delay in buying stocks or playing shooters).

=> "Compile and run the backend:"

employee-rostering-backend/README.adoc
```diff
@@ -1 +1,57 @@
...
+
+----
+mvn spring-boot:run
+----
+
+Once the application starts, test the REST API at http://localhost:8080/swagger-ui.html. Select a specific HTTP method and click on the `Try it out` button.
+
+To test the database, open http://localhost:8080/h2-console to access the H2 in-memory database console.
```
ge0ffrey: use active voice
=> "Query the database".
"test" is used a lot (think unit testing, manual testing, etc). "Query" seems more appropriate.

employee-rostering-backend/README.adoc
```diff
@@ -1 +1,57 @@
...
+In the `JDBC URL` field, enter `jdbc:h2:mem:testdb` and keep the other default values.
+Once connected, the console will display the persisted data generated on startup, and a textbox for running SQL queries on the database.
```
ge0ffrey: "will display" -> displays

employee-rostering-backend/README.adoc
```diff
@@ -1 +1,57 @@
...
+
+Note that the in-memory database is volatile and data will be lost when the application is restarted.
```
ge0ffrey: I'd drop "is volatile", concise it to "The in-memory database loses all data when the application is restarted."

employee-rostering-backend/README.adoc
```diff
@@ -1 +1,57 @@
...
+
+== Resources
+
+Learn more about https://optaplanner.org/[Optaplanner].
+
+Learn more about https://docs.spring.io/spring-boot/docs/current/reference/html/[Spring Boot].
+
+Learn more about https://swagger.io/[Swagger].
+
+Learn more about https://www.baeldung.com/spring-boot-h2-database[H2] with Spring Boot.
```
ge0ffrey: These links can get stale pretty fast. Especially that last one.
And a "resources" section is an anti-pattern: it's better to link directly in text, so the motivation for *why* that link is worth visiting is there.

I'd remove the entire resources section:
- swagger is an out-of-the-blue mention, and it's linked in the generated swagger html anyway in the app
- baeldung: this will get stale very quickly, and if we have to link a tutorial on everything we use, this is going to be a big list - I don't wish to maintain that
- spring boot - I wouldn't link that docs (there are many other resources and again we're not here to tutorial all)
but I do think we need a section about how to change the database to mysql (= explain what to change in the application.properties which entails explaining spring boot a bit)
- optaplanner - the readme should start with a paragraph explaining what this app does - in that paragraph, I'd expect to see something like "optimized with OptaPlanner" and that last word is a link.

ge0ffrey:
Good start!

employee-rostering-backend/README.adoc
rsynek: Never heard of. Interestingly, the link gives me "Internal Server Error". Is there anything that must be done to on the service first?

Christopher-Chianelli: IIRC, @yurloc used these links in vehicle routing. @yurloc, thoughts?

rsynek: my bad, it's in the readme of a "frontend" module of VRP.

cuijulian: My mistake, david-dm.org shows `node.js` dependencies and does not belong in the backend documentation. I will remove it.

employee-rostering-backend/README.adoc
cuijulian: Good points, after reading over the docs with a new user's perspective, it did seem like a tutorial rather than a quick start guide! I agree with adding a section for changing the database. As for the app explanation at the top, I plan to keep it consistent with the [Vehicle Routing documentation](https://github.com/kiegroup/optaweb-vehicle-routing/blob/master/README.adoc#optaweb-vehicle-routing), ie. link the problem description to the OptaPlanner website.

PR #287, opened 2019-09-18
Generate roster for each test so tests do not get executed before roster is generated.

Christopher-Chianelli:


employee-rostering-backend/src/test/java/org/optaweb/employeerostering/generator/RosterGeneratorTest.java
```diff
@@ -87,69 +94,55 @@
         return restTemplate.exchange(tenantPathURI, HttpMethod.GET, null,
                                      new ParameterizedTypeReference<List<Tenant>>() {});
     }
+    
+    @Before
+    public void setup() {
+        tenantId = rosterGenerator.generateRoster(2, 7).getTenantId();
+    }
```
cuijulian: We should have a `@After` method that deletes the generated rosters so that the rosters don't remain after the tests.

cuijulian:
Other than my comment, this looks fine. Another concern is that it doesn't actually check for generated data from startup since a roster is generated before each test method. We could delay the tests to make sure that the data is generated before the tests are run, but that seems like bad practice.

PR #296, opened 2019-10-03
Migrate from GWT and Wildfly to React and Spring Boot

Christopher-Chianelli:
This is a culmination of several months of hard work to rewrite OptaWeb Employee Rostering. Each commit was reviewed in a PR to react-spring-boot-refactor.


rsynek:
@Christopher-Chianelli @cuijulian congratulations for reaching this point!

A couple of comments to the SonarCloud analysis:
- let's disregard the failed check on code duplication - after quickly reviewing it, seems like a couple of false positives
- 2 major issues suggested in a css file - feel free to have a look, otherwise looks good
- please review the suggested security vulnerabilities
- impressive job on getting coverage over 85% !

Christopher-Chianelli:
@rsynek the two issues in the CSS files are false (otherwise PatternFly itself will have issues); PatternFly provides a font family for its icons, which we use to display the icons in the headers. The security vulnerabilities have to do with `tenantId` being used in a response without being checked/sanitized, but I am pretty sure Spring Boot does that for us (after all, what is the integer value of "hello"?).

PR #300, opened 2019-10-07
PLANNER-1664: Benchmarker Application

cuijulian:
- Create Benchmark module with separate Persistence Context
- Add Benchmark startup docs
- Add Benchmarker integration test that verifies a report is generated


kie-ci2:
Build finished. 1494 tests run, 2 skipped, 0 failed.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/shift/Shift.java
```diff
@@ -33,7 +34,10 @@
 
 @Entity
 @PlanningEntity
-public class Shift extends AbstractPersistable {
+public class Shift extends AbstractPersistable implements Comparable<Shift> {
+
+    private static final Comparator<Shift> PILLAR_SEQUENCE_COMPARATOR = Comparator.comparing(Shift::getStartDateTime)
```
Christopher-Chianelli: `PILLAR_SEQUENCE_COMPARATOR` is not a good name (and I argue code for the benchmarker should be in the benchmarker). If you need to have a `compareTo`, a name like `TIME_SLOT_COMPARATOR` would be better.

cuijulian: The subpillar solver configuration requires the `Shift` class to implement the `compareTo()` method (like in the [NurseRostering OptaPlanner example](https://github.com/kiegroup/optaplanner/blob/master/optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/domain/ShiftAssignment.java#L118)). I don't see a way to separate this from the `Shift` class, which is in the backend. This method isn't necessarily Benchmark code, it just specifies a way to compare `Shift` objects, so I think it's fine in the backend module. WDYT?

Christopher-Chianelli: From the docs (https://docs.optaplanner.org/7.27.0.Final/optaplanner-docs/html_single/index.html#pillarMoveSelectors), there a way to specify a comparator WITHOUT needing to implement `compareTo`: ` <subPillarSequenceComparatorClass>org.optaplanner.examples.nurserostering.domain.ShiftAssignmentComparator</subPillarSequenceComparatorClass>`

README.adoc
```diff
@@ -49,6 +49,29 @@ To run on an another port use `--server.port=...`:
...
+
+After building the project above, run the benchmark application with:
+
+[source,shell]
+----
+java -jar employee-rostering-benchmark/target/employee-rostering-benchmark-*.jar
+----
+
+Note: make sure to run the project application and benchmark application on different ports if they are run
+simultaneously.
```
triceo: How about including the config to change the port in the example above?

README.adoc
```diff
@@ -49,6 +49,29 @@ To run on an another port use `--server.port=...`:
...
+
+After benchmarking is complete, a benchmark report will be generated in the `local/benchmarkReport` directory.
+Open the `index.html` file in your browser to view the benchmark results.
+
+=== Configure the benchmarker
+
+To configure the benchmarker, modify the `employeeRosteringBenchmarkConfig.xml` file found in the
+`employee-rostering-benchmark/src/main/resources` directory.
```
triceo: Am I correct in assuming that the user also needs to rebuild to get the changes in the JAR?

employee-rostering-benchmark/src/main/java/org/optaweb/employeerostering/OptaWebEmployeeRosteringBenchmarkApplication.java
```diff
@@ -14,45 +14,56 @@
...
         List<Roster> rosterList = generateRosters();
 
-        SolverFactory<Roster> solverFactory = SolverFactory.createFromXmlResource(
-                "org/optaweb/employeerostering/service/solver/employeeRosteringSolverConfig.xml");
-        PlannerBenchmarkFactory benchmarkFactory = PlannerBenchmarkFactory.createFromSolverFactory(solverFactory);
+        PlannerBenchmarkFactory benchmarkFactory = PlannerBenchmarkFactory.createFromXmlResource(
+                "employeeRosteringBenchmarkConfig.xml", getClass().getClassLoader());
         PlannerBenchmark plannerBenchmark = benchmarkFactory.buildPlannerBenchmark(rosterList);
-        plannerBenchmark.benchmark();
+        plannerBenchmark.benchmarkAndShowReportInBrowser();
```
triceo: Not 100 % sure, but isn't there a method that wouldn't try to open the browser? Somehow opening a browser in a Spring app doesn't feel right to me.

employee-rostering-benchmark/src/test/java/org/optaweb/employeerostering/BenchmarkTest.java
```diff
@@ -0,0 +1,58 @@
...
+            oldBenchmarkFilesInDirectory = new HashSet<>(Arrays.asList(benchmarkLocalDirectory.toFile().listFiles()));
+        }
+    }
+
+    @Test
+    public void isBenchmarkReportGeneratedTest() {
+        Optional<File> benchmarkReport = Arrays.stream(Paths.get("local/benchmarkReport").toFile().listFiles())
+                .filter(f -> !oldBenchmarkFilesInDirectory.contains(f))
+                .findAny();
+        assertTrue(benchmarkReport.isPresent());
```
triceo: By using `getOrElseThrow()` on the `Optional` returned by the stream, you will be ale to get rid of the first assertion as well as of the `Optional`.

employee-rostering-benchmark/src/main/resources/employeeRosteringBenchmarkConfig.xml
```diff
@@ -0,0 +1,171 @@
...
+        <secondsSpentLimit>10</secondsSpentLimit>
+      </termination>
+    </solver>
+  </inheritedSolverBenchmark>
+
+  <solverBenchmark>
+    <name>Default</name>
+    <solver>
+    </solver>
+  </solverBenchmark>
```
triceo: Personally, I'd add a comment saying that the block is intentionally empty. Otherwise people will wonder.

employee-rostering-benchmark/src/main/resources/employeeRosteringBenchmarkConfig.xml
```diff
@@ -0,0 +1,171 @@
...
+    <name>Pillar Change and Swap Move Selector</name>
+    <solver>
+      <localSearch>
+        <unionMoveSelector>
+          <pillarChangeMoveSelector/>
+          <pillarSwapMoveSelector/>
+        </unionMoveSelector>
+      </localSearch>
+    </solver>
+  </solverBenchmark>
```
triceo: In the interest of brevity, I'd remove these. Three benchmarks (without pillars, with and with sequential) should be enough to demonstrate the functionality.

employee-rostering-benchmark/src/main/resources/employeeRosteringBenchmarkConfig.xml
```diff
@@ -0,0 +1,171 @@
...
+      <localSearch>
+        <acceptor>
+          <lateAcceptanceSize>500</lateAcceptanceSize>
+        </acceptor>
+        <forager>
+          <acceptedCountLimit>1</acceptedCountLimit>
+        </forager>
+      </localSearch>
+    </solver>
+  </solverBenchmark>
```
triceo: I don't like the fact that we're mixing benchmarks for selectors with benchmarks for acceptors/foragers. If my memory serves me correctly, in `optaplanner-examples` we solve that by having different benchmark config files.

In this case, I think that doesn't help either. We want to show benchmarking capabilities, but not necessarily show everything that could be benchmarked. That's what the documentation is for.

In my opinion, if the config includes just the selector benchmarks, it will be better. It will show benchmarking is possible, but not overwhelm users.

README.adoc
cuijulian: There are instructions to run a jar on a different port above. Adding the config command again seems redundant.

README.adoc
cuijulian: That's correct. I'll add a note in the README.

employee-rostering-benchmark/src/main/java/org/optaweb/employeerostering/OptaWebEmployeeRosteringBenchmarkApplication.java
cuijulian: Yes, the `benchmark()` method doesn't try to open the browser. You're probably right, it does feel strange, I'll change it back.

employee-rostering-benchmark/src/main/resources/employeeRosteringBenchmarkConfig.xml
cuijulian: Fair enough. Removing these will also decrease testing time.

employee-rostering-benchmark/src/main/resources/employeeRosteringBenchmarkConfig.xml
```diff
@@ -0,0 +1,81 @@
...
+    <name>Move Selector</name>
+    <solver>
+      <localSearch>
+        <unionMoveSelector>
+          <changeMoveSelector/>
+          <swapMoveSelector/>
+        </unionMoveSelector>
+      </localSearch>
+    </solver>
+  </solverBenchmark>
```
triceo: Looking at it again, I think that these two configs do the same thing.

cuijulian: Ok I double checked with @ge0ffrey , the first configuration has a localSearch construction heuristic by default, and the second configuration does not. Is this a concern?

cuijulian: Ok given that construction heuristics should be run, there really is no point in the `Move Selector` configuration since it will be equivalent to `Default` if construction heuristics are added (it's also included in the Pillar Move Selector configs). It makes sense to remove this `Move Selector` config.

PR #302, opened 2019-10-09
Create source jar for frontend and distribution

Christopher-Chianelli:


mbiarnes:
@yurloc @ge0ffrey can you please approve this - although it won't be merged then (because
it is red)  I can cherry-pick this to the r7.28.0.Final branch and hopefully with this commit the Nexus server doesn't complain any more because of missing -sources.jar.
FYI: In community 7.28.0.Final the build was OK - but the upload of binaries to Nexus failed.

employee-rostering-distribution/src/main/assembly/assembly-optaweb-employee-rostering-sources.xml
```diff
@@ -6,6 +6,7 @@
   <formats>
     <format>dir</format>
     <format>zip</format>
+    <format>jar</format>
```
yurloc: @Christopher-Chianelli Now we have both `-sources.jar` and `-sources.zip` and they're almost identical (the JAR has `META-INF/MANIFEST.MF` in addition). I think we should remove the ZIP in the next PR.

mareknovotny:
I think we need to backport this change also to 7.26.x, is that right?

yurloc:
No. 7.26.x branch doesn't contain the refactoring work (GWT+WildFly->React+Spring Boot) so it's not affected by the missing sources JAR problem. The `employee-rostering-frontend` module doesn't exist on 7.26.x branch.

PR #309, opened 2019-10-16
Build improvements

yurloc:


employee-rostering-frontend/package.json
```diff
@@ -37,13 +37,21 @@
   "scripts": {
     "start": "mvn generate-sources; react-scripts start",
     "build": "mvn generate-sources; react-scripts build",
-    "typecheck": "npx tsc --noEmit",
-    "test": "TZ=UTC react-scripts test --transformIgnorePatterns \"node_modules/(?!(lodash-es))/\"",
+    "typecheck": "tsc --noEmit",
```
Christopher-Chianelli: I am assuming this is a deliberate error to verify it fails the build?

employee-rostering-frontend/pom.xml
```diff
@@ -29,16 +29,6 @@
...
-  <properties>
-    <!--
-      Intentionally leaving these properties empty so that the Frontend plugin uses default values and downloads
-      Node.js and npm from https://nodejs.org/dist/. Having these properties here makes it possible to change the URLs
-      using system properties. This is necessary for CI builds running on Red Hat infrastructure.
-    -->
-    <nodeDownloadRoot></nodeDownloadRoot>
-    <npmDownloadRoot></npmDownloadRoot>
-  </properties>
-
```
Christopher-Chianelli: Why is this removed?

yurloc: See commit description.

employee-rostering-frontend/package.json
yurloc: I know we've discussed this on the chat but just for the record: This will actually fix the build on Jenkins where `npx` seems to be missing. Without `npx` in the script, when running `npm run typecheck`, `npm` will look for `tsc` in `node_modules/.bin` where it is provided by the `typescript` dependency.

PR #312, opened 2019-10-29
Remove Spring Boot dependency management

cuijulian:
Depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1086

### DIFF
--- a/employee-rostering-backend/src/test/java/org/optaweb/employeerostering/contract/ContractServiceTest.java
+++ b/employee-rostering-backend/src/test/java/org/optaweb/employeerostering/contract/ContractServiceTest.java
@@ -141,7 +141,7 @@ public class ContractServiceTest extends AbstractEntityRequireTenantRestServiceT
                             .accept(MediaType.APPLICATION_JSON))
                 .andDo(mvcResult -> logger.info(mvcResult.toString()))
                 .andExpect(status().isOk())
-                .andExpect(content().contentType("application/json;charset=UTF-8"))
+                .andExpect(content().contentType("application/json"))
                 .andExpect(content().string("true"));
     }
 
@@ -151,7 +151,7 @@ public class ContractServiceTest extends AbstractEntityRequireTenantRestServiceT
                             .accept(MediaType.APPLICATION_JSON))
                 .andDo(mvcResult -> logger.info(mvcResult.toString()))
                 .andExpect(status().isOk())
-                .andExpect(content().contentType("application/json;charset=UTF-8"))
+                .andExpect(content().contentType("application/json"))
                 .andExpect(content().string("false"));
     }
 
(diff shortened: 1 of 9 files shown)


employee-rostering-backend/src/test/java/org/optaweb/employeerostering/contract/ContractServiceTest.java
```diff
@@ -141,7 +141,7 @@ public void deleteContractTest() throws Exception {
                             .accept(MediaType.APPLICATION_JSON))
                 .andDo(mvcResult -> logger.info(mvcResult.toString()))
                 .andExpect(status().isOk())
-                .andExpect(content().contentType("application/json;charset=UTF-8"))
+                .andExpect(content().contentType("application/json"))
```
rsynek: What caused the omission of charset in content type?

cuijulian: This was to fix the failing tests. Bumping to the latest Spring version resulted in a different expected content type, like in [this GitHub issue](https://github.com/spring-projects/spring-framework/issues/19041). Omitting the charset fixes this issue.

rsynek: Question is, can we afford omitting the charset? If the charset is not sent via HTTP header, how is some user's client code supposed to know how the JSON is encoded?

cuijulian: @rsynek Good point. I have an alternative that also passes the tests:

`.andExpect(content().contentTypeCompatibleWith("application/json;charset=UTF-8"))`

This way we keep the charset. What do you think of this alternative?

rsynek: Depends how the method works. Does the `contentTypeCompatibleWith` just relax the matching criteria to make it pass with content being encoded via UTF-8 (without the charset being sent via header)?

cuijulian: Hm, according to the [docs](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/test/web/servlet/result/ContentResultMatchers.html#contentTypeCompatibleWith-org.springframework.http.MediaType-) that seems to be the case. Unfortunately after the Spring api changes I don't see a better alternative. I'll see if there is another way to assert contentType. Do you have any recommendations?

cuijulian: @rsynek As this PR is part of a very time-sensitive group of PR's, would it be ok to merge this PR for now, and we can discuss the `contentType` tests later? I can create a Jira for this. Wdyt?

yurloc: @rsynek Valid concern but let's merge without the charset. It's been deprecated. See https://github.com/spring-projects/spring-framework/issues/22788 for explanation.

rsynek:
Travis fails due to its inability to build related PRs to other repositories; let's ignore it.

PR #313, opened 2019-10-30
BXMSPROD-560 lock-treatment-tool added to Maven build

Ginxo:
BXMSPROD-560

### DIFF
--- a/employee-rostering-frontend/pom.xml
+++ b/employee-rostering-frontend/pom.xml
@@ -53,6 +53,26 @@
             </configuration>
           </execution>
           <execution>
+              <id>npm install lock-treatment-tool and run-node</id>
+              <phase>initialize</phase>
+              <goals>
+                  <goal>npm</goal>
+              </goals>
+              <configuration>
+                  <arguments>install lock-treatment-tool --global-style --no-package-lock</arguments>
+              </configuration>
+          </execution>
+          <execution>
+              <id>lock-treatment-tool execution</id>
+              <phase>initialize</phase>
+              <goals>
+                  <goal>npm</goal>
+              </goals>
+              <configuration>
+                  <arguments>run env -- locktt</arguments>
+              </configuration>
+          </execution>
+          <execution>
             <id>npm install</id>
             <goals>
               <goal>npm</goal>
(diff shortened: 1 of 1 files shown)


Ginxo:
Jenkins retest this please

Christopher-Chianelli:
Do we only want to run it for the productize profile? I think it might be useful to have it in the normal build. Checks pass + I verified it works on my machine (well, not via this PR, but I pretty much added your changes to something else except I use the default profile)

employee-rostering-frontend/pom.xml
```diff
@@ -29,21 +29,27 @@
   <name>Employee Rostering as a Service React Frontend</name>
   <description>React frontend for Employee Rostering as a Service</description>
 
+  <properties>
+    <frontend-maven-plugin.version>1.7.5</frontend-maven-plugin.version>
+    <node.version>v8.12.0</node.version>
+    <npm.version>6.9.0</npm.version>
+</properties>
```
Christopher-Chianelli: wrong identation

Ginxo:
@Christopher-Chianelli lock-treatment-tool moved to the default build and indentation fixed

Christopher-Chianelli:
@Ginxo still wrong indentation on line 36. Everything else looks good. Will this set of PR's be merged as one or individually?

Ginxo:
@Christopher-Chianelli fixed, sorry for that

mareknovotny:
Jenkins retest this please

Ginxo:
@mareknovotny moved to default profile.
The same done for optaweb-vehicle-routing https://github.com/kiegroup/optaweb-vehicle-routing/pull/218

Ginxo:
Jenkins retest this please

Ginxo:
@Christopher-Chianelli @mareknovotny I don't see any error in the travis log and build result is success

Christopher-Chianelli:
@Ginxo you need to remove `package-lock.json`, then run `mvn clean install`, then commit the updated `package-lock.json` file (Travis fails if it detects running an install cause package-lock changes and those changes are not commited).

employee-rostering-frontend/pom.xml
```diff
@@ -29,25 +29,53 @@
...
             </goals>
             <configuration>
-              <nodeVersion>v8.12.0</nodeVersion>
-              <npmVersion>6.9.0</npmVersion>
+              <nodeVersion>${node.version}</nodeVersion>
+              <npmVersion>${npm.version}</npmVersion>
             </configuration>
           </execution>
+          <execution>
+              <id>npm install lock-treatment-tool and run-node</id>
```
yurloc: Wrong indentation. Please keep the formatting consistent.

Ginxo:
I see, two things were happening here:

- the `--no-save` flag absence when installing lock-treatment-tool. This was adding the library to dependencies and as a result the package-lock.json changes once the `install` is executed.
- The `mvn generate-sources` command executed again the 'lock-treatment-tool execution' execution step. Moved to 'compile' phase which I don't like but it's the only way (I'm able to see) to skip it.

@yurloc I think I've solved my indentation problems for the rest of my life ;) (I hope so)
And about your question, @mareknovotny proposed my to keep all the projects aligned and the default profile is the chosen one.

Christopher-Chianelli:
@yurloc another advantage of the lock-treatment tool being executed in the default profile: the lock file will no longer contain urls; only version numbers and integrity.

Christopher-Chianelli:
@Ginxo I recommend updating `package-lock.json`, since it will be updated anyways whenever we add a new dependency (note: old entries will have their URL's removed)

Ginxo:
@Christopher-Chianelli the package-lock.json is up to date since the lock-treatment-tool library is installed - -global-style and this is the way we should. 
Travis green

Christopher-Chianelli:
I am not sure if it is working now; I did a `mvn clean install` with these commit applied and the url were still added to the package-lock.json; I needed to run `npm run env -- locktt` after the install to remove those fields. We might want to move that execution to the `package` phase.

PR #316, opened 2019-10-31
PLANNER-1636: Add REST API Reference button, enable CORS on backend.

Christopher-Chianelli:


Christopher-Chianelli:
@yurloc this PR might make the docker config easier; enable CORS on the backend and allow setting rest endpoint location (which mean we can probably remove the proxy config)

employee-rostering-frontend/src/index.tsx
```diff
@@ -29,7 +29,7 @@ import { I18nextProvider } from 'react-i18next';
 import { configureStore } from 'store';
 
 const store = configureStore({
-  restBaseURL: '/rest'
+  restBaseURL: `${process.env.REACT_APP_BACKEND_URL}/rest`
```
yurloc: +1 This is useful during development. I use this approach in VRP so that I don't have to use proxy feature of the development server, and let frontend talk directly to backend. I can then start backend on different port and change REACT_APP_BACKEND_URL in `.env.development.local` without having to touch package.json where the development server proxy can be configured.

employee-rostering-frontend/src/ui/header/Toolbar.tsx
```diff
@@ -86,23 +86,33 @@ export class ToolbarComponent extends React.Component<Props, ToolbarState> {
...
             variant={ButtonVariant.plain}
             onClick={() => this.props.history.push("/admin")}
           >
             <CogIcon />
           </Button>
         </ToolbarItem>
+        <ToolbarItem>
+          <a
+            aria-label="REST Reference"
+            href={`${process.env.REACT_APP_BACKEND_URL}/swagger-ui.html`}
```
yurloc: Good idea. I want similar info for VRP.

employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/admin/AdminController.java
```diff
@@ -22,12 +22,14 @@
...
 import org.springframework.util.Assert;
 import org.springframework.validation.annotation.Validated;
+import org.springframework.web.bind.annotation.CrossOrigin;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 
 @RestController
 @RequestMapping("/rest/admin")
+@CrossOrigin
```
yurloc: Can you explain/demo live why this is needed. I know I thought I needed this in the early days of optaweb VRP but I somehow made it work without this. I can't say I've never seen browser complain about CORS since then. It happened to me once or twice but I have no idea why.

I mean, it makes sense to have to enable CORS when frontend and backend run on different hosts but on my machine it just works with no configuration. Maybe it doesn't apply for localhost? If you could demo the CORS issue it would be great.

Christopher-Chianelli: Techically, `128.0.0.1:8080` is a different origin than `128.0.0.1:3000` (different ports). The browser will detect this and say the request fails (even for REST request). My guess why you didn't need this for vehicle routing is because you use websockets, which I don't think have any features like CORS.

Christopher-Chianelli: Just tested again without CORS; can confirm it is needed for npm start, Firefox gives the following error:
`Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://localhost:8080/rest/tenant/. (Reason: CORS header “Access-Control-Allow-Origin” missing)`.
Note the request was successful; Firefox just ignores reading it and reports it as a failure.

When using the distribution zip, CORS is not needed anymore (probably because frontend and backend exist in the same server).

Using docker backend, CORS is probably not needed (since ngrix proxies the requests).

So we do need CORS for `npm start`, which is what most developers will be using.

PR #319, opened 2019-11-06
PLANNER-1116: Bookmarkable URL's

Christopher-Chianelli:
Draft; currently added backend support for frontend URL when packaged together. Currently need to modify frontend so page state is stored in the URL


Christopher-Chianelli:
@rsynek I think there a bug in `SonarCloud`: SonarCloud is saying `util/BookmarkableUtils.ts` has 0% test coverage, when it does in fact have tests.

rsynek:
@Christopher-Chianelli there must be either something special about the `BookmarkableUtils.ts` itself, or a bug in or a wrong configuration of how we measure the coverage.

If you run:
`npm run coverage` locally, you can see the `BookmarkableUtils.ts` is not covered there.

yurloc:
Of course there's something special about it.

employee-rostering-frontend/src/setupTests.ts
```diff
@@ -54,3 +54,4 @@ function mockFunctions() {
 }
 
 jest.mock('react-responsive', () => mockFunctions());
+jest.mock('util/BookmarkableUtils');
```
yurloc: It's mocked-out here so it's actually not used by tests.

https://jestjs.io/docs/en/manual-mocks

Christopher-Chianelli: But I unmocked it in `BookmarkableUtils.test.ts:
```
describe('Bookmarkable Utils', () => {
  beforeAll(() => {
    jest.unmock("util/BookmarkableUtils");
  });
  // ...
```

employee-rostering-frontend/src/util/BookmarkableUtils.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+import * as bookmarkableUtils from './BookmarkableUtils';
+import { getRouterProps } from './BookmarkableTestUtils';
+
+// Ugly workaround for having no way to wait for all requestAnimationFrame to finish
+const flushPromisesCallback = (cb: () => void) =>
+  window.requestAnimationFrame(() => new Promise(setImmediate).then(cb));
+describe('Bookmarkable Utils', () => {
+  beforeAll(() => {
+    jest.unmock("util/BookmarkableUtils");
```
yurloc: I may be wrong. The intention here to cancel mocking of `BookmarkableUtils` just for this test makes sense. But maybe it doesn't work as expected... @Christopher-Chianelli can you investigate and shed some light on the coverage problem?

Christopher-Chianelli: I will investigate

Christopher-Chianelli: Threw errors on first lines in BookmarkableUtils; can confirm it uses mock instead of actual

employee-rostering-frontend/spring-boot/java/org/optaweb/employeerostering/frontend/WebConfig.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package org.optaweb.employeerostering.frontend;
+
+import org.springframework.context.annotation.Configuration;
+import org.springframework.web.servlet.config.annotation.ViewControllerRegistry;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurerAdapter;
+
+@Configuration
+public class WebConfig extends WebMvcConfigurerAdapter {
```
yurloc: Why mixing backend Java code to the frontend project? I think this should go to the backend.

employee-rostering-frontend/spring-boot/java/org/optaweb/employeerostering/frontend/WebConfig.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Override
+    public void addViewControllers(ViewControllerRegistry registry) {
+        registry.addViewController("/{spring:\\w+}")
+              .setViewName("forward:/");
+        registry.addViewController("/**/{spring:\\w+}")
+              .setViewName("forward:/");
+        registry.addViewController("/{spring:\\w+}/**{spring:?!(\\.js|\\.css)$}")
```
yurloc: This is very difficult to understand without documentation or comments.
- What is `spring:`? Is it a keyword or variable name?
- What is `forward:`? Is it a keyword or a variable name?
- What does `{spring:?!(\\.js|\\.css)$}` mean?

employee-rostering-frontend/spring-boot/java/org/optaweb/employeerostering/frontend/WebConfig.java
Christopher-Chianelli: I choose to put this code here because it is specific to the react frontend. Different frontend frameworks may use different conventions for URL's, and may have multiple html files. This code is not needed for the backend, only for connecting the backend to the frontend.

employee-rostering-frontend/spring-boot/java/org/optaweb/employeerostering/frontend/WebConfig.java
Christopher-Chianelli: I got this from stack overflow: https://stackoverflow.com/a/42998817; probably should add the stack overflow link as a comment, although even the answerer admitted "To be honest I have no idea why it has to be exactly in this specific format to avoid infinite forwarding loop."

yurloc: Yes, please include the stack overflow link.

employee-rostering-frontend/spring-boot/java/org/optaweb/employeerostering/frontend/WebConfig.java
yurloc: Btw, the adapter is deprecated.

yurloc: You're right that from the backend perspective it's specific to the React fronted. At the same time it is true that from the frontend perspective, it's specific to the case where the frontend is server by the Spring Boot app. If I serve the frontend with NGNIX this piece of Spring Boot configuration is irrelevant.

Could you move it to the module where frontend and backend are packaged together?

Christopher-Chianelli: +1 Great idea!

Christopher-Chianelli:
@yurloc Moved standalone jar and routing code to the `employee-rostering-standalone` module.

Christopher-Chianelli:
(Is there a way to make sure this change doesn't break publishing?)

employee-rostering-frontend/src/util/BookmarkableTestUtils.ts
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
```
yurloc: 2019 (nitpick)

employee-rostering-frontend/src/util/BookmarkableUtils.test.ts
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
```
yurloc: 2019 (nitpick)

employee-rostering-frontend/src/util/BookmarkableUtils.ts
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
```
yurloc: 2019 (nitpick)

employee-rostering-frontend/src/util/__mocks__/BookmarkableUtils.ts
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
```
yurloc: 2019 (nitpick)

employee-rostering-standalone/src/main/java/org/optaweb/employeerostering/standalone/WebConfig.java
```diff
@@ -0,0 +1,52 @@
...
+        
+        // probably handles requests with trailing paths (/employees/0 but
+        // not /app)
+        registry.addViewController("/**/{spring:\\w+}")
+              .setViewName("forward:/");
+        
+        // I have no idea what this does; does react have javascript files
+        // with a $ suffix?
+        registry.addViewController("/{spring:\\w+}/**{spring:?!(\\.js|\\.css)$}") 
+              .setViewName("forward:/");
```
yurloc: I guess `$` here is not literal `$` but a meta-character that matches the end of the URL, for example `/app/style.css`. Seems suspicious. I don't think we want to forward such requests to `/index.html`.

Christopher-Chianelli: I see "?!" in the beginning of the second part, which could mean "An optional part that does NOT end in .js or .css". So `/app/icon.png` would match but `/app/code.js` would not. I will see what happens if I remove it.

yurloc: Found this: https://docs.spring.io/spring/docs/5.2.2.RELEASE/spring-framework-reference/web.html#mvc-ann-requestmapping-uri-templates.

> Spring MVC uses the PathMatcher contract and the AntPathMatcher implementation from spring-core for URI path matching.

Also true for [ViewControllerRegistry#addViewController()](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/servlet/config/annotation/ViewControllerRegistry.html#addViewController-java.lang.String-).

According to [AntPathMatcher](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/AntPathMatcher.html), the path patterns only allow basic rules (`?`, `*`, `**`) but **path variables are matched using [regexp](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html)**. `(?!X)` means "X, via zero-width negative lookahead". [Negative lookahead](http://blackwasp.co.uk/RegexLookahead.aspx) asserts that the pattern in the lookahead must not follow the text matched by the initial part of the pattern. So `/{spring:\\w+}/**{spring:?!(\\.js|\\.css)$}` matches paths that have at least two elements and the last one doesn't end with `.js` or `.css`.

At least this is my understanding based on the documentation and source code of `AntPathMatcher`. It would be really interesting to see a unit test demonstrating which URIs are actually matched by the last mapping pattern and which aren't. Also, it would be nice to investigate whether we need all of the three mappings or whether it could be simplified.

yurloc:
"Go to the ... page" buttons are now broken because the tenant ID is missing.

employee-rostering-frontend/pom.xml
```diff
@@ -28,6 +28,7 @@
   <artifactId>employee-rostering-frontend</artifactId>
   <name>Employee Rostering as a Service React Frontend</name>
   <description>React frontend for Employee Rostering as a Service</description>
+  <packaging>pom</packaging>
```
yurloc: This won't work. You need some kind of archive (jar, war, zip) you can consume in the standalone module.

Christopher-Chianelli: Couldn't that be done with a plugin though? (And if I recall correctly, there no built-in packaging type for zip). (Probably will readd the assembly plugin, and make it assemble a zip of build).

Christopher-Chianelli: Used a plugin to add a Zip artifact, which appears to work

employee-rostering-standalone/src/main/java/org/optaweb/employeerostering/standalone/WebConfig.java
Christopher-Chianelli: It manage to work with only the middle matcher (`registry.addViewController("/**/{spring:\\w+}")`)

employee-rostering-frontend/pom.xml
```diff
@@ -147,18 +148,19 @@
...
         <artifactId>maven-assembly-plugin</artifactId>
-        <configuration>
-          <descriptors>
-            <descriptor>assembly/sources.xml</descriptor>
-          </descriptors>
-          <appendAssemblyId>true</appendAssemblyId>
-        </configuration>
         <executions>
           <execution>
+            <id>package-sources</id>
```
yurloc: `package-sources`? Isn't it packaging the build output?

employee-rostering-frontend/src/store/roster/selectors.ts
```diff
@@ -24,22 +24,29 @@ import AvailabilityRosterView from 'domain/AvailabilityRosterView';
...
 
-export function isLoading(state: AppState) {
+export function isShiftRosterLoading(state: AppState) {
   return state.spotList.isLoading || state.employeeList.isLoading || state.skillList.isLoading
-    || state.contractList.isLoading || state.shiftRoster.isLoading || state.availabilityRoster.isLoading
+    || state.contractList.isLoading || state.shiftRoster.isLoading
     || state.rosterState.isLoading;
 }
 
+export function isAvaliablityRosterLoading(state: AppState) {
```
yurloc: Typo: avaliability

employee-rostering-frontend/pom.xml
yurloc: OK, looks good. Please remove the resource plugin, it's not needed anymore.

employee-rostering-frontend/src/store/tenant/reducers.ts
```diff
@@ -17,8 +17,14 @@
 import { ActionType, TenantData, TenantAction } from './types';
 import { withElement, withoutElement } from 'util/ImmutableCollectionOperations';
 
+const path = window.location.pathname;
+let windowTenantId: number | null = null;
+if (path.indexOf('/', 1) > 0) {
+  windowTenantId = parseInt(path.substring(1, path.indexOf('/', 1)));
```
yurloc: Not nice for 2 reasons:
1. Breaks one of the [core principles](https://redux.js.org/introduction/three-principles#changes-are-made-with-pure-functions) of Redux by making the reducer impure. [Reducers](https://redux.js.org/glossary#reducer) must be pure functions—functions that return the exact same output for given inputs.
2. Mixing browser-specific code into app state management code that's meant to be UI-agnostic (https://www.quora.com/Can-Redux-be-used-for-server-side-Node-js-without-any-UI).

I suggest moving this to `index.tsx` where you call `configureStore()` and pass this information as the `preloadedState` argument.

employee-rostering-frontend/src/ui/components/FilterComponent.test.tsx
```diff
@@ -26,48 +25,35 @@ interface MockData {
...
   it('should call filter and onChange on updateFilter', () => {
-    const filter = new FilterComponent<MockData>(props);
-    const predicate = jest.fn();
-    filter.setState = jest.fn();
-    (props.filter as jest.Mock).mockReturnValue(predicate);
-    filter.updateFilter("Test");
-    expect(props.filter).toBeCalled();
-    expect(props.filter).toBeCalledWith("Test");
+    const filter = shallow(<FilterComponent {...props} />);
+    filter.find("TextInput").simulate("change", "New Filter");
```
yurloc: Prefer `find(TextInput)` over `find("TextInput")`.

employee-rostering-frontend/src/ui/components/FilterComponent.test.tsx
```diff
@@ -26,48 +25,35 @@ interface MockData {
...
-    const filter = shallow(<FilterComponent {...props} />);
-    expect(toJson(filter)).toMatchSnapshot();
+    const myProps = { ...props, filterText: "" };
+    const filter = shallow(<FilterComponent {...myProps} />);
+    expect(filter).toMatchSnapshot();
   });
 
   it('should render correctly with text', () => {
     const filter = shallow(<FilterComponent {...props} />);
-    (filter.instance() as FilterComponent<MockData>).updateFilter("Test");
```
yurloc: Please remove the `MockData` interface, it becomes unused.

employee-rostering-frontend/src/store/tenant/reducers.ts
Christopher-Chianelli: 1. Reducer is still pure: this affects only the initial state, and the reducer only references current state + params.

2. That I am okay with

The problem with moving to `index.tsx` is there is no way to actually set initial state in `configureStore`, although using actions might work but be inefficent/make extra calls

employee-rostering-frontend/src/util/ImmutableCollectionOperations.ts
```diff
@@ -96,3 +97,39 @@ export function mapWithUpdatedElement<T extends DomainObject>(map: Map<number, D
   return mapWithElement(map, updatedElement);
 }
 
+// An immutable version of a collection with a lot of helpful methods
+export class Stream<T> {
```
yurloc: Time to switch to one of the libraries out there?
- http://staltz.github.io/xstream/
- https://rxjs.dev/guide/operators
- https://immutable-js.github.io/immutable-js/

Christopher-Chianelli: +1 to immutable-js; the other libraries solves different use cases (side note: might also invalidate the entire file ImmutableCollectionOperations).

yurloc: OK, I'll file a JIRA for that. Let's not add that to this PR.

employee-rostering-frontend/src/store/tenant/reducers.ts
yurloc: 1. You're right. I didn't realize that `windowTenantId` is assigned only once when the file is loaded and it will (probably) never change (would be better to make it `const` to make it obvious that it never changes). Note that `tenantReducer()` function does reference the initial state in the case when the state argument us `undefined`.

2. I thought the state can be (partially) hydrated using the `preloadedState` argument of [createStore()](https://redux.js.org/api/createstore).

employee-rostering-frontend/src/util/ImmutableCollectionOperations.ts
yurloc: https://issues.redhat.com/browse/PLANNER-1752

employee-rostering-frontend/src/store/tenant/reducers.ts
Christopher-Chianelli: In regards to 2., preloadedState require an entire state, no partials allowed unfortunately.

employee-rostering-frontend/pom.xml
Christopher-Chianelli: Yes; probably should rename it `package-build` (was copied from distribution)

employee-rostering-frontend/src/store/tenant/reducers.ts
yurloc: 2. Can be done (tried locally). Consider optional for this PR. I can submit the change later.

employee-rostering-frontend/src/util/BookmarkableUtils.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export type UrlProps<T extends string> = { [K in T]: string|null };
+
+const pathnameToQueryStringMap: Map<string, string> = new Map();
+
+export function setTenantIdInUrl(props: RouteComponentProps, tenantId: number) {
+  const endOfTenantId = props.location.pathname.indexOf('/', 1);
+  if (endOfTenantId !== -1) {
+    props.history.push(`/${tenantId}${props.location.pathname
+      .slice(props.location.pathname.indexOf('/', 1))}`);
```
yurloc: Can be replaced with `endOfTenantId`.

employee-rostering-frontend/src/util/__mocks__/BookmarkableUtils.ts
```diff
@@ -0,0 +1,67 @@
...
+
+beforeEach(() => pathnameToQueryStringMap.clear()); 
+
+export function setTenantIdInUrl(props: RouteComponentProps, tenantId: number) {
+  const endOfTenantId = props.location.pathname.indexOf('/', 1);
+  if (endOfTenantId !== -1) {
+    props.history.push(`/${tenantId}${props.location.pathname
+      .slice(props.location.pathname.indexOf('/', 1))}`);
+    props.location.pathname = `/${tenantId}${props.location.pathname
+      .slice(props.location.pathname.indexOf('/', 1))}`;
```
yurloc: Repeating the `indexOf()` call twice. Can be replaced with `endOfTenantId`.

employee-rostering-frontend/src/store/tenant/reducers.ts
yurloc: Actually, partials are allowed: https://github.com/reduxjs/redux/blob/037ea567e9808ef4ad281f3e19bb00c1e74587ba/index.d.ts#L211.

Christopher-Chianelli: Moved it to `index.tsx` (although beware: the partial you passed in is not merged with the initial state)

employee-rostering-frontend/src/store/store.ts
```diff
@@ -39,7 +39,7 @@ export interface StoreConfig {
 
 export function configureStore(
   { restBaseURL }: StoreConfig,
-  preloadedState?: AppState,
+  preloadedState?: Partial<AppState>,
```
yurloc: Should be `DeepPartial<AppState>`, I guess.

employee-rostering-frontend/src/index.tsx
```diff
@@ -28,8 +28,20 @@ import i18n from './i18n';
...
+}
+
 const store = configureStore({
-  restBaseURL: '/rest'
+  restBaseURL: '/rest',
+}, {
+  tenantData: {
+    currentTenantId: windowTenantId,
+    tenantList: [],
+    timezoneList: []
```
yurloc: Using `DeepPartial<AppState>` will allow to omit `tenantList` and `timezoneList` completely.

employee-rostering-frontend/src/store/store.ts
Christopher-Chianelli: Nope; that introduces bugs; the deep partials are not merged with the initial state set in the reducer

employee-rostering-frontend/src/index.tsx
Christopher-Chianelli: Nope; the deep partials are not merged (that what I had before, then it complain about tenantList being empty)

employee-rostering-frontend/src/store/store.ts
yurloc: Are you sure? Based on https://github.com/reduxjs/redux/issues/433#issuecomment-220123880 it should work when using `combineReducers()`.

yurloc: But the simple example from the comment doesn't actually use deep partials... so that doesn't prove it works in our case.

Christopher-Chianelli: I tried it at first, and end up with undefined, so I switch to using Partials.

employee-rostering-frontend/src/index.tsx
yurloc: Makes sense. Thanks for clearing this up.

employee-rostering-frontend/.env
```diff
@@ -1 +1 @@
-REACT_APP_BACKEND_URL=.
+REACT_APP_BACKEND_URL=
```
yurloc: Was the `.` the reason why Cypress test failed?

Christopher-Chianelli: Yep; when set to '.', it uses the realitive url, which due to this PR, becomes something like '/{tenantId}/skills/rest/....'.

When set to ''. it uses the base, so it becomes '/rest/...'. It worked when I tested on standalone jar with this change.

Christopher-Chianelli: (also ran the cypress test locally, and it fixes the test)

Christopher-Chianelli:
It appears the Drools :: Eclipse :: Base has a compiler error causing Jenkins to fail:
```
18:37:53 [INFO] ------------------------------------------------------------------------
18:37:53 [INFO] Reactor Summary:
18:37:53 [INFO] 
18:37:53 [INFO] Drools and jBPM Tools multiproject ................. SUCCESS [  1.546 s]
18:37:53 [INFO] Drools, jBPM and Guvnor plugins for Eclipse ........ SUCCESS [  0.939 s]
18:37:53 [INFO] KIE :: Eclipse :: Common ........................... SUCCESS [  3.065 s]
18:37:53 [INFO] Drools :: Eclipse :: Base .......................... FAILURE [  4.449 s]
18:37:53 [INFO] Drools :: Eclipse :: Base feature .................. SKIPPED
18:37:53 [INFO] Drools :: Eclipse :: Base test ..................... SKIPPED
18:37:53 [INFO] Drools :: Eclipse :: Webdav for Guvnor tools ....... SUCCESS [  3.066 s]
18:37:53 [INFO] Drools :: Eclipse :: Guvnor tools .................. SUCCESS [  1.416 s]
18:37:53 [INFO] Drools :: Eclipse :: Guvnor tools feature .......... SUCCESS [  0.783 s]
18:37:53 [INFO] Drools :: Eclipse :: Guvnor tools test ............. SUCCESS [  0.970 s]
18:37:53 [INFO] jBPM :: Eclipse :: Base ............................ SKIPPED
18:37:53 [INFO] jBPM :: Eclipse :: Base feature .................... SKIPPED
18:37:53 [INFO] Drools :: Eclipse :: Common feature ................ SUCCESS [  0.545 s]
18:37:53 [INFO] KIE :: Eclipse :: Kie Navigator .................... SKIPPED
18:37:53 [INFO] Drools :: Eclipse :: Kie Navigator feature ......... SKIPPED
18:37:53 [INFO] org.jboss.tools.runtime.drools.detector ............ SKIPPED
18:37:53 [INFO] org.jboss.tools.runtime.drools.detector.feature .... SKIPPED
18:37:53 [INFO] Drools :: Eclipse :: Update site ................... SKIPPED
18:37:53 [INFO] ------------------------------------------------------------------------
18:37:53 [INFO] BUILD FAILURE
18:37:53 [INFO] ------------------------------------------------------------------------
18:37:53 [INFO] Total time: 05:07 min (Wall Clock)
18:37:53 [INFO] Finished at: 2019-12-19T13:37:53-05:00
18:37:54 [INFO] Final Memory: 190M/962M
18:37:54 [INFO] ------------------------------------------------------------------------
18:37:54 [ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:1.0.0:compile (default-compile) on project org.drools.eclipse: Compilation failure: Compilation failure: 
18:37:54 [ERROR] /home/jenkins/workspace/KIE/master/pullrequest/optaweb-employee-rostering-pullrequests/upstream-repos/droolsjbpm-tools/drools-eclipse/org.drools.eclipse/src/main/java/org/drools/compiler/commons/jci/compilers/EclipseJavaCompiler.java:[388] 
18:37:54 [ERROR] 	final CompilationProblem problem = new EclipseCompilationProblem(iproblem);
18:37:54 [ERROR] 	                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
18:37:54 [ERROR] The constructor EclipseCompilationProblem(IProblem) is undefined
18:37:54 [ERROR] 1 problem (1 error)
18:37:54 [ERROR] -> [Help 1]
```

yurloc:
Yes. @mariofusco broke the build. Unfortunately we won't be able to fix it today. I'll try to make sure it's green again tomorrow so we could merge this this year.

yurloc:
Jenkins retest this.

yurloc:
@Christopher-Chianelli looks good now. Please go ahead and merge.

PR #320, opened 2019-11-08
Move locktt execution to package; updated package-lock.json

Christopher-Chianelli:
I moved locktt execution to the package phase so it will be executed after the install (and thus the urls will actually be removed), and updated the package-lock.json so the URL are removed (even though the versions are the same)

### DIFF
--- a/employee-rostering-frontend/package-lock.json
+++ b/employee-rostering-frontend/package-lock.json
@@ -4,16 +4,12 @@
   "dependencies": {
     "@babel/code-frame": {
       "version": "7.5.5",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.5.5.tgz",
-      "integrity": "sha512-27d4lZoomVyo51VegxI20xZPuSHusqbQag/ztrBC7wegWoQ1nLREPVSKSW8byhTlzTKyNE4ifaTA6lCp7JjpFw==",
       "requires": {
         "@babel/highlight": "^7.0.0"
       }
     },
     "@babel/core": {
       "version": "7.5.5",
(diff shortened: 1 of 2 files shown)


Ginxo:
@Christopher-Chianelli the locktt needs to be executed before the install, otherwise it will fail in case we want to use a different npmregistry.

Ginxo:
I see the confusion from https://github.com/kiegroup/optaweb-employee-rostering/pull/313#issuecomment-551895141.
The life cycle is:

1. npm installation
2. locktt installation
3. locktt execution -> resolved and integrity fields removed
4. npm install -> resolved and integrity fields written based on the npm registry

In case the package-lock.json and "npm install" registries are the same (which is probably your case), the package-lock.json file should remain the same.

Christopher-Chianelli:
Registries are not the same since we are forced to use a proxy for downloads, meaning we always need to check if package-lock.json is updated, and if so, run a sed command to replace the registry. I guess I can wait for the `locktt` update that adds replacing registries. I don't think the install will fail if you use different registries if the package-lock.json doesn't have registries to begin with.

Ginxo:
@Christopher-Chianelli Ok, I will upgrade locktt during today, I let you know once it's updated. My question is, should we upload the lock files with the proxy we are using? (just asking)

Ginxo:
@Christopher-Chianelli the lock-treatment-tool@0.4.0 published.
adding `--registry=https://registry.redhat.com -p` the package-lock.json registry will be replaced.

Christopher-Chianelli:
@Ginxo No, we should not upload the lock file with the proxy we are using, since then other people outside redhat who use this repo will not be able to compile since they do not have VPN access. Thanks for the update; I will add a git pre-commit hook that runs it. Closing PR since I now understand the issue that `locktt` fixes.

PR #321, opened 2019-11-12
PLANNER-1708 Run Cypress tests in CI

rsynek:
- starts the distribution as a separate process
- runs cypress tests from a docker container to isolate its dependencies from the test environment

### DIFF
--- a/employee-rostering-distribution/pom.xml
+++ b/employee-rostering-distribution/pom.xml
@@ -24,18 +24,6 @@
   </properties>
 
   <build>
-    <pluginManagement>
-      <plugins>
-        <plugin>
-          <groupId>com.bazaarvoice.maven.plugins</groupId>
-          <artifactId>process-exec-maven-plugin</artifactId>
-          <version>0.9</version>
-          <configuration>
-            <skip>${skipITs}</skip>
-          </configuration>
-        </plugin>
-      </plugins>
-    </pluginManagement>
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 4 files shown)


rsynek:
The failure in Travis job is caused by instability of the UI test itself, see the skill select box on the screenshot:

A new tenant can be created, who can have their own employees, spots, skills and shifts -- basic tenant workflow (failed)

employee-rostering-distribution/pom.xml
```diff
@@ -103,6 +122,75 @@
...
+        <executions>
+          <execution>
+            <id>run-cypress-tests</id>
+            <phase>integration-test</phase>
+            <goals>
+              <goal>exec</goal>
+            </goals>
+            <configuration>
+              <skip>${skipITs}</skip>
+              <executable>docker</executable>
```
Christopher-Chianelli: I am assuming the test will fail if we don't have docker installed on our system? (I prefer `podman`, which I have a `docker` alias for)

rsynek: Correct - and this is exactly the behaviour I count on in the CI. On some machines there might be docker, on others there is podman and docker alias for it.

Christopher-Chianelli: Might want to add "if you don't have `docker` or `podman` with a `docker` alias installed on your system, pass `-DskipITs` to maven" so a developer without `docker`/`podman` (they exist) doesn't get confused about why IT tests are failing.

Christopher-Chianelli: (to `README.md`); does `-DskipTests` imply `DskipITs`?

rsynek: No, `skipITs` is reserved for integration tests (e.g. by maven failsafe plugin). I noticed that in the README.md we skipTests, so I can add skipITs there too.

yurloc: Maybe use [fabric8io/docker-maven-plugin](https://dmp.fabric8.io/) or https://github.com/spotify/docker-maven-plugin instead of exec?

rsynek: Unfortunately the plugin required docker service to be running. Instead, I made the container runtime we use configurable, so that it's possible to use podman as well.

@yurloc thanks a lot for testing this with a podman and a docker alias.

rsynek:
The unstable test has been turned off by default. If there are no more comments, we can merge when the builds pass.

yurloc:
Follow-up: https://issues.jboss.org/browse/PLANNER-1715

PR #322, opened 2019-11-13
Remove a benchmark config that cause GC Overhead exception ONLY on my machine

Christopher-Chianelli:
The odd thing is my machine is fairly powerful:

- 12 cores, 24 threads
- 32 GB RAM

And it works on machines with less cores and less RAM than it.

VM max memory (as in -Xmx but lower): 954,728,448 bytes


Christopher-Chianelli:
Issue fixed by #323

PR #325, opened 2019-11-14
PR-321 Fix

Christopher-Chianelli:
@rsynek This might fix the test in #321


Christopher-Chianelli:
That didn't work; I don't have access to the video and cannot reproduce it on my machine so I cannot even see what went wrong

PR #326, opened 2019-11-15
Make http://backend:8080 default BACKEND_URL in frontend Dockerfile

yurloc:
OpenShift Online complains about a build with docker strategy for the
frontend. However backend build with the same strategy is OK. The only
difference is that frontend build defines an environment variable with
backend URL. The attempt to create the build failed with:

    error: failed to set env: admission webhook "validate.build.create" denied
       the request: Builds with docker strategy are prohibited on this cluster

Removing the enviroment variable makes the build work (even thought it
still uses the docker strategy ¯\_(ツ)_/¯). Might be an OpenShift bug.

The same error as above is produced by:

    oc set env bc/frontend BACKEND_URL=http://backend:8080

so maybe it's just that docker builds can't have environment variables.

### DIFF
--- a/employee-rostering-frontend/docker/Dockerfile
+++ b/employee-rostering-frontend/docker/Dockerfile
@@ -1,7 +1,7 @@
 FROM docker.io/library/nginx:1.17.5
 COPY nginx.conf /etc/nginx
 COPY default.conf /tmp/default.template
-ARG BACKEND_URL
+ARG BACKEND_URL=http://backend:8080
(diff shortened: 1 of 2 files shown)


Christopher-Chianelli:
Side note: the environment variable probably will have no effect until https://github.com/kiegroup/optaweb-employee-rostering/pull/316 is merged

yurloc:
This is unrelated to #316. Here, `BACKEND_URL` is an argument of the Dockerfile and its value is used in `/etc/nginx/conf.d/default.conf` inside the image.

Christopher-Chianelli:
Ah okay; it just shares a similar name as the variables in the .env files

PR #335, opened 2019-11-25
Update architecture images in community docs

cuijulian:
Jira: https://issues.jboss.org/browse/PLANNER-1717


ge0ffrey:
@cuijulian Good work. Do note that for the architecture slide: we don't use Node.js, we use NPM to build, but we don't actually launch a Node.js server.

cuijulian:
@ge0ffrey Good catch, thanks for the tip. Here is a PR for that fix: https://github.com/kiegroup/optaweb-employee-rostering/pull/344

PR #336, opened 2019-11-26
BXMSPROD-597: frontend project locktt execution improved

Ginxo:
BXMSPROD-597

* lock-treatment-tool added to dev-dependencies
* locktt script added
* pom execution changed to npm run locktt


Christopher-Chianelli:
It "works" on windows, but the build now fails to other reasons, which is addressed by https://github.com/kiegroup/optaweb-employee-rostering/pull/339. Will merge when it passes.

PR #341, opened 2019-11-28
PLANNER-1732: Give Airbnb config highest priority

yurloc:
https://issues.jboss.org/browse/PLANNER-1732

For code style and project configuration consistency it is important to unify `.eslintrc.json` between Vehicle Routing and Employee Rostering.

Some time ago I have introduced the airbnb ESLint config (#251) but I've given it a low priority to avoid introducing big changes. Now's the time to take a step further and give it the top priority.

In the next step I'd like to eliminate most of the exceptions to the style (disabled rules).

### DIFF
--- a/employee-rostering-frontend/cypress/plugins/index.js
+++ b/employee-rostering-frontend/cypress/plugins/index.js
@@ -27,14 +27,11 @@
 // This function is called when a project is opened or re-opened (e.g. due to
 // the project's config changing)
 
-/* eslint-disable global-require */
-/* eslint-disable import/no-extraneous-dependencies */
+// eslint-disable-next-line no-unused-vars,@typescript-eslint/no-unused-vars
 module.exports = (on, config) => {
   // `on` is used to hook into various events Cypress emits
   // `config` is the resolved Cypress config
   on('task', {
     failed: require('cypress-failed-log/src/failed')(),
-  })
-}
-/* eslint-enable global-require */
-/* eslint-enable import/no-extraneous-dependencies */
+  });
+};
--- a/employee-rostering-frontend/src/ui/Alerts.tsx
+++ b/employee-rostering-frontend/src/ui/Alerts.tsx
@@ -45,10 +45,8 @@ const mapDispatchToProps: DispatchProps = {
 export type Props = StateProps & DispatchProps;
 
 export function mapToComponent(component: AlertComponent, componentProps: BasicObject): React.ReactNode {
-  switch (component) {
-    case AlertComponent.SERVER_SIDE_EXCEPTION_DIALOG: {
-      return <ServerSideExceptionDialog {...componentProps as unknown as ServerSideExceptionInfo} />
-    }
+  if (component === AlertComponent.SERVER_SIDE_EXCEPTION_DIALOG) {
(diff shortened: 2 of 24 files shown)


employee-rostering-frontend/cypress/plugins/index.js
```diff
@@ -1,19 +1,3 @@
-/*
```
Christopher-Chianelli: What happened to the license?

yurloc: Because those were generated files - we're not the author => no need to license them. The license was only added to make ESLint happy. In my opinion it's better to disable ESLint completely for those folders (until we start adding code there) but keep the generated examples because:
- they're good examples, they might spark a new ideas when facing Cypress configuration issues,
- they would be generated on each Cypress run so we need to either gitignore them or check them in. The second option is better in my opinion.

Christopher-Chianelli: We added lines there (`require('cypress-failed-log');`) and ```
  on('task', {
    failed: require('cypress-failed-log/src/failed')(),
  })
```

Comment (Christopher-Chianelli) [reply employee-rostering-frontend/cypress/plugins/index.js]: (those lines show actions when a cypress task fail in the log)

**Comment by yurloc** [comment]
- Reverted `.eslintignore` => support and plugins are checked again.
- Added `cypress/**` under `devDependencies` of the `import/no-extraneous-dependencies` rule.
- Disabled `global-require` rule for cypress dir.

**Comment by yurloc** [comment]
Jenkins retest this.

**Comment by yurloc** [comment]
Added test coverage for the new default cases.

File: employee-rostering-frontend/src/ui/Alerts.tsx [hunk 2]
```diff
@@ -45,10 +45,8 @@ const mapDispatchToProps: DispatchProps = {
 export type Props = StateProps & DispatchProps;
 
 export function mapToComponent(component: AlertComponent, componentProps: BasicObject): React.ReactNode {
-  switch (component) {
-    case AlertComponent.SERVER_SIDE_EXCEPTION_DIALOG: {
-      return <ServerSideExceptionDialog {...componentProps as unknown as ServerSideExceptionInfo} />
-    }
+  if (component === AlertComponent.SERVER_SIDE_EXCEPTION_DIALOG) {
+    return <ServerSideExceptionDialog {...componentProps as unknown as ServerSideExceptionInfo} />;
   }
```
Christopher-Chianelli: Is "undefined" a `React.ReactNode`? If not, I don't think it will pass a typecheck, unless typescript sees "SERVER_SIDE_EXCEPTION_DIALOG" is the only `AlertComponent`, in which TypeScript should complain about an if that is always true.

yurloc: Good question. Yes it is.
```ts
    type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;
```

yurloc: Currently, there's only one `AlertComponent`:
```ts

export enum AlertComponent {
  SERVER_SIDE_EXCEPTION_DIALOG = 'SERVER_SIDE_EXCEPTION_DIALOG'
}
```
but it's still technically possible for the condition to be `false` in runtime.

PR #346, opened 2019-11-29
Added YAML parser to frontend so we don't need to convert YAML to JSON before build

Christopher-Chianelli:


Christopher-Chianelli:
jenkins execute full downstream build

employee-rostering-frontend/package-lock.json
```diff
@@ -2071,6 +2071,12 @@
       "resolved": "https://registry.npmjs.org/@types/stack-utils/-/stack-utils-1.0.1.tgz",
       "integrity": "sha512-l42BggppR6zLmpfU6fq9HEa2oGPEI8yrSPL3GITjfRInppYFahObbIQOQK3UGxEnyQpltZLaPe75046NOZQikw=="
     },
+    "@types/yaml": {
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/@types/yaml/-/yaml-1.2.0.tgz",
+      "integrity": "sha512-GW8b9qM+ebgW3/zjzPm0I1NxMvLaz/YKT9Ph6tTb+Fkeyzd9yLTvQ6ciQ2MorTRmb/qXmfjMerRpG4LviixaqQ==",
+      "dev": true
+    },
```
ge0ffrey: I always wonder: should package-lock.json be part of .gitignore or not? @yurloc

yurloc: Without package-lock.json, builds wouldn't be reproducible.

> It describes the exact tree that was generated, such that subsequent installs are able to generate identical trees, regardless of intermediate dependency updates.
>
> This file is intended to be committed into source repositories

https://docs.npmjs.com/files/package-lock.json

PR #353, opened 2019-12-16
Update node version to v11.6.0

michaltomco:
This is needed for the w2k16 Jenkins jobs to build; the older repository doesn't contain win-x64/node.exe.


michaltomco:
Closed the branch because of trying to merge fork-master->master, which causes linux tests to fail.

PR #355, opened 2019-12-18
[7.30.x] Update node version to v11.6.0

michaltomco:
This is needed for the w2k16 Jenkins jobs to build; the older repository doesn't contain win-x64/node.exe.

### DIFF
--- a/employee-rostering-frontend/pom.xml
+++ b/employee-rostering-frontend/pom.xml
@@ -31,7 +31,7 @@
 
   <properties>
     <frontend-maven-plugin.version>1.7.5</frontend-maven-plugin.version>
-    <node.version>v8.12.0</node.version>
+    <node.version>v11.6.0</node.version>
     <npm.version>6.9.0</npm.version>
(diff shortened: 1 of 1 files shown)


krisv:
This seems a quite significant version bump, are we sure this is required for CR2 and this is safe to apply? @ge0ffrey

ge0ffrey:
Jiri's our node expert, and he seems ok with it.
My only concern is that optaweb-employee-rostering and optaweb-vehicle-routing on their 7.30.x branches are using the same node version.

ge0ffrey:
@Christopher-Chianelli Please make sure that master of optweb-employee rostering also uses node.version 11.6.0 (if it doesn't already)

yurloc:
For the record, vehicle-routing uses `<node.version>v11.6.0</node.version>` on 7.30.x.

Christopher-Chianelli:
@ge0ffrey can confirm master uses node version 11.6.0

PR #359, opened 2019-12-20
[7.30.x] Add locktt script to package.json

Christopher-Chianelli:
I noticed production integration tests are failing due to a missing script in package.json that is referenced from the productized profile. This PR simply adds the missing script.


yurloc:
This is incomplete. `locktt` would still fail on Windows because. We forgot to backport a whole PR #336. 
Replacing this with #360.

PR #365, opened 2020-01-17
README: rework for developer onboarding

LoneRifle:
- Add Quick Start section
- Add section on setting up developer environment

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -17,17 +17,6 @@ An end-to-end https://www.optaplanner.org/[OptaPlanner] web application ready to
 
 See https://www.youtube.com/watch?v=sOWC4qrXxFk[the video demo].
 
-== Development
-
-=== Backend
-
-Configure your IDE to run the main class, which is in
-`org.optaweb.employeerostering.OptaWebEmployeeRosteringApplication` in `employee-rostering-backend` repository.
-
-=== Frontend
-
-Run `npm install` and then `npm start` in the `optaweb-employee-rostering-frontend` directory.
-
 == Building
 
 Build the project with maven in the project's root directory:
@@ -149,3 +138,12 @@ oc start-build frontend --from-dir=docker --follow
 === Code formatter
 
 Both IntelliJ and Eclipse formatters are available here: https://github.com/kiegroup/droolsjbpm-build-bootstrap/tree/master/ide-configuration
+
+=== Backend
+
+Configure your IDE to run the main class, which is in
+`org.optaweb.employeerostering.OptaWebEmployeeRosteringApplication` in `employee-rostering-backend` repository.
+
+=== Frontend
+
+Run `npm install` and then `npm start` in the `optaweb-employee-rostering-frontend` directory.
--- a/README.adoc
+++ b/README.adoc
@@ -28,35 +28,13 @@ docker run -p 8080:8080 --rm -it -e SPRING_PROFILES_ACTIVE=dev rostering
 ----
 This takes a few minutes on the first time. The application will be automatically populated with test data.
 
-== Development
+You can then access the application on localhost:8080, once it finishes starting
 
+== Development
 
 === Backend
 
-==== Set up OpenJDK Java 8
-
-[source,shell]
-----
-$ brew tap AdoptOpenJDK/openjdk
-$ brew cask install adoptopenjdk8
-$ java -version
-openjdk version "1.8.0_232"
-OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_232-b09)
-OpenJDK 64-Bit Server VM (AdoptOpenJDK)(build 25.232-b09, mixed mode)
-$ which java
-/usr/bin/java
-----
-
-==== Add the following line to `~/.bash_profile`
-
-[source,shell]
-----
-export JAVA_HOME="/Library/Java/JavaVirtualMachines/adoptopenjdk-8.jdk/Contents/Home"
-----
-
-==== Configure your IDE
-
-Configure your IDE to use OpenJDK 1.8 and run the main class, which is in
+Configure your IDE to run the main class, which is in
 `org.optaweb.employeerostering.OptaWebEmployeeRosteringApplication` in `employee-rostering-backend` repository.
 
 You will need to specify this program argument for Spring Boot to run in development mode and provision the H2 in-memory database:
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -17,6 +17,55 @@ An end-to-end https://www.optaplanner.org/[OptaPlanner] web application ready to
...
+
+Install Docker and run the following on the command line:
+
+[source,shell]
+----
+docker build -t rostering .
+docker run -p 8080:8080 --rm -it -e SPRING_PROFILES_ACTIVE=dev rostering
+----
+This takes a few minutes on the first time. The application will be automatically populated with test data.
+
```
Christopher-Chianelli: Probably want to say something like: "You can access the application on localhost:8080."
```suggestion
You can access the application on localhost:8080 once the application finish starting.
```

README.adoc
```diff
@@ -17,6 +17,55 @@ An end-to-end https://www.optaplanner.org/[OptaPlanner] web application ready to
...
+== Development
+
+
+=== Backend
+
+==== Set up OpenJDK Java 8
+
+[source,shell]
+----
+$ brew tap AdoptOpenJDK/openjdk
```
Christopher-Chianelli: I don't think we want to include install instructions for OpenJDK, especially since the install instructions differs between operating systems (for instance, brew is mostly specific to MacOS). Maybe include a link to the [OpenJDK install page](https://openjdk.java.net/install/) or [AdoptOpenJDK Installation Page](https://adoptopenjdk.net/installation.html)?

README.adoc
```diff
@@ -17,6 +17,55 @@ An end-to-end https://www.optaplanner.org/[OptaPlanner] web application ready to
...
+
+Configure your IDE to use OpenJDK 1.8 and run the main class, which is in
+`org.optaweb.employeerostering.OptaWebEmployeeRosteringApplication` in `employee-rostering-backend` repository.
+
+You will need to specify this program argument for Spring Boot to run in development mode and provision the H2 in-memory database:
+
+`--spring.profiles.active=dev`
+
+=== Frontend
+Run `npm install` and then `npm run start` in the `employee-rostering-frontend` directory.
```
Christopher-Chianelli: You don't need to run "start"
```suggestion
Run `npm install` and then `npm start` in the `employee-rostering-frontend` directory.
```

README.adoc
LoneRifle: I agree - I'll drop this and the addition of the bash profile instructions - it's fair to assume that the would-be developer is already familiar with developing in Java

README.adoc
```diff
@@ -17,6 +17,34 @@ An end-to-end https://www.optaplanner.org/[OptaPlanner] web application ready to
...
+== Development
+
+=== Backend
+
+Configure your IDE to run the main class, which is in
+`org.optaweb.employeerostering.OptaWebEmployeeRosteringApplication` in `employee-rostering-backend` repository.
+
+You will need to specify this program argument for Spring Boot to run in development mode and provision the H2 in-memory database:
+
+`--spring.profiles.active=dev`
```
yurloc: Side discussion: I'm starting to think that configuring persistence for PostgreSQL in the default (production) profile was a mistake. My thinking was that production is the app's ultimate purpose so default configuration must target production (and file-based H2 database is certainly not meant for production, while PostgreSQL is). But now I realize that the app is being run 99% in development mode and having to provide `--spring.profiles.active=dev` every time doesn't make life much easier. So I'm thinking about moving file-based H2 configuration to the default (production) profile (see the contradiction?) and PostgreSQL to a postgresql profile. Haven't decided yet. What are your opinions guys, @Christopher-Chianelli, @cuijulian, @rsynek, @ge0ffrey.

Christopher-Chianelli: Could you put `spring.profiles.active=dev` in `application.properties` to achieve the same effect (using `-Dspring.profiles.active=production` should override it)?

cuijulian: @yurloc I agree with your point, it makes sense that most users try out the application with an in-memory db first, rather than production. I also think that using H2 as the default makes sense (ie. using the `dev` profile as the default), so @Christopher-Chianelli's suggestion makes sense.

ge0ffrey: " that the app is being run 99% in development mode and having to provide --spring.profiles.active=dev every time doesn't make life much easier." Absolutely. It increased the hurdle to get started. I experienced it on Friday - the lack of spring profile dev - and I'd really like to see the readme with a short, simple command that just builds and runs the app locally (without a need to install docker etc).

rsynek: @yurloc that's a good point, the getting started should be as simple as possible.

LoneRifle:
> I don't know. The root Dockerfile wasn't intended for a general purpose usage. It's for a specific OpenShift deployment scenario. I'm trying the Quick Start section locally and it takes ages to build the image because the Maven build run inside Docker can't reuse local `~/.m2/repository/`. And I suspect that subsequent docker builds won't be faster.

I largely agree, though I would say that it is quick in that only a single command has to be entered to get something up and running, which is useful for a developer starting out for the first time

LoneRifle:
If it assuages concerns, I am willing to drop the section on Quick Start and limit the PR to talking about dev setup. We can always revisit how we want to document the quickest way to get up and running at a later point

Christopher-Chianelli:
@LoneRifle the development section is definitely mergeable on its own. I think the reason why this PR isn't merge yet is we are debating if we want Docker to be an official way to quick start the application. We already have a section on how to build and run the application using `mvn` and `java` in the README (which has the advantage that it will reuse the .m2 cache). Moreover, as @yurloc already said, The root Dockerfile wasn't intended for a general purpose usage. If you remove the section on using Docker to Quick Start the application, I would merge the PR, and we could discuss using Docker to Quick Start the application in a separate issue/PR.

README.adoc
```diff
@@ -17,33 +17,16 @@ An end-to-end https://www.optaplanner.org/[OptaPlanner] web application ready to
...
 
 === Backend
 
 Configure your IDE to run the main class, which is in
 `org.optaweb.employeerostering.OptaWebEmployeeRosteringApplication` in `employee-rostering-backend` repository.
 
-You will need to specify this program argument for Spring Boot to run in development mode and provision the H2 in-memory database:
-
-`--spring.profiles.active=dev`
-
```
Christopher-Chianelli: I think this is still required (although we are working on changing that)

LoneRifle: Was that not addressed with #379?

Christopher-Chianelli: Thanks for pointing to the PR; that appears to fix the issue (the commit message was less than obvious when I was looking at the commit history).

LoneRifle:
Ah, thanks for the catch - our own README was rejigged and I failed to take that into account

PR #366, opened 2020-01-17
PLANNER-1783,PLANNER-1785: Add standalone JAR to distribution

yurloc:


employee-rostering-distribution/src/main/assembly/filtered-resources/README.adoc
```diff
@@ -0,0 +1,30 @@
+= Welcome to OptaWeb Employee Rostering
```
Christopher-Chianelli: I am not sure about using .adoc in the distribution, since it is not that common of a format for people that are not developers and just want to run the application

yurloc: Yes, but the Asciidoc source is still pretty readable and some "markup" (lists, headings, etc.) is needed anyway. Plus, if you open the file in Vim or an IDE (and I believe most of our users will be developers) you'll get some nice syntax highlighting.

Christopher-Chianelli: What happens when you double click .adoc in Windows? Does it open it automatically in text editor, or does it say something like "Windows doesn't recongize the .adoc filetype"

yurloc: Good point. That doesn't work on Windows.

PR #369, opened 2020-01-20
Clean redundant configuration of code coverage measurement

rsynek:
Locally generated reasonable jacoco.exec files, let's see what shows up in SonarCloud.


.travis.yml
```diff
@@ -17,7 +17,7 @@ cache:
 install: mvn de.qaware.maven:go-offline-maven-plugin:1.2.1:resolve-dependencies -Pjacoco,sonar --quiet --show-version
 # do not install to avoid dirtying the cache
 script:
-  - mvn verify -Pjacoco,sonar -Dintegration-tests=true --show-version
+  - mvn verify -Prun-code-coverage,sonarcloud-analysis -Dintegration-tests=true --show-version
```
yurloc: Please replicate this change 3 lines above.

pom.xml
```diff
@@ -143,44 +143,22 @@
...
+      <!--
+        TODO: For moving the coverage measurement from Travis to Jenkins, add the repository name to
+        SONARCLOUD_ENABLED_REPOSITORIES in
+        https://github.com/kiegroup/kie-jenkins-scripts/blob/master/job-dsls/jobs/pr_jobs.groovy
+      -->
       <id>run-code-coverage</id>
       <properties>
         <jacoco.excludes>*Lexer</jacoco.excludes>
-        <jacoco.exec.file>${project.build.directory}/jacoco.exec</jacoco.exec.file>
         <jacoco.agent.line>-javaagent:${settings.localRepository}/org/jacoco/org.jacoco.agent/${version.jacoco.plugin}/org.jacoco.agent-${version.jacoco.plugin}-runtime.jar=destfile=${jacoco.exec.file},append=true,excludes=${jacoco.excludes}</jacoco.agent.line>
```
yurloc: The agent line can be generated using `prepare-agent` goal.

.travis.yml
yurloc: I would say this can't work because under `sonarcloud-analysis` profile, Jacoco report execution and sonarcloud scanner are bound to `generate-sources` phase. Let's talk about this tomorrow.

rsynek: Nice catch! Yes, there has to be a separate maven run as we have for VRP:
https://github.com/kiegroup/optaweb-vehicle-routing/blob/master/.travis.yml#L24

pom.xml
rsynek: I am not in favour of this change; I prefer the `jacoco.agent.line` value being visible to having it set by the jacoco maven plugin under the hood.

yurloc: The value is logged by the plugin during build so it is visible, isn't it?

> [INFO] --- jacoco-maven-plugin:0.8.5:prepare-agent (jacoco-prepare-agent) @ employee-rostering-backend ---
[INFO] jacoco.agent.argLine set to -javaagent:/home/jlocker/.m2/repository/org/jacoco/org.jacoco.agent/0.8.5/org.jacoco.agent-0.8.5-runtime.jar=destfile=/home/jlocker/src/github.com/kiegroup/optaweb-employee-rostering/employee-rostering-backend/target/jacoco.exec,excludes=*Lexer


 I thought that `prepare-agent` is the intended way to populate the property unless you have some corner case scenario or need to override low-level defaults that cannot be set using the plugin configuration.

I also think it is safer to use `prepare-agent` because if the default content of the agent argLine changes in a newer version and we don't reflect the change in this place, the plugin might stop working correctly, whereas it would most likely keep working if the property was populated using `prepare-agent`.

Please correct me if I'm missing something. Is there a reason to have the value hard-coded here other than visibility? Is the format documented?

rsynek: There are no other reasons than visibility that I am aware of.

The jacoco agent argLine format is documented here:
https://www.eclemma.org/jacoco/trunk/doc/agent.html

> whereas it would most likely keep working if the property was populated using prepare-agent.
I agree this point might overweight the visibility. Let me take a look.

rsynek: Changed to jacoco:prepare-agent

pom.xml
```diff
@@ -143,105 +146,26 @@
...
   <profiles>
     <profile>
+      <!--
+        TODO: For moving the coverage measurement from Travis to Jenkins, add the repository name to
+        SONARCLOUD_ENABLED_REPOSITORIES in
+        https://github.com/kiegroup/kie-jenkins-scripts/blob/master/job-dsls/jobs/pr_jobs.groovy
+      -->
       <id>run-code-coverage</id>
-      <properties>
-        <jacoco.excludes>*Lexer</jacoco.excludes>
```
yurloc: @rsynek We forgot to move this to the `prepare-agent` configuration. I can see this in Travis console now:
```
java.lang.instrument.IllegalClassFormatException: Error while instrumenting org/drools/compiler/lang/DRL6Lexer.
```
It's a mystery to me that I can't see this error in vehicle routing despite the fact that they have the same Drools artifacts in their dependency trees.

rsynek: Good catch, the fix is here:
https://github.com/kiegroup/optaweb-employee-rostering/pull/371

I checked optaweb-vehicle-routing too and as you said, no sign of this error message. The coverage reported to SonarCloud is over 80%, which looks ok, I can see in the log that the `prepare-agent` goal of jacoco maven plugin was triggered. Yet, no exception about the *Lexer class.

PR #373, opened 2020-01-28
PLANNER-1811: Add optaweb prefix to distribution ZIP

yurloc:


employee-rostering-distribution/pom.xml
```diff
@@ -15,6 +15,7 @@
   <name>Employee Rostering as a Service Distribution</name>
 
   <build>
+    <finalName>optaweb-${project.artifactId}-${project.version}</finalName>
```
mareknovotny: why don't you rename the directory and name of the maven module instead of this finalName? It is always better to have in sync the files/folders on filesystem with maven names. This creates a confusion for first view.

yurloc: I've just tested it locally and the change has some consequences that I didn't expect. The build now produces `optaweb-employee-rostering-distribution-7.33.0-SNAPSHOT.zip` in `target/` (expected) but
```
$ ls -1 ~/.m2/repository/org/optaweb/employeerostering/employee-rostering-distribution/7.33.0-SNAPSHOT/
employee-rostering-distribution-7.33.0-SNAPSHOT.pom
employee-rostering-distribution-7.33.0-SNAPSHOT.zip
maven-metadata-local.xml
_remote.repositories
```
so the `artifactId` remains unchanged. That makes sense. What I didn't realize is that the [upload scripts](https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1174/files) take artifacts from `~/.m2/repository` which is unaffected by this change.

So we have two options:

1. Change `artifactId`. Add `optaweb` prefix to artifact ID of this module and all other modules for consistency. And probably also rename the directories.
2. Rename the artifact in the upload script when copying from deploy dir to upload dir or when uploading.

PR #374, opened 2020-01-28
Exclude upstream-repos from the distribution

yurloc:


yurloc:
Jenkins build green, API rate limit reached.

PR #376, opened 2020-01-29
Replace module-relative paths in sonar inclusion with project-relative

yurloc:
Removes following warning:

Specifying module-relative paths at project level in the property
'sonar.exclusions' is deprecated. To continue matching files like
'employee-rostering-frontend/src/store/alert/alert.test.ts', update
this property so that patterns refer to project-relative paths.

### DIFF
--- a/optaweb-employee-rostering-frontend/pom.xml
+++ b/optaweb-employee-rostering-frontend/pom.xml
@@ -34,7 +34,6 @@
     <node.version>v11.6.0</node.version>
     <npm.version>6.9.0</npm.version>
     <sonar.sources>src</sonar.sources>
-    <sonar.test.inclusions>**/*test.ts,**/*test.tsx</sonar.test.inclusions>
     <sonar.javascript.lcov.reportPaths>coverage/lcov.info</sonar.javascript.lcov.reportPaths>
   </properties>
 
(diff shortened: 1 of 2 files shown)


yurloc:
Do not merge. Needs to go after #377.

pom.xml
```diff
@@ -62,6 +62,10 @@
     <jacoco.agent.argLine/>
     <sonar.projectKey>optaweb-employee-rostering</sonar.projectKey>
     <sonar.moduleKey>${project.artifactId}</sonar.moduleKey>
+    <sonar.test.inclusions>
+      optaweb-employee-rostering-frontend/**/*test.ts,
+      optaweb-employee-rostering-frontend/**/*test.tsx
+    </sonar.test.inclusions>
```
yurloc: We can't move it to the parent as that would affect other modules. This would result in excluding Java test files from scanning (note that's something different from coverage analysis).

optaweb-employee-rostering-frontend/pom.xml
```diff
@@ -34,7 +34,6 @@
     <node.version>v11.6.0</node.version>
     <npm.version>6.9.0</npm.version>
     <sonar.sources>src</sonar.sources>
-    <sonar.test.inclusions>**/*test.ts,**/*test.tsx</sonar.test.inclusions>
```
yurloc: And we can't simply use `optaweb-employee-rostering-frontend/**/*test.ts` here because that doesn't work - results in counting tests as sources and measuring their coverage.

yurloc:
This doesn't work. I couldn't find a way to get rid of the warning without breaking some part of the analysis.

I need to file a ticket in https://jira.sonarsource.com/secure/BrowseProjects.jspa?selectedCategory=all&selectedProjectType=software&s=view_projects.

PR #379, opened 2020-02-03
PLANNER-1797: Move PostgreSQL to a production profile

yurloc:


optaweb-employee-rostering-backend/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM adoptopenjdk/openjdk8:ubi-minimal-jre
 COPY target/*-exec.jar /opt/app/optaweb-employee-rostering.jar
 WORKDIR /opt/app
-CMD ["java", "-jar", "optaweb-employee-rostering.jar"]
+CMD ["java", "-jar", "optaweb-employee-rostering.jar", "--spring.profiles.active=production"]
```
LoneRifle: From what I understand of this, backend will _always_ run with the production profile and it can no longer be determined by env var. Is this intended?

yurloc: Good catch, you're right. I was aware of this consequence. But until you made this comment, I didn't realize that it's very easy to make this configurable using an environment variable like this:
```
ENV SPRING_PROFILES_ACTIVE production
```
instead of using the command line argument.

Thanks for pointing this out. I'll update the PR.

PR #380, opened 2020-02-03
RHDM-1171: Replace Patternfly's Select with react-select so it can go out of modals

Christopher-Chianelli:
Also Updated dependency versions.


LoneRifle:
The sole integration test is failing, probably because of the refactor:
```
       Spec                                              Tests  Passing  Failing  Pending  Skipped  
  ┌───────────────────────────────────────────────────────────────────────────────────────���────────┐
  │ ✖  newTenantBasicWorkflow.js                02:30        1        -        1        -        - │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘
    ✖  1 of 1 failed (100%)                     02:30        1        -        1        -        -  
```

Thanks for looking into this btw - I was about to update the frontend deps and submit a PR upstream

Christopher-Chianelli:
It indeed from the refactor; issue is for some reason sometimes clicking the select option works in the IT, other times it does not. I was unable to reproduce the error when redoing the steps the IT does manually. This either mean my selector is ambiguous and it click the wrong thing, or there is a missing step (that only need to be performed on some selects)

LoneRifle:
I will send in a separate PR that embodies only the dependency updates, so that we can pick up the changes downstream. Hope that's okay with you

optaweb-employee-rostering-frontend/.eslintrc.json
```diff
@@ -1,7 +1,7 @@
 {
   "parser": "@typescript-eslint/parser",
   "parserOptions": {
-    "project": "./tsconfig.json"
+    "project": ["./tsconfig.json", "cypress/tsconfig.json"]
```
yurloc: Why this? Do we have any TypeScript code in `cypress` directory?

optaweb-employee-rostering-frontend/.eslintrc.json
```diff
@@ -61,6 +62,7 @@
         " "
       ]
     ],
+    "import/extensions": "off", // Suddenly, the new norm for imports is to include the file extension
```
yurloc: Please don't turn off the rule, rather override its configuration so that it matches the current code style. The rule should remain active to ensure consistent import style.

This works for me:
```json
"import/extensions": ["error", "never", {"css":  "always"}],
```

yurloc:
@Christopher-Chianelli what do you think about handling the version upgrades separately. Can we work on #384 separately or would you rather finish all of that yourself?

optaweb-employee-rostering-frontend/.eslintrc.json
Christopher-Chianelli: Because the cypress directory is passed to eslint, and eslint fails if it finds a file not in a project. (It just javascript files, but I need a project so eslint know the files are in the project).

Christopher-Chianelli:
@yurloc Upgrading the dependencies to the latest version required non-trivial changes to the code. Additionally, both `TypeaheadSelect` and `MultiTypeaheadSelect` require changes without this refactor (namely, an event now can be undefined). The work on this refactor is almost done (as in it just making the cypress tests pass again), so I don't think handling version upgrades separately is going to be useful.

PR #390, opened 2020-02-18
Allow copying of shifts

LoneRifle:
Screenshot 2020-02-18 at 11 17 46 AM

It is somewhat common to have multiple people scheduled within the same
time period of a roster, so introduce a button in ShiftPopupHeader (and
template equivalent) that allows the quick copying of shifts

- Popup headers: Create a button that invokes an `onCopy()` callback
  - make this optional for ShiftPopupHeader usage in
    AvailabilityRosterPage as it makes little sense in the user-level
    view to copy a slot with either the same user or nobody assigned
    (the latter because the copied slot then disappears from view)
- Implement callbacks for RotationPage and ShiftRosterPage


ge0ffrey:
+1 to include this feature upstream. Any user in our userbase will want this. I 'll leave the review to @Christopher-Chianelli and/or @yurloc

PR #393, opened 2020-02-19
Add Mandarin translation for ScoreDisplay

LoneRifle:


Christopher-Chianelli:
@liangyuanruo that is already done in master:
https://github.com/kiegroup/optaweb-employee-rostering/blob/9a561b7db5510774ad2ddf777fab3a34f40abba5/optaweb-employee-rostering-frontend/src/ui/components/ScoreDisplay.tsx#L25-L41

PR #396, opened 2020-02-19
WeekPicker -> IntervalPicker

LoneRifle:
image

Give users the ability to look at their availability or shift rosters by toggling between day, week and month views

- Rename WeekPicker to IntervalPicker, and update all references
- Give IntervalPicker the ability to toggle between day/week/month views
- Rostering pages
  - Use the start/end dates given by IntervalPicker instead of recomputing from scratch
  - Hold the desired interval as a state field, injecting this into IntervalPicker
  - Let IntervalPicker change the page interval through a callback, in turn influencing the calendar view in Schedule

### DIFF
--- a/optaweb-employee-rostering-frontend/src/ui/components/IntervalPicker.css
+++ b/optaweb-employee-rostering-frontend/src/ui/components/IntervalPicker.css
@@ -13,11 +13,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
- .week-picker-container {
+ .interval-picker-container {
     background-color: var(--pf-global--BackgroundColor--100);
  }
 
-.week-picker {
+.interval-picker {
     z-index: 100;
     background-color: var(--pf-global--BackgroundColor--200);
 }
--- a/optaweb-employee-rostering-frontend/src/ui/components/IntervalPicker.css
+++ b/optaweb-employee-rostering-frontend/src/ui/components/IntervalPicker.css
@@ -21,3 +21,7 @@
     z-index: 100;
     background-color: var(--pf-global--BackgroundColor--200);
 }
+
+.interval-picker-toggle {
+    display: inline;
+}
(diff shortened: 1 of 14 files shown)


LoneRifle:
Hello, what do you think of this? Is this something you would like to take forward?

Christopher-Chianelli:
We'll need to check with our UX team for massive changes to the UI. From what I seen from manually running the PR, this basically add `react-big-calendar`'s interval selector to the Week Picker, allowing the Week Picker to change what interval the `react-big-calendar` display. A couple of issues with this approach:

- It becomes hard to see all the Shifts in the day if a day has many Shifts
- Cannot tell the difference between a day shift and a night shift

Before UX work can continue, we need to spec out what issue we are trying to address (for instance, are we trying to show the days an Employee is working in a month). This can lead to several different designs (for instance, having two separate screens, one for managers and one for employees, which default to the view that is most helpful to them). I don't see us merging the PR as is, and we might come up with a completely different design depending on the issue we are solving.

LoneRifle:
Fine by me, I look forward to seeing what you all come up with!

LoneRifle:
I am revisiting the PR now - if you are indeed considering making a UX design change to allow selection of intervals, could we at least merge the first of the commits (ie, af4cebb) in this PR? I would be happy to submit a separate PR just merging that commit, and feel that the renaming of WeekPicker would place this codebase in a slightly better position in that respect

Christopher-Chianelli:
@LoneRifle That would lock us into a "One Component for all Views" approach, which I am not sure is the correct way to go about it (also, looking at that first commit, the extended description seems to lie; I don't see anything that would accomplish "Roster pages: use the picker start/end dates, do not recompute" (I see it is still recomputing the dates from the URL)).

LoneRifle:
Thanks for replying! In that commit, the roster page validates the interval dates from the URL by ensuring they are set to "beginning of day at start date" to "end of day at end date with fallbacks" from the URL props, as you have pointed out. That is a significant change from being solely reliant on a single date and calculating the interval from there, but still retains some semblance of the original flow, before more invasive changes are made in subsequent commits.

I contend that your concerns of having one component for all views would be limited to the picker itself. This should make sense: an interval is just a pair of dates, so we should not need to build distinct components for each kind of interval (or should it?). 

It _might_ actually be helpful given that a user might want an interval to be user-defined, though admittedly that may not happen often enough.

PR #402, opened 2020-03-26
PLANNER-1924: Refactor the domain for COVID-19

Christopher-Chianelli:


optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/tenant/RosterParametrization.java
```diff
@@ -34,6 +34,14 @@
...
     private Integer rotationEmployeeMatchWeight = 500;
     @NotNull
+    private Integer lowRiskEmployeeInCovidWardMatchWeight = 1; // Soft
+    @NotNull
+    private Integer moderateRiskEmployeeInCovidWardMatchWeight = 5; // Soft
+    @NotNull
+    private Integer highRiskEmployeeInCovidWardMatchWeight = 10; // Soft
+    @NotNull
+    private Integer extremeRiskEmployeeInCovidWardMatchWeight = 1; // Hard
+    @NotNull
```
triceo: I have noticed that there is a `...View` version of this class, which is more or less identical to this one.
What is the purpose of that one, and do I need to duplicate the code there?

triceo:
I have added the new constraints for PLANNER-1924 to this PR. Also, I left one comment.

triceo:
I see that some tests are now failing. I'm assuming that it is because of my new constraints, but I don't know the code base well enough yet to be able to tell for certain.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/roster/RosterGenerator.java
```diff
@@ -578,9 +583,10 @@ public RosterState createRosterState(GeneratorType generatorType, Tenant tenant,
             String name = employeeNameGenerator.generateNextValue();
             HashSet<Skill> skillProficiencySet = new HashSet<>(extractRandomSubList(generalSkillList,
                                                                                     0.1, 0.3, 0.5, 0.7, 0.9, 1.0));
+            CovidRiskType covidRisk = extractRandomElement(Arrays.asList(CovidRiskType.values()));
```
triceo: I believe that the distribution of covid risk type should follow a bell curve. As best as I can tell, the current distribution is uniform.

triceo:
Some tests are now failing, because we have newly introduced hard constraints:

    Explanation of score (-63hard/-520medium/-360soft):
        Constraint match totals:
            -63hard: constraint (Extreme-risk employee assigned to a COVID ward) has 63 matches:
                -1hard: justifications ([Chassis, Chad Green, Chassis 2020-03-23T14:00+01:00-2020-03-23T22:00+01:00, [RosterParametrization-17659]])
                -1hard: justifications ([Chassis, Beth Fox, Chassis 2020-03-23T22:00+01:00-2020-03-24T06:00+01:00, [RosterParametrization-17659]])

Specifically, the following assumption in the tests no longer holds:

    // Due to overconstrained planning, the score is always feasible

I wonder what to do about that. 
Can we generate a solution that is easier to make feasible?
Or should we simply stop asserting feasibility?

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/tenant/RosterParametrization.java
Christopher-Chianelli: @cuijulian Why do we have a separate view class for `RosterParametrization`? I don't see anything that need to be serialized differently for the client.

Christopher-Chianelli: (we don't call getRosterParametrization/updateRosterParametrization in the client yet, so you do not need to update it unless we want to add a config screen.

cuijulian: Pretty much all domain entities have a `...View` version. This was for Spring security reasons iirc, it wasn't good practice to pass entities directly to RestControllers, so the `...View` objects are passed instead, which then get converted to the non-view entities.

Christopher-Chianelli: Why not create a new Entity and update the fields manually?
(i.e:
```
       oldRosterParametrization.setDesiredTimeSlotWeight(updatedRosterParametrization.getDesiredTimeSlotWeight());
        oldRosterParametrization.setRotationEmployeeMatchWeight(
                updatedRosterParametrization.getRotationEmployeeMatchWeight());
        oldRosterParametrization.setUndesiredTimeSlotWeight(
                updatedRosterParametrization.getUndesiredTimeSlotWeight());
        oldRosterParametrization.setWeekStartDay(updatedRosterParametrization.getWeekStartDay());
```

Christopher-Chianelli: (we are not persisting an entity directly, but updating the fields of an entity)

cuijulian: What would be passed into the RestController in that case? Wouldn't you still need to pass in the `updatedRosterParametrization` (which is still an `@Entity`) in the PUT request?

Christopher-Chianelli: Code will look like this:
```
createEntity(Entity newEntity) {
    Entity entity = new Entity();
    // verify legal
    entity.setX(newEntity.getX());
    // ...
    persist(entity);
}

updateEntity(Entity updatedEntity) {
    Entity oldEntity = getOldEntity();
    // verify legal
    oldEntity.setX(updatedEntity.getX());
    // ...
    persist(oldEntity)
}
```

PR #406, opened 2020-03-28
PLANNER-1926 COVID Staff Migration Constraints

triceo:


optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/shift/Shift.java
```diff
@@ -91,6 +91,14 @@ public String toString() {
         return spot + " " + startDateTime + "-" + endDateTime;
     }
 
+    public boolean follows(Shift other) {
+        return startDateTime.equals(other.endDateTime) || startDateTime.isAfter(other.endDateTime);
```
Christopher-Chianelli: Can be simplified to `!startDateTime.isBefore(other.endDateTime)`

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/shift/Shift.java
```diff
@@ -91,6 +91,14 @@ public String toString() {
...
+    }
+
+    public boolean precedes(Shift other) {
+        return endDateTime.equals(other.startDateTime) || endDateTime.isBefore(other.startDateTime);
```
Christopher-Chianelli: Can be simplified to `!endDateTime.isAfter(other.startDateTime)`

PR #409, opened 2020-03-30
PLANNER-1925 Final set of COVID constraints

triceo:


optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/shift/Shift.java
```diff
@@ -46,6 +51,8 @@
     private OffsetDateTime startDateTime;
     @NotNull
     private OffsetDateTime endDateTime;
+    @Transient
+    private Long lengthInHours = null;
```
Christopher-Chianelli: I am guessing you are caching `startDateTime.until(endDateTime, ChronoUnit.HOURS);`? Is that operation expensive enough to justify this?

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/shift/Shift.java
```diff
@@ -99,6 +106,52 @@ public boolean precedes(Shift other) {
...
+        }
+        return lengthInHours;
+    }
+
+    public static long calculateLoad(Collection<Integer> hourlyCounts) {
+        long sumSquares = 0;
+        for (int hourlyCount: hourlyCounts) {
+            sumSquares += hourlyCount * hourlyCount;
+        }
+        long squareRootOfSums = Math.round(Math.sqrt(sumSquares) * 1000);
```
Christopher-Chianelli: Should the 1000 be here, or in the score constraint file?

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/shift/Shift.java
triceo: I'll remove it. I haven't done enough benchmarking to show that this is a problem.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/shift/Shift.java
triceo: This is not so much a constraint weight as a way to avoid losing precision. In this context, 2.123 is something far different from 2.0.

Once I convert this code to CS-D, we may be able to make this a `BIgDecimal` easily, and , keep it in its original decimal form and only use constraint configuration to weight it properly.

triceo:
I agree about a load balancing constraint collector - there is even a JIRA for it already.

PR #410, opened 2020-03-30
Add id to Switches since they do not work without them now, make popover appear for all shifts regardless of position.

### LINKED ISSUE #408: Unable to pin
Hello

I'm unable to enable a pin for a a shift, either a new or existing shift.

# Steps to reproduce

1. Using version (`git describe`) 0.1.0-1091-g04844f35
1. `mvn clean install -DskipTests -DskipITs`
1. `java -jar optaweb-employee-rostering-standalone/target/optaweb-employee-rostering-standalone-*-exec.jar`
1. Open brower, ether FF or Chrome, on Ubuntu 19, go to, for example http://localhost:8080/55417/shift?spot=Airbag+line+15
1. Click like a maniac on the `Is Pinned` toggle switch.

# Expected result

Expect I can change the pin state. And am hoping this will also be visually represented.

It is the same issue on either an existing, or when creating a new, shift.

Christopher-Chianelli:
Fixes #408


Christopher-Chianelli:
Jenkins passed, merging

PR #412, opened 2020-03-30
(DO NOT MERGE) PLANNER-1936 Constraint Streams for Optaweb ER

triceo:
The performance of CS in this case is worse than that of the DRL. I don't have exact numbers, but quick glance suggests a 50 % drop.
I'll be asking QE if that is acceptable.


optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/solver/EmployeeRosteringConstraintProvider.java
```diff
@@ -0,0 +1,332 @@
...
+                undesiredEmployeeTimeSlot(constraintFactory),
+                desiredEmployeeTimeSlot(constraintFactory),
+                employeeNotRotationEmployee(constraintFactory)
+        };
+    }
+
+    private Constraint riskEmployeeAssignedToCovidWard(ConstraintFactory constraintFactory, CovidRiskType riskType,
+            String constraintName) {
+        return constraintFactory.from(Shift.class)
+                .filter(shift -> shift.getEmployee() != null)
```
Christopher-Chianelli: Doesn't `from` already do this (v.s. `fromUnfilitered`)?

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/solver/EmployeeRosteringConstraintProvider.java
```diff
@@ -0,0 +1,332 @@
...
+    Constraint maximizeInoculatedHours(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Shift.class)
+                .filter(shift -> shift.getEmployee() != null)
+                .filter(shift -> shift.getEmployee().getCovidRiskType() == INOCULATED)
+                .rewardConfigurableLong(CONSTRAINT_COVID_MAXIMIZE_INOCULATED_HOURS, Shift::getLengthInHours);
+    }
+
+    Constraint migrationBetweenCovidAndNonCovidWards(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Shift.class)
+                .filter(shift -> shift.getEmployee() != null)
```
Christopher-Chianelli: I wonder if this can be improved using a joiner:

```
constraintFactory.from(Shift.class)
    .factFilter(s -> s.getSpot().isCovidWard) // It be nice if we have a separate filter for facts that do not change,
     // so that this filter does not get rerun on a Shift when it changes
    .join(Shift.class, Joiners.notEquals(s -> s.getSpot().isCovidWard()), Joiners.equals(Shift::getEmployee), Joiners.greaterThanOrEqual(Shift::getEndDateTime, Shift::getStartDateTime))
    .ifNotExists(Shift.class, /* same code */)
    .penalizeConfigurable(...)
```

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/solver/EmployeeRosteringConstraintProvider.java
```diff
@@ -0,0 +1,332 @@
...
+                equal(Shift::getEmployee),
+                equal(shift -> shift.getStartDateTime().toLocalDate()))
+                .filter((shift1, shift2) -> shift1.getEmployee() != null)
+                .penalizeConfigurable(CONSTRAINT_AT_MOST_ONE_SHIFT_ASSIGNMENT_PER_DAY_PER_EMPLOYEE);
+    }
+
+    Constraint noTwoShiftsWithin10HoursFromEachOther(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Shift.class)
+                .filter(shift -> shift.getEmployee() != null)
+                .join(Shift.class, equal(Shift::getEmployee))
```
Christopher-Chianelli: Add another Joiner for lessThanOrEqual to?
`Joiners.lessThanOrEqualTo(Shift::getEndTime, Shift:getStartTime)`?

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/solver/EmployeeRosteringConstraintProvider.java
triceo: Yes and no.

`employee` on `Shift` is nullable, and therefore even initialized `Shift`s may have the value `null`. This is correct.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/solver/EmployeeRosteringConstraintProvider.java
triceo: It would be possible, only `notEqual()` joiner currently does not exist. I have a JIRA for that in current sprint, but it will likely have to take a back seat due to this COVID work.

Christopher-Chianelli: Do booleans implement `Comparable` (i.e. is `false < true` or vice-versa)?

triceo:
@cuijulian I remove the DRL as I consider it a maintenance burden. Effectively it's a fork of the scoring function with all the downsides of DRL, only maintained for `FULL_ASSERT`. I'd rather introduce a `ConstraintVerifier`. That said, since there is none at the moment, I do see your point. I will have to think about it.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/solver/EmployeeRosteringConstraintProvider.java
triceo: They do, but I don't consider it very intuitive to rely on that behavior.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/solver/EmployeeRosteringConstraintProvider.java
```diff
@@ -165,29 +166,35 @@ Constraint migrationBetweenCovidAndNonCovidWards(ConstraintFactory constraintFac
...
 
     Constraint nonCovidShiftStartedLessThan8HoursAfterFinishingCovidShift(ConstraintFactory constraintFactory) {
         return constraintFactory.from(Shift.class)
                 .filter(shift -> shift.getEmployee() != null)
-                .filter(shift -> !shift.getSpot().isCovidWard())
-                .join(Shift.class, equal(Shift::getEmployee))
-                .filter((nonCovidShift, otherShift) -> otherShift.getSpot().isCovidWard())
-                .filter((nonCovidShift, covidShift) -> covidShift.precedes(nonCovidShift))
-                .filter((nonCovidShift, covidShift) ->
+                .filter(shift -> shift.getSpot().isCovidWard())
```
Christopher-Chianelli: I think you need to negate the `isCovidWard` filters for this constraint

yurloc:
@triceo Beware of https://issues.redhat.com/browse/PLANNER-1909. I've just check with master and found out that employee rostering is affected too.

Takeway: when comparing score calculation speed before and after CS, always use `java -jar` with backend or standalone JARs.

triceo:
Replaced by #434.

PR #417, opened 2020-04-01
Add requiredSkillSet to Shift and ShiftTemplate, add originalEmployee to Shift

Christopher-Chianelli:
Note: the Modal for Shift and Shift Templates do not include the requiredSkillSet and originalEmployee yet.


optaweb-employee-rostering-backend/src/main/resources/org/optaweb/employeerostering/service/solver/employeeRosteringScoreRules.drl
```diff
@@ -279,6 +279,13 @@ end
 // Soft constraints
 // ############################################################################
 
+rule "Change of Employee for Published Shift"
+    when
+        Shift(moved == true)
+    then
+        scoreHolder.penalize(kcontext);
+end
+
```
triceo, line 288: I'm missing a `ConstraintConfiguration` for this one.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/shift/Shift.java
```diff
@@ -107,17 +137,26 @@ public long getLengthInHours() {
         return startDateTime.until(endDateTime, ChronoUnit.HOURS);
     }
 
+    public boolean isMoved() {
+        return originalEmployee != null && originalEmployee != employee;
+    }
+
+    public boolean hasRequiredSkills() {
+        return employee.getSkillProficiencySet().containsAll(spot.getRequiredSkillSet()) &&
+                employee.getSkillProficiencySet().containsAll(requiredSkillSet);
```
triceo, line 146: This smells of inefficiency. How about:
    
    Set<...> skillProficiencySet = employe.getSkillProficiencySet();
    return Stream.concat(spot.getRequiredSkillSet().stream(), requiredSkillSet.stream())
        .distinct() // Test every skill at most once.
        .allMatch(skillProficiencySet::contains);

Assuming that there is a significant overlap between the two skill sets, this will be faster. Otherwise, not so much. Maybe not worth the hassle.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/roster/RosterGenerator.java
```diff
@@ -493,20 +503,43 @@ public RosterState createRosterState(GeneratorType generatorType, Tenant tenant,
...
+                        entityManager.persist(shiftTemplate);
+                        shiftTemplateList.add(shiftTemplate);
+
+                        rotationEmployeeIndex = generatorType.rotationEmployeeIndexCalculator
+                                .apply(startDayOffset, timeslotRangesIndex).get(1);
+                        rotationEmployee = rotationEmployeeIndex >= nurseRotationEmployeeList.size() ? null :
+                                nurseRotationEmployeeList.get(rotationEmployeeIndex);
+                        shiftTemplate = new ShiftTemplate(tenantId, spot, startDayOffset, startTime,
+                                                          endDayOffset, endTime, rotationEmployee,
+                                                          new HashSet<>(Arrays.asList(skillList.get(2))));
```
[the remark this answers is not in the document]
Christopher-Chianelli: I disagree using varargs, because then you cannot pass a Collection directly in. I think  a good compromise would be making it take a Collection, and create a new HashSet in the constructor. WDYT?

optaweb-employee-rostering-backend/src/main/resources/org/optaweb/employeerostering/service/solver/employeeRosteringScoreRules.drl
Christopher-Chianelli: Technically not implemented yet, will remove until we implement non-distruptive planning (so not in scope)

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/shift/Shift.java
Christopher-Chianelli: There should be 0 overlap between the spot skill set and the shift skill set (shift skill set is additional skills required for the shift).

PR #423, opened 2020-04-06
Add separate page for published shifts vs draft shifts, add non-disruptive planning, add indictment summary

Christopher-Chianelli:

### DIFF
--- a/optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/roster/Roster.java
+++ b/optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/roster/Roster.java
@@ -16,6 +16,7 @@
 
 package org.optaweb.employeerostering.domain.roster;
 
+import java.time.OffsetDateTime;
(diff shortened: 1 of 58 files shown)


triceo:
I suggest we make the following changes:

- "Adjust/Draft" buttons move to the right end of the menu.
- "Adjust" is changed to "Adjustments" and is the last item in the menu.
- "Draft" is changed to "Scheduling".

It's a bit confusing that, in the "Adjustments" view, I still need to first create a fully initialized solution. Only then, when I solve again, does the replanning work properly. How to solve that? Don't know. Maybe don't allow replanning buttons until we've had a feasible solution?

Also, playing a bit more with the UI, I suggest the following improvements:

- From Wards, I can directly click through to a schedule (not draft) for the particular ward.
- From Employees, I can directly click through to see Availability for the particular employee.
- Next to the hard/medium/soft display, we should show a warning if we have broken hard/medium constraints. Something like "Schedule may not be feasible." Possibly with a tooltip of the hard/medium indictments..

PR #428, opened 2020-04-16
Fix Sonar code coverage analysis

yurloc:
`mvn generate-resources -Psonarcloud-analysis` runs an Ant task that
aggregates reports from individual modules but this project doesn't seem
to be configured correctly for code coverage aggregation.

Adding the report goal fixes code coverage analysis for individual
modules because it generates a JaCoCo XML report in each module that can
be understood by Sonar scanner.

Radek will fix report aggregation later.


rsynek:
@yurloc nice catch, thank you! As we discussed, the cross-module coverage will be fixed later, this fix is very good on its own.

PR #432, opened 2020-04-27
Restore cross-module coverage measurement

rsynek:
The reporting is done in the sonarcloud-analysis profile in the kie-parent based on a single file every module appends coverage data to.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -181,17 +181,14 @@
                   <goal>prepare-agent</goal>
                 </goals>
                 <configuration>
+                  <!-- Append coverage data so a single file defined in the kie-parent -->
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -181,18 +181,15 @@
                   <goal>prepare-agent</goal>
                 </goals>
                 <configuration>
+                  <!-- Append coverage data so a single file defined in the kie-parent -->
```
yurloc, line 184:
```suggestion
                  <!-- Append coverage data to a single file defined in the kie-parent -->
```

rsynek: Thanks!

PR #434, opened 2020-04-29
PLANNER-1936 ConstraintProvider for Employee Rostering

triceo:
Provides Constraint Streams as an alternative scoring function.
Does not replace DRL.

### DIFF
--- a/optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/solver/EmployeeRosteringConstraintProvider.java
+++ b/optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/solver/EmployeeRosteringConstraintProvider.java
@@ -88,7 +88,9 @@ public final class EmployeeRosteringConstraintProvider implements ConstraintProv
                 .join(Shift.class,
                         equal(EmployeeAvailability::getEmployee, Shift::getEmployee),
                         lessThanOrEqual(EmployeeAvailability::getStartDateTime, Shift::getEndDateTime),
-                        greaterThanOrEqual(EmployeeAvailability::getEndDateTime, Shift::getStartDateTime));
+                        greaterThanOrEqual(EmployeeAvailability::getEndDateTime, Shift::getStartDateTime))
+                .filter((availability, shift) ->
+                        !Objects.equals(availability.getEndDateTime(), shift.getStartDateTime()));
     }
 
     private static UniConstraintStream<Shift> getAssignedShiftConstraintStream(ConstraintFactory constraintFactory) {
@@ -219,12 +221,11 @@ public final class EmployeeRosteringConstraintProvider implements ConstraintProv
         return getAssignedShiftConstraintStream(constraintFactory)
                 .join(Shift.class,
                         equal(Shift::getEmployee),
-                        lessThanOrEqual(Shift::getStartDateTime),
-                        lessThanOrEqual(Shift::getStartDateTime, Shift::getEndDateTime),
-                        greaterThanOrEqual(Shift::getEndDateTime, Shift::getStartDateTime))
-                .filter((s1, s2) -> !Objects.equals(s1, s2))
-                .filter((s1, s2) -> !Objects.equals(s1.getEndDateTime(), s2.getStartDateTime()))
-                .penalizeConfigurableLong(CONSTRAINT_NO_OVERLAPPING_SHIFTS, (s1, s2) -> s2.getLengthInMinutes());
+                        greaterThanOrEqual(Shift::getStartDateTime),
+                        lessThanOrEqual(Shift::getEndDateTime, Shift::getStartDateTime))
+                .filter((shift, startDuringFirstShift) -> !Objects.equals(shift, startDuringFirstShift))
+                .penalizeConfigurableLong(CONSTRAINT_NO_OVERLAPPING_SHIFTS,
+                        (shift, startsDuringFirstShift) -> startsDuringFirstShift.getLengthInMinutes());
     }
 
     Constraint noMoreThanTwoConsecutiveShifts(ConstraintFactory constraintFactory) {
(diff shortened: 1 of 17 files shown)


optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/solver/EmployeeRosteringConstraintProvider.java
```diff
@@ -0,0 +1,341 @@
...
+                .filter(employeeAvailability -> employeeAvailability.getState() == employeeAvailabilityState)
+                .join(Shift.class, equal(EmployeeAvailability::getEmployee, Shift::getEmployee))
+                .filter((employeeAvailability, shift) -> DateTimeUtils.doTimeslotsIntersect(
+                        employeeAvailability.getStartDateTime(), employeeAvailability.getEndDateTime(),
+                        shift.getStartDateTime(), shift.getEndDateTime()));
+    }
+
+    private static UniConstraintStream<Shift> getAssignedShiftConstraintStream(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Shift.class)
+                .filter(shift -> shift.getEmployee() != null);
```
ge0ffrey, line 96: from does this automaticaly

triceo: Does not apply for nillable variables.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/solver/EmployeeRosteringConstraintProvider.java
```diff
@@ -0,0 +1,341 @@
...
+import static org.optaweb.employeerostering.domain.tenant.RosterConstraintConfiguration.CONSTRAINT_YEARLY_MINUTES_MUST_NOT_EXCEED_CONTRACT_MAXIMUM;
+
+public final class EmployeeRosteringConstraintProvider implements ConstraintProvider {
+
+    private static BiConstraintStream<EmployeeAvailability, Shift> getAvailabilityConstraintStream(
+            ConstraintFactory constraintFactory, EmployeeAvailabilityState employeeAvailabilityState) {
+        return constraintFactory.from(EmployeeAvailability.class)
+                .filter(employeeAvailability -> employeeAvailability.getState() == employeeAvailabilityState)
+                .join(Shift.class, equal(EmployeeAvailability::getEmployee, Shift::getEmployee))
+                .filter((employeeAvailability, shift) -> DateTimeUtils.doTimeslotsIntersect(
```
Christopher-Chianelli, line 89: Can you replace intersects with Joiners?

(in particular, `StartA < EndB) and (EndA > StartB)`, which would be something like this:
```
Joiners.lessThan(EmployeeAvailability::getStartDateTime, Shift::getEndDateTime), Joiners.greaterThan(EmployeeAvailability::getEndDateTime, Shift::getStartDateTime)
```
which will probably be better performance wise

triceo:
Ready to merge. I have done several things now:

- Rebased onto latest changes upstream.
- Unified CS-D and DRL justifications, they should be identical now.
- Improved performance of both CS-D and DRL. Both are faster now, but CS-D should be a bit closer to DRL than it was. (Although it would still lose by a lot to DRL, and should not be enabled by default.)

Also, when I was working on this, I noticed how non-linear our Git history is.
nonlinear-history

Can we please avoid merge commits? Rebasing or squashing works just as well, without producing this barely decipherable ugliness.

yurloc:
@triceo In my opinion, merge commits are good for long feature branches. They make it clear where the work with dozens of commits starts and ends. So I suggest to keep merge commits enabled and **use them judiciously**.

Having said that, I completely agree with you that the `covid-19` branch looks messy and merge commits shouldn't be used for every PR, especially those single-commit PRs.

triceo:
@yurloc My approach to work with dozens of commits is "not". :-) I always squash mine - one unit of work = one commit. Works well with agile - small units of work => small merges => decent commit history.

Although I understand that would not be everyone's preference.

yurloc:
@triceo I see your point and I don't have a strong opinion on this but I tend to keep all the commits because:
1. They they document the work being done. You can read the history and get an idea of what has been done on a feature branch.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/solver/EmployeeRosteringConstraintProvider.java
```diff
@@ -0,0 +1,347 @@
...
+                .penalizeConfigurableLong(CONSTRAINT_UNAVAILABLE_TIME_SLOT_FOR_AN_EMPLOYEE,
+                        ((employeeAvailability, shift) -> shift.getLengthInMinutes()));
+    }
+
+    Constraint noOverlappingShifts(ConstraintFactory constraintFactory) {
+        return getAssignedShiftConstraintStream(constraintFactory)
+                .join(Shift.class,
+                        equal(Shift::getEmployee),
+                        lessThanOrEqual(Shift::getStartDateTime),
+                        lessThanOrEqual(Shift::getStartDateTime, Shift::getEndDateTime),
```
Christopher-Chianelli, line 223: Not sure if this is optimal; you are eliminating the `startTime = endTime` case in a filiter. Why not use `lessThan` and `greaterThan` instead of `lessThanOrEqual` and `greaterThanOrEqual`?

Reasoning:

Case 1: Shift A is completely before Shift B
then A.getEndDateTime() < B.getStartDateTime(), so no match

Case 2: Shift A ends when Shift B begins
then A.getEndDateTime() = B.getStartDateTime(), so no match (it a match currently)

Case 3: Shift A starts before Shift B, but ends within Shift B
then A.getEndDateTime > B.getStartDateTime and A.getStartDateTime < B.getEndDateTime, so match

Case 4: Shift A starts when Shift B starts, but ends before Shift B ends
then  A.getEndDateTime > B.getStartDateTime and A.getStartDateTime < B.getEndDateTime, so match

Case 5: Shift B is a subinterval of Shift A
then  A.getEndDateTime > B.getStartDateTime and A.getStartDateTime < B.getEndDateTime, so match

Case 6: Shift A starts when Shift B starts, and ends when Shift B ends
Then  A.getEndDateTime > B.getStartDateTime and A.getStartDateTime < B.getEndDateTime, so match

Case 7: Shift A starts when Shift B starts, and ends after Shift B ends
Then  A.getEndDateTime > B.getStartDateTime and A.getStartDateTime < B.getEndDateTime, so match

The first lessThanOrEqual eliminates all other cases

PR #435, opened 2020-04-29
PLANNER-1958: Allow exporting of Shift Roster to Excel

Christopher-Chianelli:
Current format is:
Start, End, Employee

with the following caveats to maximize human readability:

- Shifts with the same date and time in the same spot don't have their date/time repeated
- Employee(s) are 1 line under date/time

Ex:

Start,       End,           Employee
date 1,     date 2,
                                  Amy
                                  Bob
date 3,     date 4,
                                  Charlie

### DIFF
--- a/optaweb-employee-rostering-frontend/src/ui/pages/shift/ExportScheduleModal.tsx
+++ b/optaweb-employee-rostering-frontend/src/ui/pages/shift/ExportScheduleModal.tsx
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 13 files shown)


optaweb-employee-rostering-frontend/src/ui/pages/shift/ExportScheduleModal.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { useTranslation } from 'react-i18next';
+import MultiTypeaheadSelectInput from 'ui/components/MultiTypeaheadSelectInput';
+import { Spot } from 'domain/Spot';
+import { AppState } from 'store/types';
+import { spotSelectors } from 'store/spot';
+import { connect } from 'react-redux';
+import moment from 'moment';
+
+interface StateProps {
+    tenantId: number;
```
yurloc, line 28: This file is breaking many code style rules. Please follow the ESLint rules that we have configured to ensure consistent code style.

Christopher-Chianelli:
@yurloc fixed your issues; the data being reset when solving was due to shallow equality (and among other pitfalls, Dates are not shallowly equal if they refer to the same date). I fix the shallow equality issue by making selectors return the same list if the idMap did not change since the last call, and using the getTime method on the dates. Although I did disable the export button if scheduling is in process, as it doesn't make sense to export the data while it is changing, but I should prevent the issue in other components and reduce rerendering/recomputation.

PR #436, opened 2020-05-01
PLANNER-1959: Add Employee List Import

Christopher-Chianelli:


Christopher-Chianelli:
See https://issues.redhat.com/browse/PLANNER-1959 for file format. Example file attached.
[EmployeeList.xlsx](https://github.com/kiegroup/optaweb-employee-rostering/files/4689672/EmployeeList.xlsx)

yurloc:
Build failed.

Christopher-Chianelli:
@yurloc should be fixed now

optaweb-employee-rostering-frontend/src/ui/pages/employee/EmployeesPage.tsx
```diff
@@ -209,9 +212,24 @@ export class EmployeesPage extends DataTable<Employee, Props> {
   }
 
   render(): JSX.Element {
+    const importElement = (
+      <div>
+        <FileUpload
```
yurloc, line 217: Suggested improvement: add file type restriction (`xlsx` extension).

yurloc: https://www.patternfly.org/v4/documentation/react/components/fileupload

optaweb-employee-rostering-frontend/src/ui/pages/employee/EmployeesPage.tsx
```diff
@@ -209,9 +212,24 @@ export class EmployeesPage extends DataTable<Employee, Props> {
...
+          id="file"
+          name="file"
+          onChange={
+            (file) => {
+              if (typeof file !== 'string') {
```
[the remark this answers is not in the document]
Christopher-Chianelli: The onChange method returns `string|File` as is also support uploading text from a text area instead. This is so we can make sure it a file.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/contract/ContractService.java
```diff
@@ -103,4 +103,19 @@ public Contract updateContract(Integer tenantId, ContractView contractView) {
...
+                .stream().filter(contract -> contract.getName().equals("Default Contract"))
+                .findAny();
+        if (defaultContract.isPresent()) {
+            return defaultContract.get();
+        } else {
+            Contract contract = new Contract();
+            contract.setName("Default Contract");
+            contract.setTenantId(tenantId);
+            return contractRepository.save(contract);
+        }
```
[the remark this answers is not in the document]
Christopher-Chianelli: The PR was created when we were merging features into the `covid-19` branch without testing. Since I have a separate PR that now add this testing to the `covid-19` branch, I will add tests to this PR.

optaweb-employee-rostering-frontend/src/ui/pages/employee/EmployeesPage.tsx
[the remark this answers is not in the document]
yurloc: Okay, then `if (typeof file === File)` would make better sense, wouldn't it?

Also, `string` is only passed to onChange when FileUpload's type is text or dataURL so the type check might be redundant.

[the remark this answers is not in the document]
yurloc: Update: `typeof` can't be used with `File`. So my proposal should look like this: `if (file instanceof File)`.

yurloc:
The SQL error after uploading the file for the second time spawns an interesting question - what should happen in such case?

Proposal A) Forbid uploading data if the there already are some employees in the tenant.

Proposal B) Detect existing employees and refuse to overwrite them. Write the new ones or refuse the whole batch?

Proposal C) Update everything. So if I upload _the same_ file, it will be successful but effectively nothing happens. I can use this technique to make changes in the XLSX and update employees on the server by uploading the updated XLSX. I doubt that is useful in practice.

Proposal A seems to be the safest way to go.

yurloc:
UX-wise, the upload input blends in with the background. I'm sure it could be fixed by nesting it inside some other PF form element (Form? InputGroup?). Or moving it to a modal dialog that pops up after clicking a nice blue "Upload employees" button.

This is just a comment about a non-blocking issue.

Christopher-Chianelli:
What is currently implemented is Proposal B (with refuse the whole batch), although we can have a better error message. I dislike Proposal A as people might want to upload a new batch of Employees. I prefer C

Christopher-Chianelli:
Implemented C)

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/employee/EmployeeController.java
```diff
@@ -84,6 +98,40 @@ public EmployeeController(EmployeeService employeeService) {
         return new ResponseEntity<>(employeeService.createEmployee(tenantId, employeeView), HttpStatus.OK);
     }
 
+    @ApiOperation("Import employees from an Excel file")
+    @PostMapping("/import")
+    public ResponseEntity<List<Employee>> addEmployeesFromExcelFile(@PathVariable @Min(0) Integer tenantId,
+                                                                    @RequestParam("file") MultipartFile excelDataFile)
+            throws IOException {
+
+        List<EmployeeView> excelEmployeeList = employeeListXlsxFileIO
```
yurloc, line 107: All of this belongs to `EmployeeService`. Motivation:
- Less dependencies for the controller (only the service).
- Clear separation of concerns. Controller = network layer, Service = application layer.
- Imagine we have a CLI in addition to REST API. The CLI shouldn't depend on a REST controller in order to read employees from a file. The CLI should only rely on the service to do that.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/util/EmployeeListXlsxFileIO.java
```diff
@@ -0,0 +1,93 @@
...
+        try (Workbook workbook = new XSSFWorkbook(excelFileStream)) {
+            Sheet worksheet = workbook.getSheetAt(0);
+            List<EmployeeView> out = new ArrayList<>(worksheet.getPhysicalNumberOfRows() - 1);
+            Map<String, Skill> skillMap = skillService.getSkillList(tenantId).stream()
+                    .collect(Collectors.toMap(s -> s.getName().toLowerCase(), Function.identity()));
+            Contract defaultContract = contractService.getOrCreateDefaultContract(tenantId);
+            for (int i = 1; i < worksheet.getLastRowNum(); i++) {
+                Row row = worksheet.getRow(i);
+
+                if (row.getCell(0) == null) {
```
yurloc, line 67: Throws NPE if the row is empty.

Christopher-Chianelli: The Irony of getting a NPE on a Null Pointer check... will fix

optaweb-employee-rostering-backend/src/test/java/org/optaweb/employeerostering/employee/EmployeeServiceTest.java
```diff
@@ -763,4 +774,82 @@ public void updateNonExistentEmployeeAvailabilityTest() throws Exception {
...
+                             expected.getCovidRiskType(), actual.getCovidRiskType());
+                assertEquals("Wrong tenant id for " + expected.getName(),
+                             expected.getTenantId(), actual.getTenantId());
+                assertEquals("Wrong number of skills for " + expected.getName(),
+                             expected.getSkillProficiencySet().size(), actual.getSkillProficiencySet().size());
+                expected.getSkillProficiencySet().forEach(skill -> assertTrue("Missing skill " + skill.getName() +
+                                                                                      " for employee " + 
+                                                                                      expected.getName(),
+                                                                              actual.getSkillProficiencySet().stream()
+                                                                                      .filter(s -> s.getName()
```
yurloc, line 841: Can replace `filter().findAny().isPresent()` with `anyMatch()`.

I also recommend replacing JUnit assertions with AssertJ. With AssertJ this can be written as:
```java
assertThat(actual.getSkillProficiencySet())
        .as("Unexpected skill set for employee %s", expected.getName())
        .extracting(Skill::getName)
        .containsExactlyInAnyOrderElementsOf(expected.getSkillProficiencySet().stream()
                .map(Skill::getName)
                .collect(Collectors.toSet()));
```
which gives you additional info describing the difference between the actual and expected skill sets:
```
java.lang.AssertionError: [Unexpected skill set for employee My Name] 
Expecting:
  <["Nur se"]>
to contain exactly in any order:
  <["Nurse"]>
elements not found:
  <["Nurse"]>
and elements not expected:
  <["Nur se"]>
```

Christopher-Chianelli: I don't think we have AssertJ as a test dependency; I do prefer AssertJ tests as they are easier to read + write.

yurloc: We do. The AssertJ dependency is brought by `spring-boot-starter-test`. AssertJ assertions are already being used in this project.

Christopher-Chianelli: Odd, Eclipse doesn't show any import recommendations when I typed in `assertThat`.

PR #439, opened 2020-05-08
Add Covid-19 class diagram to docs

Christopher-Chianelli:


Christopher-Chianelli:
The integration test fails somehow from this change; I tested the covid-19 branch locally and the integration test passes. I don't think adding images to docs could fail the test; merging as false positive.

PR #441, opened 2020-05-12
PLANNER-1975: Merge non-COVID-19 specific features into master

Christopher-Chianelli:
Do not merge; still need tests


optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/employee/Employee.java
```diff
@@ -122,6 +123,7 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        return 31 * ((31 * name.hashCode()) ^ contract.hashCode()) ^ skillProficiencySet.hashCode();
+        return 31 * ((31 * name.hashCode()) ^ contract.hashCode())
+                ^ skillProficiencySet.hashCode();
     }
```
triceo, line 128: Do we still need this insanity when we no longer use GWT?
How about just `Objects.hash()`?

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/tenant/RosterConstraintConfiguration.java
```diff
@@ -30,24 +30,19 @@
 @ConstraintConfiguration(constraintPackage = "org.optaweb.employeerostering.service.solver")
 public class RosterConstraintConfiguration extends AbstractPersistable {
```
triceo, line 31: Any chance you could reuse my ConstraintProvider PR here? It refactors this class to use String constants to describe constraints, and therefore reduces duplication and potential for human error.

optaweb-employee-rostering-frontend/public/assets/images/extreme.svg
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
```
triceo, line 1: I think these files shouldn't be in this PR, right? Incl. the COVID diagrams etc.

optaweb-employee-rostering-frontend/public/assets/translations/en/Common.yaml
```diff
@@ -26,4 +26,11 @@ selectEmployee: Select an Employee...
...
-  UNAVAILABLE: Unavailable
\ No newline at end of file
+  UNAVAILABLE: Unavailable
+
+CovidRisk:
+  INOCULATED: Inoculated
+  LOW: Low
+  MODERATE: Moderate
+  HIGH: High
+  EXTREME: Extreme
```
triceo, line 36: Again, probably shouldn't be here.

optaweb-employee-rostering-frontend/src/store/employee/employee.test.ts
```diff
@@ -236,6 +236,7 @@ describe('Employee selectors', () => {
         maximumMinutesPerMonth: 10,
         maximumMinutesPerYear: null,
       },
+      covidRiskType: 'INOCULATED',
```
triceo, line 239: Maybe go over the entire PR and grep for `covid`? :-)

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/tenant/RosterConstraintConfiguration.java
triceo: We might as well just merge the ConstraintProvider work - even if we don't enable it (and we shouldn't!), it doesn't hurt for the provider to be there. Mario and I are working on CS-D performance PoCs right now.

Christopher-Chianelli: I say merging the ConstraintProvider work is out of scope for this PR, but it could be done in a separate PR (although it might confuse people why there are two implementations. @ge0ffrey what do you think?

Christopher-Chianelli:
I think I covered all the major areas that needed tests (after checking what is covered in Sonar Cloud). Let me know if you see an area that should have tests that isn't covered.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/tenant/RosterConstraintConfiguration.java
triceo: The PR actually has two parts:
- One is the constraint provider,
- but the more important part is that is refactors the app to reuse constraint names across the app. Which, I would argue, is pretty important.

Christopher-Chianelli:
IT tests are failing; investigating (and is it just me, or is there an abnormal amount of logging in the CI? Took me forever to scroll to the test that failed).

Christopher-Chianelli:
IT test passes locally.... can we rerun the CI @yurloc ?

Christopher-Chianelli:
My guess: maybe timeout? The CI failed on a status call to the Solver Manager

Christopher-Chianelli:
Rerunning CI (turns out there a rerun button next to the CI in the "Checks" page.

Christopher-Chianelli:
IT test passes on Jenkins, so look like the IT test might need to have timeouts increased for GitHub CI

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/tenant/RosterConstraintConfiguration.java
Christopher-Chianelli: The second point looks good; your PR has merge conflicts currently. I would prefer to merge it separately to keep the scope of this PR manageable.

Christopher-Chianelli:
Okay, I think the "rerun" button on the checks page does not work; I am unable to see the actual test being run and the travis check still refers to the old build.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/tenant/RosterConstraintConfiguration.java
triceo: Merge conflicts solved, PR updated.
Please port at your earliest convenience.

yurloc:
@Christopher-Chianelli Have you tried reruning the Travis build?

yurloc:
I mean on Travis, not on the GitHub checks page.

Christopher-Chianelli:
It also fail with the same error on #434. So I think Travis is overwhelmed with traffic, causing cypress to timeout on some requests, as I cannot reproduce the error locally. Can you try running integration tests locally @yurloc ? If it also passes for you, we can try increasing the timeouts to see if that helps.

Christopher-Chianelli:
Or rather, it did fail on #434 before, so look like it is probably timeouts

yurloc:
@Christopher-Chianelli running the build locally.

Not sure about the timeouts. I was under the impression that Cypress has flexible timeouts that reduce test flakiness. A rerun of the Travis build could have give us a hint of what's going on.

Have you tried rerunning the build on Travis? If we only have a single failing build it's hard to tell what's going on. If it fails twice in a row, the test needs tweaking (even if it passes locally). If the rerun is successful, it's probably just bad weather.

Christopher-Chianelli:
For referenced, the past three builds for this PR passed after checking the Travis build history. When it fails, it always fails on a call to SolverStatus.

Christopher-Chianelli:
Kinda hesitant to sign into Travis as it involves giving Travis write access to all of my repositories. (GitHub probably need a better permission system)

Christopher-Chianelli:
I need to update this PR, so travis CI will rerun then

triceo:
Wrt. Travis - Github Actions is a fairly nice way of doing CI that is internal to Github, so many of the security issues go away. Actions is pretty powerful, actually.

yurloc:
@Christopher-Chianelli I didn't know you haven't signed up on Travis. Not going to force you to do that if you have concerns, especially when an alternative (Actions) is on the table now :slightly_smiling_face:. let me know and I can re-trigger the build for you.

Christopher-Chianelli:
I authorised Travis (look like GitHub need to update it permission system; Travis need write access as there no such thing as read:private). So hopefully no bad actor decides to wipe out all of my repositories via Travis CI.

Christopher-Chianelli:
(Plus, I can revoke it access if we switch to something else)

Christopher-Chianelli:
Travis now passes

Christopher-Chianelli:
Failed again on a different call, so it almost definitely time outs it seems

triceo:
I ran the integration tests a couple times on my own machine and did not reproduce the timeout.
Since we'll be removing Travis very soon, let's see what this does on Github Actions.
Merging as agreed.

PR #442, opened 2020-05-28
Enable Github actions

triceo:


triceo:
I had to merge this, as when GA is being first set up, there is no way to test it in a PR. Travis will be removed in a separate PR, once I've made sure GA works properly.

PR #443, opened 2020-05-28
PLANNER-2000 Replace Travis CI with Github Actions

triceo:


triceo:
In order to enable Sonar properly, we need to add `SONARCLOUD_TOKEN` among secrets in Github. I don't seem to have the permissions to be able to do that. Christopher, can you please take care of it?

Christopher-Chianelli:
@triceo it seems I don't have permission to do that either; can't access repository's settings. We probably need the KIE group maintainer to set it for us.

triceo:
@Christopher-Chianelli Can you please take care of that? Although I'm really not sure why Sonar's even done in Travis.

yurloc:
Because it's faster.

Christopher-Chianelli:
@triceo Done; "SONARCLOUD_TOKEN" is defined; thanks @rsynek!

triceo:
@rsynek graciously agreed to look into the Sonarcloud failure.

pom.xml
```diff
@@ -198,54 +198,5 @@
...
-        <property>
-          <name>env.TRAVIS_EVENT_TYPE</name>
-          <value>pull_request</value>
-        </property>
-      </activation>
-      <!-- See https://docs.travis-ci.com/user/environment-variables/ for list of Travis environment variables. -->
-      <!-- See https://docs.sonarqube.org/latest/analysis/pull-request/ for properties required by Sonar for PR analysis. -->
-      <properties>
-        <!--suppress UnresolvedMavenProperty -->
-        <sonar.pullrequest.branch>${env.TRAVIS_PULL_REQUEST_BRANCH}</sonar.pullrequest.branch>
```
rsynek, line 228: How does GitHub Actions CI provide these properties? Without these, the PR won't get paired with SonarCloud.

.github/workflows/ci.yml
```diff
@@ -2,6 +2,9 @@ name: CI
 
 on: [push, pull_request]
 
+env:
+  SONAR_TOKEN: ${{ secrets.SONARCLOUD_TOKEN }}
```
rsynek, line 6: Is it required to use `SONAR_TOKEN` environment variable name as the `SONARCLOUD_TOKEN` has been already taken by the secret? If so, we can rename the secret (see the next comment).

pom.xml
triceo: Somehow Sonar can read them from the environment, see the log - it has all the info.

triceo:
Ok, together with @rsynek we've finally discovered why Sonar doesn't work: As a security measure, the `SONARCLOUD_TOKEN` secret is not passed to PRs from external repositories.

Travis had the same problem, but there we actually circumvented the security precautions and set the secret to env variable directly in Travis settings. Github Actions doesn't allow that, and good for them.

The only solution here is to rely on Jenkins to run Sonar analysis for us. And since most of the KIE repositories can live with that, I would argue that we can, too. 

I'm merging this PR as is - if there are any other solutions to the problem, I'll be happy to discuss them in a follow-up conversation.

triceo:
The following PR enables Sonarcloud for this repo in Jenkins:
https://github.com/kiegroup/kie-jenkins-scripts/pull/751

PR #444, opened 2020-05-29
RHDM-1260: Show a "Connection lost" message when the frontend fails to connect to backend

Christopher-Chianelli:


optaweb-employee-rostering-frontend/public/assets/translations/en/ConnectionStatus.yaml
```diff
@@ -0,0 +1,3 @@
+cannotConnect: Cannot Connect
+cannotConnectMsg:
+  Please check your network connection. When connection is available the application will be functional again.
```
triceo, line 3: I wouldn't make those kinds of promises. :-)

My suggestion: "No connection to server"
"Connection to the planning backend has been lost. Please check your network connection."

Christopher-Chianelli: Copied from https://github.com/kiegroup/optaweb-vehicle-routing/blob/039716a4ee333144abac9286523dbd1460ffa832/optaweb-vehicle-routing-frontend/src/ui/connection/ConnectionManager.tsx#L55,

I can reword it

Christopher-Chianelli: The "When connection is available the application will be functional again." is meant to inform the user that if the backend comes online, it will automatically reconnect. Although we might not want to promise "functional".

Christopher-Chianelli:
@rsynek do you know how to remove files in certain directories from test coverage? FIles in a `__mocks__` directory do not need to be tested (as they are mocks for tests)

yurloc:
Issue 1: Connection failure is only detected when switching to another tenant. It's not detected when navigating between pages that refresh data. And it's not detected when making requests like Refresh, Schedule, Publish.

optaweb-employee-rostering-frontend/src/store/store.ts
```diff
@@ -77,5 +78,6 @@ export function configureStore(
   );
 
   restServiceClient.setDispatch(store.dispatch);
+  restServiceClient.setStore(store);
```
yurloc, line 81: Circular dependency. REST client needs store (to check connection status - I think it shouldn't) and a REST client instance is needed to create the store. I would classify this as a code smell.

optaweb-employee-rostering-frontend/src/store/rest/RestServiceClient.ts
```diff
@@ -61,15 +69,17 @@ export default class RestServiceClient {
     if (res.status >= 200 && res.status <= 300) {
       return Promise.resolve(res.data);
     }
-    if (this.dispatch !== null) {
-      if (typeJsonRegex.test(res.headers['content-type'])) {
+    if (this.dispatch !== null && this.store !== null) {
+      if (!this.store.getState().isConnected) {
```
yurloc, line 73: I would argue it's not REST client's business to check store state or dispatch actions. In my opinion it should simply reject the promise with as much _useful_ information as reasonably possible. Then it's operation's responsibility to act upon a rejected promise of a get/post/delete/... call and dispatch a corresponding action based on that.

optaweb-employee-rostering-frontend/src/store/tenant/operations.ts
```diff
@@ -80,6 +87,11 @@ ThunkCommandFactory<void, RefreshTenantListAction> = () => (dispatch, state, cli
       dispatch(actions.refreshTenantList({ tenantList, currentTenantId: 0 }));
     }
     refreshData(dispatch);
+  }).catch(() => {
```
yurloc, line 90: OK, this explains my question why only switching between tenants detects a connection failure. I think every operation needs to do this.

rsynek:
> @rsynek do you know how to remove files in certain directories from test coverage? FIles in a `__mocks__` directory do not need to be tested (as they are mocks for tests)
@Christopher-Chianelli I wonder if adding them here https://github.com/kiegroup/optaweb-employee-rostering/blob/master/optaweb-employee-rostering-frontend/pom.xml#L35 might help.

See https://docs.sonarqube.org/latest/project-administration/narrowing-the-focus/.

yurloc:
Issue 2: I don'ŧ think this solves RHDM-1260 because
> Server Exception: The object you requested was not found: No RosterState entity found with tenantId (0). See Stack Trace.

alert is still shown.

optaweb-employee-rostering-frontend/src/store/tenant/operations.ts
Christopher-Chianelli: Issue is, operations can fail without meaning no connections. I think it better to handle generic connection failures in the RestServiceClient so we don't need to repeat the code everywhere

optaweb-employee-rostering-frontend/src/store/rest/RestServiceClient.ts
Christopher-Chianelli: Many of these operations have a common error handler (namely: display an alert if an operation fails). Not handling it here will result in massive code duplication.

Christopher-Chianelli:
@yurloc Issue 1 and 2 are fixed. I disagree with handling errors in the operations, since all errors are handled in the same way and thus it leads to massive code duplication, and is more error phone as one can forget to add the catch. I managed to remove the state dependency in the RestClient, but it still need dispatch so it can send alerts and set connection status.

optaweb-employee-rostering-frontend/src/store/tenant/reducers.ts
```diff
@@ -45,4 +47,13 @@ const tenantReducer = (state = initialState, action: TenantAction): TenantData =
   }
 };
 
+export const connectionReducer = (state = initialConnectionState, action: ConnectAction): boolean => {
```
yurloc, line 50: Belongs to a separate file inside a connection `duck`. But the `store` is already breaking `ducks` organization on many places. Let's solve this outside of this PR.

optaweb-employee-rostering-frontend/src/store/rest/RestServiceClient.test.ts
```diff
@@ -73,6 +77,21 @@ describe('Rest Service Client', () => {
...
+    const handleErrorSpy = jest.spyOn(restServiceClient, 'handleError');
+    const axiosError = { isAxiosError: false };
+    const response = new Promise((res, reject) => { reject(axiosError); });
+    mockGet.mockReturnValue(response);
+    try {
+      await restServiceClient.get(targetURL);
+    } catch (error) {
+      expect(handleErrorSpy).toBeCalledWith(axiosError);
+    }
+  });
```
yurloc, line 93: Lot of repeated code. Optional issue.

optaweb-employee-rostering-frontend/src/store/rest/RestServiceClient.test.ts
```diff
@@ -73,6 +77,21 @@ describe('Rest Service Client', () => {
     expect(handleResponseSpy).toBeCalledWith(response);
   });
 
+  it('Should call errorHandler if axios.get fails', async () => {
+    const baseURL = '/rest';
+    const restServiceClient = new RestServiceClient(baseURL, axios);
+    const targetURL = '/endpoint';
+    const handleErrorSpy = jest.spyOn(restServiceClient, 'handleError');
+    const axiosError = { isAxiosError: false };
+    const response = new Promise((res, reject) => { reject(axiosError); });
```
yurloc, line 86: Redundant braces around arrow function body. Optional issue.

PR #451, opened 2020-06-15
Export Schedule, Import Employees, Java Constraint Provider

Christopher-Chianelli:
This merges the rest of the work based on covid-19 branch


Christopher-Chianelli:
jenkins retest this

yurloc:
@Christopher-Chianelli Too early, I'm afraid. https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1364 needs to be merged to make the build green again.

mareknovotny:
Jenkins retest this please

yurloc:
Jenkins retest this.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/roster/RosterController.java
```diff
@@ -108,6 +120,36 @@ public RosterController(RosterService rosterService) {
...
+    public ResponseEntity<byte[]> getShiftRosterViewAsExcel(@PathVariable @Min(0) Integer tenantId,
+                                                            @RequestParam(name = "startDate")
+                                                                    String startDateString,
+                                                            @RequestParam(name = "endDate") String endDateString,
+                                                            @RequestParam(name = "spotList") String spotListString) {
+        Set<Long> spotIdSet = Arrays.stream(spotListString.split(",")).map(Long::parseLong)
+                .collect(Collectors.toSet());
+        List<Spot> spotList = spotRepository.findAllByTenantId(tenantId, PageRequest.of(0, Integer.MAX_VALUE))
+                .stream().filter(s -> spotIdSet.contains(s.getId()))
+                .collect(Collectors.toList());
```
yurloc, line 135: Important piece of application logic that should have a test coverage. Also, filtering spots by ID is outside of the controller's responsibility. This should be done by the `RosterService`.

`RosterServiceTest`:
- _given_ **spots** in the repository and a **list** of spot IDs,
- _when_ roster view is requested for the given **list**,
- _then_ roster view is returned with spots from an intersecion of IDs in the list and the repository.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/roster/RosterController.java
```diff
@@ -108,6 +120,36 @@ public RosterController(RosterService rosterService) {
...
 
+    @ApiOperation("Get a shift roster view between two dates for a subset of the spots as an excel file")
+    @GetMapping(value = "/shiftRosterView/excel",
+            produces = "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet")
+    public ResponseEntity<byte[]> getShiftRosterViewAsExcel(@PathVariable @Min(0) Integer tenantId,
+                                                            @RequestParam(name = "startDate")
+                                                                    String startDateString,
+                                                            @RequestParam(name = "endDate") String endDateString,
+                                                            @RequestParam(name = "spotList") String spotListString) {
+        Set<Long> spotIdSet = Arrays.stream(spotListString.split(",")).map(Long::parseLong)
```
yurloc, line 131: If the list contains a bad value, we get an uncaught exception. The try/catch below should probably cover this as well. 
The `RosterControllerTest` should cover this:
- wrong spot ID list => error response entity
- good spot ID list => call roster service with the same list of IDs

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/util/ShiftRosterXlsxFileIO.java
```diff
@@ -0,0 +1,94 @@
...
+public final class ShiftRosterXlsxFileIO {
+
+    public static final DateTimeFormatter DATE_TIME_FORMATTER
+            = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm", Locale.ENGLISH);
+
+    // Cannot build instances of this, so private empty constructor
+    private ShiftRosterXlsxFileIO() {
+    }
+
+    public static byte[] getExcelBytesForShiftRoster(ShiftRosterView shiftRosterView) throws IOException {
```
yurloc, line 46: No test coverage. At least verify that the number of sheets in the workbook equals to the number of spots, etc.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/util/ShiftRosterXlsxFileIO.java
```diff
@@ -0,0 +1,94 @@
...
+                headerCell = headerRow.createCell(2);
+                headerCell.setCellValue("Employee");
+
+                int rowNumber = 1;
+                LocalDateTime lastShiftStartDateTime = null;
+                LocalDateTime lastShiftEndDateTime = null;
+                for (ShiftView shift : shiftRosterView.getSpotIdToShiftViewListMap().get(spot.getId())) {
+                    Row shiftRow = sheet.createRow(rowNumber);
+                    if (!shift.getStartDateTime().equals(lastShiftStartDateTime) ||
+                            !shift.getEndDateTime().equals(lastShiftEndDateTime)) {
```
yurloc, line 68: This deserves a test.

yurloc: Hint: writing tests forces us to write testable code.

In this case, the method `getExcelBytesForShiftRoster()` needs to break down into A) _roster model_ -> _excel model (workbook)_, B) _workbook_ -> _bytes_.

Christopher-Chianelli:
@yurloc I was waiting for Sonarcloud to run so I can look up what needs test coverage as this is the remainder of the work on the covid-19 (untested) branch. Thanks for the review and pin-pointing areas that need testing.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/roster/RosterController.java
Christopher-Chianelli: It appears `RosterService` does not have unit tests, unlike the other services. Out of scope for this PR (This PR will only have the controller tests as a result; the unit tests for RosterService will come in a different PR)

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/util/ShiftRosterXlsxFileIO.java
Christopher-Chianelli: I don't understand your hint. It is easy to check the workbook contents by reading the bytes in a workbook, and functions calling the method expect the bytes representing the workbook not the workbook itself. Additionally, unlike A), B) is extremely simple.

yurloc: The hint was a shortcut for "tests imply testable code" and "testable code is better than code that is hard to test".

The point is that a code that is testable usually also exhibits some good design qualities, for example Single Responsibility Principle. Writing tests guarantees testable code. Therefore writing test sometimes forces us to improve the code design simply out of necessity.

I don't think that it is easy to verify the workbook content by reading bytes. At least it's not necessary. This code is not concerned about how the workbook is represented in bytes. That's Apache POI's business. Our concern are the sheets, rows and columns in the workbook.

The second part of my previous comment suggests to break the method down into smaller pieces, specifically:
```java
public static Workbook workbook(ShiftRosterView shiftRosterView) throws IOException { ... }
public static byte[] bytes(Workbook workbook) throws IOException { ... }
```
If you do this, you can:
1. Have a test that focuses on the workbook content without jumping through hoops (reading bytes):
    ```java
    // given
    ShiftRosterView shiftRosterView = new ...;
    // when
    Workbook workbook = ShiftRosterXlsxFileIO.workbook(shiftRosterView);
    // then
    assertThat(workbook.getNumberOfSheets()).isEqualTo(...);
    // deeper inspection...
    ```
2. You can even test the second part of the story:
    ```java
    // given
    Workbook workbook = new ...;
    // when
    byte[] bytes = ShiftRosterXlsxFileIO.bytes(workbook);
    // then
    // ... (convert bytes to an input stream) ...
    // verify reading the output bytes gives the original workbook
    assertThat(new Workbook(inputStream)).isEqualTo(workbook);
    ```

All of this without going through the controller and initializing Spring context and web environment.

> ...and functions calling the method expect the bytes representing the workbook not the workbook itself.

And, of course, you can keep the original method and let it compose the smaller pieces:
```java
public static byte[] getExcelBytesForShiftRoster(ShiftRosterView shiftRosterView) throws IOException {
    return bytes(workbook(shiftRosterView));
}
```

optaweb-employee-rostering-backend/src/test/java/org/optaweb/employeerostering/solver/AbstractSolverTest.java
```diff
@@ -70,37 +64,43 @@
...
+import static org.optaweb.employeerostering.domain.tenant.RosterConstraintConfiguration.CONSTRAINT_WEEKLY_MINUTES_MUST_NOT_EXCEED_CONTRACT_MAXIMUM;
+import static org.optaweb.employeerostering.domain.tenant.RosterConstraintConfiguration.CONSTRAINT_YEARLY_MINUTES_MUST_NOT_EXCEED_CONTRACT_MAXIMUM;
+
+public abstract class AbstractSolverTest {
 
     private static final int TENANT_ID = 0;
     private static final LocalDate START_DATE = LocalDate.of(2019, 5, 13);
     private static final RosterConstraintConfiguration ROSTER_CONSTRAINT_CONFIGURATION =
             new RosterConstraintConfiguration();
+    private static final String ROSTER_PATH_URI = "http://localhost:8080/rest/tenant/{tenantId}/roster/";
```
yurloc, line 86: Duplicates `rosterPathURI` on line 91.

optaweb-employee-rostering-backend/src/test/java/org/optaweb/employeerostering/roster/RosterRestControllerTest.java
```diff
@@ -446,7 +470,101 @@ public void testPublishAndProvision() {
...
+        assertThat(spotSheet.getRow(0).getCell(0).getStringCellValue())
+                .as("Header Column A should be Start").isEqualTo("Start");
+        assertThat(spotSheet.getRow(0).getCell(1).getStringCellValue())
+                .as("Header Column B should be End").isEqualTo("End");
+        assertThat(spotSheet.getRow(0).getCell(2).getStringCellValue())
+                .as("Header Column C should be Employee").isEqualTo("Employee");
+
+        // Sort startEndTimePairs by start time then end time
+        List<List<LocalDateTime>> sortedStartEndTimePairs = requestShiftMapByStartAndEndTime.keySet().stream()
+                .sorted(Comparators.lexicographical(LocalDateTime::compareTo))
```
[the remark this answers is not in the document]
Christopher-Chianelli: I don't recall adding any dependencies to the `pom.xml`, so any dependency on `Guava` was there before. I might change it depending on how much cleaner the code looks. (code would probably look better with Records (https://openjdk.java.net/jeps/359), but we can't use them until minimum release is Java 14).

optaweb-employee-rostering-backend/src/test/java/org/optaweb/employeerostering/solver/AbstractSolverTest.java
```diff
@@ -668,31 +668,38 @@ private RosterConstraintConfiguration getRosterConstraintConfiguration(AtomicLon
...
-        ASSIGN_EVERY_SHIFT("Assign every shift",
-                ROSTER_CONSTRAINT_CONFIGURATION.getAssignEveryShift().negate()),
-        UNDESIRED_TIME_SLOT_FOR_AN_EMPLOYEE("Undesired time slot for an employee",
-                ROSTER_CONSTRAINT_CONFIGURATION.getUndesiredTimeSlot().negate()),
-        DESIRED_TIME_SLOT_FOR_AN_EMPLOYEE("Desired time slot for an employee",
-                ROSTER_CONSTRAINT_CONFIGURATION.getDesiredTimeSlot()),
-        EMPLOYEE_IS_NOT_ROTATION_EMPLOYEE("Employee is not rotation employee",
-                ROSTER_CONSTRAINT_CONFIGURATION.getNotRotationEmployee().negate());
+        REQUIRED_SKILL_FOR_A_SHIFT(CONSTRAINT_REQUIRED_SKILL_FOR_A_SHIFT,
+                                   ROSTER_CONSTRAINT_CONFIGURATION.getRequiredSkill().negate()),
```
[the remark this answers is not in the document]
Christopher-Chianelli: I think this is the IDEA KIE team indentation; I have `git` auto-format before a commit using the IDEA KIE style. Although I agree we should employ the OptaPlanner auto-formatter here too.

Christopher-Chianelli:
@yurloc The majority of time is spent in integration tests, that run for 30 seconds each (namely, making sure the rosters we generate are feasible). I might be able to reduce the time the solver runs to 1-2 seconds, although the test mail fail on slower machines.

Christopher-Chianelli:
(or maybe editing the integration test to use  `withBestScoreLimit`)

Christopher-Chianelli:
@yurloc tests should now only take ~"30 seconds" (still double of vehicle routing; the longest test is the Admin test, which takes 10 seconds (need to set up spring context, create roster, delete all the entities created, and recreate all those entities).

optaweb-employee-rostering-backend/src/test/java/org/optaweb/employeerostering/roster/RosterRestControllerTest.java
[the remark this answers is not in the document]
yurloc: > I don't recall adding any dependencies to the pom.xml, so any dependency on Guava was there before.

Correct. This is called a transitive dependency (aka undeclared used dependency). You're depending on something that one of your declared dependencies uses but you don't declare it yourself. This is dangerous from the maintenance perspective. Two bad things may happen:
- **You remove** the dependency that transitively provides Guava because you're replacing the technology or dropping the feature. Easy to fix: you declare Guava dependency.
- You or **your parent upgrades** the dependency that transitively provides Guava but the new version no longer depends on it. That will be annoying for the parent maintainer because the FDB will fail surprisingly. They will need to go to this repo and declare the dependency. Wouldn't happen if the dependency was already declared.

PR #458, opened 2020-06-29
PLANNER-1452: Make it easier and quicker to create a rotation

Christopher-Chianelli:


optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/common/ColorUtils.java
```diff
@@ -0,0 +1,43 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaweb.employeerostering.domain.common;
+
+public class ColorUtils {
```
triceo, line 19: This is a prime use case for an enum.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/employee/Employee.java
```diff
@@ -65,6 +77,18 @@ public Employee(Integer tenantId, String name, Contract contract,
                     Set<Skill> skillProficiencySet) {
         super(tenantId);
         this.name = name;
+        this.shortId = Arrays.stream(name.split(" ")).limit(3).reduce("", (a,b) -> a + b.charAt(0));
```
triceo, line 80: This deserves extraction into a separate function, one that could be named in an explanatory manner.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/rotation/TimeBucket.java
```diff
@@ -0,0 +1,210 @@
...
+    @CollectionTable(name = "seat_list",
+        joinColumns = {
+        @JoinColumn(name = "seat_id",
+            referencedColumnName = "id",
+            foreignKey=@ForeignKey(name="SEAT_FK",
+            foreignKeyDefinition = "FOREIGN KEY (seat_id) references public.time_bucket (id)" +
+            " ON UPDATE NO ACTION ON DELETE CASCADE"))
+        }
+    )
+    private List<Seat> seatList;
```
triceo, line 77: I am not familiar with this part of JPA, but the syntax doesn't look particularly DB-agnostic. Which DBs did you check?

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/rotation/TimeBucket.java
```diff
@@ -0,0 +1,210 @@
...
+        this.additionalSkillSet = additionalSkillSet;
+        this.repeatOnDaySet = repeatOnDaySet;
+        this.seatList = new ArrayList<>(rotationLength);
+        DayOfWeek rotationDay = startOfWeek;
+        for (int i = 0; i < rotationLength; i++) {
+            if (repeatOnDaySet.contains(rotationDay)) {
+                seatList.add(new Seat(i,null));
+            }
+            rotationDay = rotationDay.plus(1);
+        }
```
triceo, line 97: Please extract into a separate function. Constructors ought to be short and concise.
Also, this could be replaced by a relatively simple stream.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/rotation/TimeBucket.java
```diff
@@ -0,0 +1,210 @@
...
+        setSeatList(updatedTimeBucket.getSeatList());   
+    }
+    
+    public Optional<Shift> createShiftForOffset(LocalDate startDate, int offset, ZoneId zoneId,
+                                                boolean defaultToRotationEmployee) {
+        Optional<Seat> maybeSeat = seatList.stream().filter(seat -> seat.getDayInRotation() == offset).findAny();
+        
+        if (!maybeSeat.isPresent()) {
+            return Optional.empty();
+        }
```
triceo, line 186: Although I like the use of `Optional`, the `isPresent()` makes things a bit ugly.
I suggest refactoring this method so that `.findAny().map()` can be used. Don't be afraid of extracting some of the logic into another method.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/rotation/TimeBucket.java
```diff
@@ -0,0 +1,210 @@
...
+        LocalDateTime startDateTime = startDate.atTime(getStartTime());
+        LocalDate endDate;
+        
+        if (getStartTime().isBefore(getEndTime())) {
+            // Ex: 9am-5pm
+            endDate = startDate;
+        } else {
+            // Ex: 9pm-6am
+            endDate = startDate.plusDays(1);
+        }
```
triceo, line 197: LocalDate endDate = (getStartTime().isBefore(getEndTime())) ?
       startDate  :
       startDate.plusDays(1);

Nicer, isn't it?

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/roster/RosterGenerator.java
```diff
@@ -619,47 +624,43 @@ public RosterState createRosterState(GeneratorType generatorType, Tenant tenant,
...
-                    final int dayOfWeek = (startDayOffset % 7) + 1;
-                    // Only generate shifts for days in the timeslot
-                    if (timeslotRange.getRight().stream().anyMatch(d -> d.getValue() == dayOfWeek)) {
-                        LocalTime startTime = timeslotRange.getLeft();
-                        LocalTime endTime = timeslotRange.getMiddle();
-                        int endDayOffset = startDayOffset;
-                        if (endTime.compareTo(startTime) < 0) {
-                            endDayOffset = (startDayOffset + 1) % rotationLength;
-                        }
+            final AtomicInteger timeslotRangesIndex = new AtomicInteger(0);
```
triceo, line 627: The fact that you're mutating state outside of the stream suggests that maybe streams are not the answer here.
Especially since you have a stream within a stream, and a for loop following it.

I suggest refactoring and breaking down into smaller methods in the process.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/roster/RosterGenerator.java
```diff
@@ -668,48 +669,51 @@ public RosterState createRosterState(GeneratorType generatorType, Tenant tenant,
...
         List<Shift> shiftList = new ArrayList<>();
-        Map<Pair<Integer, Spot>, List<ShiftTemplate>> dayOffsetAndSpotToShiftTemplateListMap = shiftTemplateList
-                .stream()
-                .collect(groupingBy(shiftTemplate -> Pair.of(shiftTemplate.getStartDayOffset(),
-                                                             shiftTemplate.getSpot())));
-        int dayOffset = 0;
-        while (date.compareTo(firstUnplannedDate) < 0) {
+
+        int nextDayOffset = 0;
+        while (nextDate.compareTo(firstUnplannedDate) < 0) {
```
triceo, line 682: Another loop that is somewhat hard to read.
Breaking it down into smaller parts would help.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/rotation/TimeBucket.java
Christopher-Chianelli: Hibernate/H2. Getting Cascade delete to work on an ElementCollection is non-trivial (see https://stackoverflow.com/questions/7695831/how-can-i-cascade-delete-a-collection-which-is-part-of-a-jpa-entity). I might be able to change it to https://stackoverflow.com/a/62848296

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/rotation/TimeBucket.java
Christopher-Chianelli: Neat; didn't know Optional have map

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/rotation/TimeBucket.java
Christopher-Chianelli: Sadly, that answer use hibernate specific annotation, so I am keeping it as is.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/employee/Employee.java
```diff
@@ -44,7 +47,17 @@
...
+    
+    @NotNull
+    @Size(min = 1, max = 3)
+    @Pattern(regexp = "^(?!\\s).*(?<!\\s)$", message = "Name should not contain any leading or trailing whitespaces")
+    private String shortId;
+    
+    @NotNull
+    @Size(min = 7, max = 7)
+    @Pattern(regexp = "^#[0-9A-F][0-9A-F][0-9A-F][0-9A-F][0-9A-F][0-9A-F]$")
+    private String color;
```
yurloc, line 59: I don't think this belongs to the domain model. Unless I'm missing something, the color and shortId are only needed for the rotation editor, which is strictly a UI feature. Color and shortID should go to frontend to keep the code cohesion high.

optaweb-employee-rostering-frontend/src/ui/pages/rotation/SeatJigsaw.tsx
```diff
@@ -0,0 +1,201 @@
...
+} from '@patternfly/react-core';
+import moment from 'moment';
+import { rosterSelectors } from 'store/roster';
+import { useSelector } from 'react-redux';
+import { EditIcon, TrashIcon } from '@patternfly/react-icons';
+import { TimeBucket } from 'domain/TimeBucket';
+import { EditTimeBucketModal } from './EditTimeBucketModal';
+import { Stub, EmployeeNickName } from './EmployeeStub';
+
+/* tslint:disable:no-nested-ternary react/no-array-index-key */
```
yurloc, line 30: Seems to be unused.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/employee/Employee.java
Christopher-Chianelli: @ge0ffrey advocated for it so that employee color/short id could also be used in Shift screens

optaweb-employee-rostering-frontend/src/ui/pages/rotation/SeatJigsaw.tsx
Christopher-Chianelli: Nested Ternary here:
` backgroundColor: seat ? seat.employee ? seat.employee.color : 'white' : 'gray',`

No array index key used here:
            {new Array(Math.ceil(rotationLength / 7)).fill(null).map((_, index) => index).map(weekNumber => (
              <FlexItem key={`${weekNumber}`}>

yurloc: - ternary - already disabled with `// eslint-disable-next-line no-nested-ternary` on line 84. No need to disable it for the whole file.
- index key - the example on line 68 doesn't use the index as the key. `weekNumber` is not the index, it's the element. So the "no array index key" rule is not broken in this case.

Also note that this line (30) uses a _tslint_ ignore comment but TSLint is obsolete. We no longer use TSLint.

optaweb-employee-rostering-frontend/src/ui/pages/rotation/SeatJigsaw.tsx
```diff
@@ -0,0 +1,201 @@
...
+          userSelect: 'none',
+        }}
+      >
+        {moment(props.timeBucket.startTime, 'HH:mm').format('LT')}
+        -
+        {moment(props.timeBucket.endTime, 'HH:mm').format('LT')}
+        {props.timeBucket.additionalSkillSet.length > 0
+          && (
+            <Text>
+          (
```
yurloc, line 58: Redundant parentheses.

Christopher-Chianelli:
@yurloc we want to give user the ability to pick colors and a short id for employees and persist that choice. (this is relevant outside of rotation also; other use cases include coloring shifts by employee). 

TODO:
Add i18n translations and maybe edit the color selector

optaweb-employee-rostering-frontend/src/ui/pages/employee/EmployeesPage.tsx
```diff
@@ -153,6 +166,18 @@ export class EmployeesPage extends DataTable<Employee, Props> {
...
       />,
+      <TextInput
+        key={3}
+        name="shortId"
+        defaultValue={data.shortId}
+        aria-label="shortId"
+        onChange={value => setProperty('shortId', value)}
+      />,
+      <ColorPicker
+        key={4}
```
yurloc, line 177: This is not a dynamic list of elements. I think the key is not needed.

yurloc:
Problems on the EmployeesPage:
- Color picker doesn't work (in row editing mode).
- When not editing, a hexadecimal code is displayed instead of a color.
- The client code crashes when sorting employees by color in descending order (http://localhost:3000/1/employees?sortBy=4&asc=false). "TypeError: sorter is not a function".

optaweb-employee-rostering-frontend/src/ui/pages/employee/EmployeesPage.tsx
Christopher-Chianelli: You still need keys for a static list of elements; For instance, if you add a button to the actions prop of a modal without a key, you get the missing key warning in the web console.

yurloc:
@Christopher-Chianelli Tests are failing. Please fix.

Christopher-Chianelli:
@yurloc done; I forgot to update the Employee List import to set color and short id, and I made the frontend generate a short id + color if you don't pick them

PR #462, opened 2020-07-16
Issue #274: Use uuid for generating keys in stack traces

### LINKED ISSUE #274: Duplicate React keys of stack trace elements
Server-side exceptions with long stack traces and multiple caused-by levels usually contain some duplicate stack trace elements. Using the whole line as a React key will result in duplicate keys:

```
index.js:1375 Warning: Encountered two children with the same key, `sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
    in div (created by ModalBoxBody)
    in ModalBoxBody (created by ModalContent)
    in div (created by ModalBox)
    in ModalBox (created by ModalContent)
    in div (created by FocusTrap)
    in FocusTrap (created by ModalContent)
    in div (created by Backdrop)
    in Backdrop (created by ModalContent)
    in ModalContent (created by Modal)
    in Modal (at ServerSideExceptionDialog.tsx:62)
    in ServerSideExceptionDialog (at Alerts.tsx:50)
    in Trans (at Alerts.tsx:116)
    in div (created by Alert)
    in div (created by Alert)
    in Alert (at Alerts.tsx:102)
    in div (at Alerts.tsx:87)
    in Alerts (created by ConnectFunction)
    in ConnectFunction (at App.tsx:41)
    in section (created by PageSection)
    in PageSection (at App.tsx:32)
    in main (created by Page)
    in div (created by Page)
    in Page (at App.tsx:30)
    in App (at src/index.tsx:45)
    in Suspense (at src/index.tsx:44)
    in Router (created by BrowserRouter)
    in BrowserRouter (at src/index.tsx:43)
    in Provider (at src/index.tsx:42)
```

Would be nice to avoid this.

To reproduce, do something that results in a server-side error and click "See Stack Trace" on the error alert. Check browser console.

Christopher-Chianelli:
Fixes #274

### DIFF
--- a/optaweb-employee-rostering-frontend/src/ui/components/ServerSideExceptionDialog.tsx
+++ b/optaweb-employee-rostering-frontend/src/ui/components/ServerSideExceptionDialog.tsx
@@ -17,7 +17,6 @@ import * as React from 'react';
 import { ServerSideExceptionInfo } from 'types';
 import { Text, Title, Button, ButtonVariant, Modal } from '@patternfly/react-core';
 import { useTranslation } from 'react-i18next';
-import { v4 as uuid } from 'uuid';
 
 function createStackTrace(exceptionInfo: ServerSideExceptionInfo|null): JSX.Element {
   if (exceptionInfo === null) {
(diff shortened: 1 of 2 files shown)


Christopher-Chianelli:
In regards to "no random key", I honesty think React should have `nokey` when there is no way to identify the same object across sibling (example: Creating a color palette of N colors. The color palette is NOT persisted in the database, so the colors in the palette do not have id's. But you could add/remove colors from the palette, meaning using index as key isn't right either. Now if the color in the palette has state, the state could be inherited by the wrong component using index. With rerendering the component each time, that risk is avoided (at a performance cost).)

In this case, I agree with removing random keys (although since this component will not be rerendered,  it would not make a performance difference).

yurloc:
You need to provide a way to identify elements if you want effective re-rendering. That's how React is designed. In a static list (like the stack trace), `key={index}` will work although it's not the best practice. The only issue is an ESLint warning, which is not a real issue.

In the color palette example, if same colors are permitted, you have to be creative :-). I would say randomly generated IDs would be OK, but you have to generate the ID when the color object is created and it must not ever change. You simply cannot generate a different value every time the component is re-rendered. The price for that would be poor performance.

PR #464, opened 2020-07-23
Fix issue when drag full day availability time range

Zhwt:
`dropLocation.end` is already the correct date, no need to add one extra day


Christopher-Chianelli:
ok to test

IIRC, the reason we add one extra day is if you click on the day row, start and end day will be set to the same time for availability creation.

Zhwt:
Thank you for your response, I believe the following code is adding one extra day when creating new availability, so we don't need to add one extra day when dragging or resizing:

https://github.com/kiegroup/optaweb-employee-rostering/blob/d655e930cac9d407539697e579b699177d11c53c/optaweb-employee-rostering-frontend/src/ui/components/calendar/Schedule.tsx#L88-L97

PR #470, opened 2020-07-28
Change configuration to new OCP 4.5.1 standard

dupliaka:
Made runOnOpenShift.sh to store configurations into Deployments instead of DeploymentsConfig.

The same fix as for VRP, see #334

@Christopher-Chianelli @yurloc


Christopher-Chianelli:
Motivation can be found at https://github.com/kiegroup/optaweb-vehicle-routing/pull/334

PR #478, opened 2020-08-04
PLANNER-936: Add OptaPlanner's code formatter to the backend

Christopher-Chianelli:
Only changes in this PR are:
- adding automatic format on compile of back end (copied from optaplanner-core)
- any formatting changes from said format

### DIFF
--- a/optaweb-employee-rostering-backend/pom.xml
+++ b/optaweb-employee-rostering-backend/pom.xml
@@ -284,21 +284,7 @@
           </plugin>
         </plugins>
       </build>
-    </profile>
-    <profile>
-      <!-- Fail the build in CI if code does not follow the standards. -->
-      <id>validate-formatting-ci</id>
-      <activation>
-        <property>
-          <name>env.CI</name>
-        </property>
-      </activation>
-      <properties>
-        <formatter.goal>validate</formatter.goal>
-        <impsort.goal>check</impsort.goal>
-      </properties>
-    </profile>
-    
+    </profile> 
   </profiles>
 
 </project>
--- a/optaweb-employee-rostering-backend/pom.xml
+++ b/optaweb-employee-rostering-backend/pom.xml
@@ -31,11 +31,6 @@
 
   <properties>
     <java.version>1.8</java.version>
-    <format.directory>${project.basedir}/../ide-configuration</format.directory>
-    <formatter.skip>false</formatter.skip>
-    <formatter.goal>format</formatter.goal>
-    <!-- TODO name impsort should indicate it's related to formatting -->
-    <impsort.goal>sort</impsort.goal>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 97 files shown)


optaweb-employee-rostering-backend/pom.xml
```diff
@@ -230,65 +235,102 @@
...
+              <goals>
+                <goal>${impsort.goal}</goal>
+              </goals>
+            </execution>
+          </executions>
+        </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-checkstyle-plugin</artifactId>
+          <version>3.1.0</version>
```
triceo, line 277: I'd remove this, just like we removed it in core.

Christopher-Chianelli: It not removed in core; that the only reason why I kept it.
`optaplanner-build-parent/pom.xml`:
```
      <plugin>
        <artifactId>maven-checkstyle-plugin</artifactId>
      </plugin>
```

triceo: Interesting, must have missed that one when I was removing the checkstyle config. Will finish it. :-)

optaweb-employee-rostering-backend/pom.xml
```diff
@@ -230,65 +235,53 @@
             </excludedTestClasses>
           </configuration>
         </plugin>
+        <plugin>
+          <groupId>net.revelc.code.formatter</groupId>
+          <artifactId>formatter-maven-plugin</artifactId>
+          <version>2.11.0</version>
+          <configuration>
+            <configFile>${format.directory}/eclipse-format.xml</configFile>
```
yurloc, line 243: 1. The plugin configuration should go to the parent's `<pluginManagement>` so that it can be reused in child modules.
2. The plugin should be added to the build (`<build>/<plugins>`) in modules that contain Java source code (backend, benchmark).

yurloc (changes requested):
The formatter and impsort plugins should run in the benchmark module as well. Otherwise looks good.

I would recommend to merge this after finishing the work in progress. It will be easier to rebase this PR, drop conflicting changes and rerun formatter than the other way around.

optaweb-employee-rostering-backend/pom.xml
```diff
@@ -333,6 +285,20 @@
         </plugins>
       </build>
     </profile>
+    <profile>
```
yurloc, line 288: One last thing. This profile should go to the parent as well so that the build fails if formatting is broken in benchmark or any other module.

optaweb-employee-rostering-backend/pom.xml
```diff
@@ -332,7 +284,7 @@
           </plugin>
         </plugins>
       </build>
-    </profile>
+    </profile>
```
yurloc, line 287: Trailing whitespace. Would be nice to get rid of that before merging the PR.

PR #481, opened 2020-08-06
PLANNER-2105: Update TimeBuckets on an interval to prevent multiple concurrent requests updating the same TimeBucket

Christopher-Chianelli:


optaweb-employee-rostering-frontend/src/ui/pages/rotation/SeatJigsaw.tsx
```diff
@@ -34,10 +35,66 @@ export interface SeatJigsawProps {
...
+  const [editedTimeBucket, setEditedTimeBucket] = React.useState(props.timeBucket);
+
+  // Update from props when timebucket id or version changes
+  React.useEffect(() => {
+    if (props.timeBucket.version !== editedTimeBucket.version || props.timeBucket.id !== editedTimeBucket.id) {
+      setEditedTimeBucket(props.timeBucket);
+    }
+  }, [editedTimeBucket.id, editedTimeBucket.version, setEditedTimeBucket, props.timeBucket]);
+
+  // Every 5 seconds, update the timeBucket if it change (so we don't overwhelm the server)
```
yurloc, line 51: I'm afraid this is not the right approach. If user closes the browser within 5 seconds after editing the bucket, the changes will be lost.

Christopher-Chianelli: The issue is caused from drag events, which fire multiple times per second (sometimes on the same seat). An alternative would be to fire on mouse released, although it would have to be a general "mouse release", as we don't know what component the mouse will be released on. The exception is "I can't complete that request because someone else is already working on that entity" (from the multiple drag requests).

Christopher-Chianelli:
@yurloc Changed it to update on mouse-leave or mouse-up on the flex list that contains all the seats in the timebucket.

optaweb-employee-rostering-frontend/src/ui/pages/rotation/SeatJigsaw.tsx
```diff
@@ -35,11 +35,35 @@ export interface SeatJigsawProps {
   onDeleteTimeBucket: () => void;
 }
 
+export const UPDATE_TIMEBUCKET_INTERVAL_DELAY = 5000;
```
yurloc, line 38: Unused const.

yurloc (approved):
One code style issue, otherwise looks good!

PR #497, opened 2020-09-18
Make planning items be in the future

dupliaka:
For sake of stable tests, we need to pass valid data that would dynamically adjust

Early the schedule has been trying to plan for the past, since dates
in the test have been written in the stone. Currently, the date of shift planning is related
to the current date and the planning activity is set to the Wednesday of the next week
So 18th of September you will test schedule on the 23rd of September (18 -> next week -> Wed)

### JIRA

[PLANNER-2110](https://issues.redhat.com/browse/PLANNER-2110)

### DIFF
--- a/optaweb-employee-rostering-frontend/cypress/integration/newTenantBasicWorkflow.js
+++ b/optaweb-employee-rostering-frontend/cypress/integration/newTenantBasicWorkflow.js
@@ -77,7 +77,7 @@ describe('A new tenant can be created, who can have their own employees, spots, 
     cy.get('[data-cy=settings]').click();
     cy.get('[data-cy=add-tenant]').click();
     cy.get('[data-cy=name]').type('Test Tenant');
-    cy.get('[data-cy=schedule-start-date]').type(Cypress.moment().format('YYYY-MM-DD'));
+    cy.get('[data-cy=schedule-start-date]').type('2018-01-01');
     cy.get('[data-cy=draft-length]').type('7');
     cy.get('[data-cy=publish-notice]').type('7');
     // TODO: publish length is disabled/not supported by backend, uncomment when supported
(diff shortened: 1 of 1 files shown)


Christopher-Chianelli:
I don't think that is the issue. We want the tests to be stable, and typically that means using static dates (even if those dates are in the past). I think the main issue is different timezones have different week starts (in EST it Sunday, in Europe Monday), meaning we are choosing the wrong date to create the shift on. An adjust that should work REGARDLESS of week start is to create the shift 1 week past the default schedule. (And a better solution might involve using https://docs.cypress.io/api/commands/clock.html#Syntax to set the current date to be in the past)

optaweb-employee-rostering-frontend/cypress/integration/newTenantBasicWorkflow.js
```diff
@@ -77,7 +77,7 @@ describe('A new tenant can be created, who can have their own employees, spots,
     cy.get('[data-cy=settings]').click();
     cy.get('[data-cy=add-tenant]').click();
     cy.get('[data-cy=name]').type('Test Tenant');
-    cy.get('[data-cy=schedule-start-date]').type('2018-01-01');
+    cy.get('[data-cy=schedule-start-date]').type(Cypress.moment().format("YYYY-MM-DD"));
```
Christopher-Chianelli, line 80: I dislike not using a static date. I would prefer using https://docs.cypress.io/api/commands/clock.html#No-Args to change the current date.

Christopher-Chianelli (changes requested):
The fix (that is, creating the shift 1 week past the default) looks good, but I dislike not using a static date. I would prefer setting the container + system time to a static date.

optaweb-employee-rostering-frontend/cypress/integration/newTenantBasicWorkflow.js
dupliaka: No problem, could you only specify why the clock is more appropriate here since I was feeling that clock is just for time not a date.

Christopher-Chianelli: See https://docs.cypress.io/api/commands/clock.html#Now . JavaScript treats dates as time, time as dates.

dupliaka:
Jenkins retest this please

Christopher-Chianelli (approved):
LGTM

mareknovotny:
Jenkins retest this

rsynek:
The CI failures are unrelated (happen in an upstream repository). Given the fact that the remaining CI job passed and the PR stabilizes existing tests, I am merging it now.

PR #530, opened 2020-11-17
PLANNER-1752: Move frontend to Immutable.js

Christopher-Chianelli:
### JIRA

https://issues.redhat.com/browse/PLANNER-1752

### Referenced pull requests

<details>
<summary>
</summary>

</details>


yurloc:
Jenkins retest this.

optaweb-employee-rostering-frontend/src/store/contract/reducers.ts
```diff
@@ -14,32 +14,27 @@
...
 };
 
 const contractReducer = (state = initialState, action: ContractAction): ContractList => {
   switch (action.type) {
     case ActionType.SET_CONTRACT_LIST_LOADING: {
       return { ...state, isLoading: action.isLoading };
     }
-    case ActionType.ADD_CONTRACT: {
-      return { ...state, contractMapById: mapWithElement(state.contractMapById, action.contract) };
+    case ActionType.ADD_CONTRACT: case ActionType.UPDATE_CONTRACT: {
```
yurloc, line 33: 1 case per line please.

optaweb-employee-rostering-frontend/src/store/contract/selectors.ts
```diff
@@ -24,20 +25,19 @@ export const getContractById = (state: AppState, id: number): Contract => {
...
   }
   if (oldContractMapById === state.contractList.contractMapById && contractListForOldContractMapById !== null) {
     return contractListForOldContractMapById;
   }
-
-  const out: Contract[] = [];
-  state.contractList.contractMapById.forEach((value, key) => out.push(getContractById(state, key)));
+  const out = state.contractList.contractMapById.keySeq().map(id => getContractById(state, id)).toList();
 
   oldContractMapById = state.contractList.contractMapById;
```
yurloc, line 39: IDEA says:
>Assigned expression type Map<number, DomainObjectView<Contract>> is not assignable to type Map<number, Contract> | null 
which kinda makes sense since
```ts
let oldContractMapById: Map<number, Contract> | null = null;
```
versus
```ts
readonly contractMapById: Map<number, DomainObjectView<Contract>>;
```
I wonder why `npm run typecheck` doesn't fail on this.

Or am I missing something?

optaweb-employee-rostering-frontend/src/store/employee/employee.test.ts
```diff
@@ -276,19 +274,21 @@ describe('Employee reducers', () => {
...
-      employeeMapById: mapWithElement(storeState.employeeList.employeeMapById, addedEmployee) });
+      employeeMapById: storeState.employeeList.employeeMapById
+        .set(addedEmployee.id as number, mapDomainObjectToView(addedEmployee)) });
   });
   it('remove employee', () => {
     expect(
       reducer(state.employeeList, actions.removeEmployee(deletedEmployee)),
     ).toEqual({ ...state.employeeList,
-      employeeMapById: mapWithoutElement(storeState.employeeList.employeeMapById, deletedEmployee) });
+      employeeMapById: storeState.employeeList.employeeMapById.delete(deletedEmployee.id as number) });
```
yurloc, line 284: Alternatively, you can say `delete(deletedEmployee.id!)`. The exclamation mark is a non-null assertion. It's more concise and maybe a more precise type assertion. But we have an ESLint rule that complains about that. Dunno which of the two is worse.

optaweb-employee-rostering-frontend/src/ui/pages/contract/ContractsPage.test.tsx
```diff
@@ -36,14 +37,14 @@ describe('Contracts page', () => {
 
   it('should render the viewer correctly', () => {
     const contractsPage = new ContractsPage(twoContracts);
-    const contract = twoContracts.tableData[1];
+    const contract = twoContracts.tableData.get(1) as Contract;
```
yurloc, line 40: Rant about the Immutable API: this looks very inconvenient. `get()` returns `V | undefined`. This is similar to throwing a checked exception in Java. And the `as V` type assertion is similar to swallowing the exception.

Is there a way to avoid it? Like using a method that simply returns `V` or throws an "unchecked" exception?

yurloc (commented):
First round of review. Minor issues detected.

My IDE shows errors on some parts of the code but typecheck passes. I need to investigate this.

Otherwise looks nice. Minimal code changes given the API transition.

optaweb-employee-rostering-frontend/src/store/contract/selectors.ts
Christopher-Chianelli: `DomainObjectView<Contract> = Contract` since `Contract` does not refer to any other domain objects (such as skill, spot, etc.). `DomainObjectView<Employee> != Employee`, since Employee refers to `Skill` and `Contract` (which will be replaced by numbers in their definitions. 

Now that https://github.com/microsoft/TypeScript/issues/12754 is fixed via https://github.com/microsoft/TypeScript/pull/40336, if and when we update to a new typescript release, I can remove the two different definitions of DomainObjectViews in the store (which has to deal with converting spotId -> spot), which would make this a type error.

optaweb-employee-rostering-frontend/src/ui/pages/contract/ContractsPage.test.tsx
Christopher-Chianelli: We would need to provide a not-null default value. Ideally, something that is never so the type `<T | never>` resolves to `T`. https://immutable-js.github.io/immutable-js/docs/#/List/get

Christopher-Chianelli: Wait, wouldn't the `!` trick you posted above work?

yurloc: Yes, it works. But let me ask this question: what happens if you try to `get()` contract with a wrong index?

In Java Collections API, the code would throw a runtime exception indicating a programming error (you haven't made absolutely sure the index is valid using `containsKey()` or `size()`) and you're immediately aware of the problem because the program crashes or displays an error in the UI.

An alternative approach (still talking about Java) is to return `null` if the required element is not found. Unfortunately, in Java, `null` is a permitted value for any object type so as a programmer you bear the burden of making null checks or facing possible NPEs.

According to my quick research, Immutable.js takes the latter approach where it simply returns `undefined` if the element doesn't exist instead of throwing an error. This is reflected in the return type: `get(index: number): T | undefined;`.

Since we're using the [strictNullCheck](https://www.typescriptlang.org/docs/handbook/advanced-types.html#nullable-types) option with the TS compiler, variables are not allowed to include `null` or `undefined` unless declared so using the union type (`T | undefined`).

Using the type assertion (`!`) we erase the information that the variable may contain `undefined`. I would argue this is unsafe and we should check for `undefined` first and probably do that as soon as possible to fail fast.

yurloc: Using a [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-type-assertions) function should allow us to erase `undefined` safely and react accordingly if it's detected (show error popup, log error in the console, etc.). This is probably not necessary in tests but I think we should use type assertion very carefully in the production code.

Christopher-Chianelli: There an another way:
```
function error(msg? str): never {
    throw Error(msg ?? "")
}

const x = myList.get(0) ?? error();
```

The "??" operation returns the first argument if it is not null/undefined, otherwise it returns the second argument. It a short circuited operation, so the second argument is not executed unless the first argument is null or undefined. Since `error` return type is never, the type of x is the type of elements in myList. On the case that the impossible did happen, an error is thrown. This basically converts the null approach to the throw error approach

yurloc: Looks like a good idea. Can you POC that in this PR?

Christopher-Chianelli: I did it in a stash of this PR (only did it in one part). The application worked and tests passed. Still need to test the unhappy path via manually inserting breakage. I will push when I completed it in the rest of the places.

Christopher-Chianelli: Pushed; can confirm it throws an error when get returns undefined (it actually happened in two of our existing tests, since we used type assertions when spot could be undefined (which was checked on later in the code anyways)).

optaweb-employee-rostering-frontend/src/ui/pages/shift/ExportScheduleModal.tsx
```diff
@@ -121,10 +122,10 @@ export const ExportScheduleModal: React.FC<StateProps & OwnProps> = (props) => {
           <MultiTypeaheadSelectInput
             aria-label={t('forSpots')}
             emptyText={t('selectSpots')}
-            value={exportedSpots}
-            options={props.spotList}
+            value={exportedSpots.toArray()}
+            options={props.spotList.toArray()}
             optionToStringMap={spot => spot.name}
-            onChange={setExportedSpots}
+            onChange={newExportedSpots => setExportedSpots(List(newExportedSpots))}
```
yurloc, line 128: This seems like an unnecessary complexity. Notice that every time user changes the input and the component is rerendered, a conversion from array to List and from List to array happens. For what benefit?

I think we should either A) draw a clear line between the Redux store and the React UI and only use Immutable.js in the store, where side effects must be avoided or B) use Immutable collections everywhere and only do the `toArray()` conversion when sending data to third-party components.

I'm slightly inclined toward option A at this moment.

Christopher-Chianelli: I prefer A) since then we don't need to use a type cast or a type guard, although mutable values in components can cause hard to find bugs in the frontend (I am looking at you, moment...). Arrays aren't typically modified directly in components, so they less prone to bugs then dates.

Christopher-Chianelli: Thoughts on the selectors? (that is, should selectors return the immutable List, or the List.toArray? If the second, should we memoize the List or List.toArray? (the problem with memozing List.toArray is that an modifications to it in one component affects all components.)

yurloc: Ad Moment: #532.

yurloc: I would say let's do the minimum necessary changes to get rid of the in-house immutable collections support (`withElements` etc.) without introducing significant performance regression.

When this is done, we can take a look at the next steps:
- Replace `mapStateToProps` with [useSelector](https://react-redux.js.org/api/hooks)?
- [Memoize selectors with Reselect](https://programmerden.com/2020/04/06/optimize-your-redux-selectors-with-useselector-hook-and-memoize-them-with-reselect/) for performance optimization or not?

optaweb-employee-rostering-frontend/src/ui/pages/employee/EmployeesPage.tsx
```diff
@@ -66,9 +66,9 @@ const mapStateToProps = (state: AppState, ownProps: Props): StateProps => ({
     ownProps.t('shortId'),
     ownProps.t('color'),
   ],
-  tableData: employeeSelectors.getEmployeeList(state),
-  skillList: skillSelectors.getSkillList(state),
-  contractList: contractSelectors.getContractList(state),
+  tableData: employeeSelectors.getEmployeeList(state).toArray(),
+  skillList: skillSelectors.getSkillList(state).toArray(),
+  contractList: contractSelectors.getContractList(state).toArray(),
```
yurloc, line 71: @Christopher-Chianelli I think one of the questions from the other thread is about this. I would say let's not do this. The selector function should return exactly the data that the UI component needs (so the `toArray()` transformation should happen inside the selector).

I'm not 100% sure but I think about components using plain arrays like this: yes, we give up immutability but we get a simpler code (`employees[0]` vs `employees.get(0) ?? error()`). Correct me if I'm wrong but IMO we don't need immutability in components because they do not modify (redux) state directly (only reading props, dispatching actions).

optaweb-employee-rostering-frontend/src/ui/pages/shift/ExportScheduleModal.tsx
Christopher-Chianelli: Some `mapStateToProps` are used in class components, so we won't be able to use `useSelector` there (most new functional components already use `useSelector`.

The result is already memoized manually.

optaweb-employee-rostering-frontend/src/ui/pages/employee/EmployeesPage.tsx
Christopher-Chianelli: Correct. That question had two parts: 1) Should the selector return List or array, and 2) If array, should the selector clone it (providing local immutable at a performance cost) or cache a single result array (mutable but faster)

yurloc: Return mutable array, I would say.

Christopher-Chianelli:
@yurloc thoughts on the current state of the PR?

optaweb-employee-rostering-frontend/src/ui/Alerts.test.tsx
```diff
@@ -85,7 +85,7 @@ const someAlerts: Props = {
       id: 0,
       createdAt: date,
       i18nKey: 'infoMessage',
-      variant: 'info',
+      variant: 'info' as 'info',
```
yurloc, line 88: This seems redundant. I confirm that I saw an error after removing this at some point but now it works without the assertion (on my machine). Can you check and remove the assertion if it works for you as well?

yurloc (approved):
Sorry for taking so long, I missed the update last week.

LGTM, one small comment.

PR #542, opened 2020-12-08
Use @ManyToMany instead of @OneToMany in TimeBucket additionalSkillSet

### LINKED ISSUE #541: Can the unique key constraint of  additional_skill_set_id  field be removed?In the table of time_bucket_additional_skill_set
This constraint causes me to create timebuckets of different periods, but cannot select the same skill. This is unreasonable. 
If I forcibly remove constraints from the database and insert duplicate values, will the scheduling result be incorrect?

Christopher-Chianelli:
Using `@OneToMany` adds an additional constraint of no two time buckets
sharing any additional skills, which is unwarranted. The relation
should be `@ManyToMany` instead.

Fixes #541.

### Referenced pull requests

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/rotation/TimeBucket.java
+++ b/optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/rotation/TimeBucket.java
@@ -31,10 +31,13 @@ import java.util.Set;
 import javax.persistence.CollectionTable;
 import javax.persistence.ElementCollection;
 import javax.persistence.Entity;
+import javax.persistence.FetchType;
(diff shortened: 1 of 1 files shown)


triceo (approved):
I don't really like to see `FetchType.EAGER`, as it's a great way to kill your performance.
That said, there may be a valid reason why you chose to use it - so I'm just pointing it out, and not  blocking the PR on it.

Christopher-Chianelli:
@triceo We send the entire object during REST calls, so we need to do eager fetching (although for our frontend, we basically disregard these copies and store them separately; I agree it not ideal, but the backend is probably already being consumed by something else, so we can't break API).

PR #558, opened 2021-02-03
Upgrade Patternfly Version, Refactor DataTable to be composable

Christopher-Chianelli:
Still need to add back some tests; this refactor changes quite a bit of the DataTable API, so quite a few tests need to be rewritten. 

### JIRA

Should resolve
https://issues.redhat.com/browse/PLANNER-2321

### Referenced pull requests

<details>
<summary>
</summary>

</details>


yurloc (commented):
Issues found:
- The background image disappeared.
- Admin page filtering broken. The screen goes blank after typing two characters in the search box.

yurloc:
I went through the code and found a few minor issues (mostly typos). Two functional issues: background has disappeared, admin page filtering is broken.

Otherwise looks good. I haven't yet fully grasped the change but it seems the data table was greatly simplified.

Christopher-Chianelli:
Cause of functional issue 1: passing`TenantRow` instead of `tenant => <TenantRow {...tenant} />` to `<TheTable>`, causing a hard crash in React since the number of hooks rendered in a single rendering step is not the same (see https://medium.com/@shanplourde/react-hooks-rendered-more-hooks-than-during-the-previous-render-d2c026d7cca3).

Christopher-Chianelli:
Cause of second issue: opaque background-color of header blocking the background

Christopher-Chianelli:
The summary of this change is using the new Patternfly Composable Table which was introduced by the patternfly upgrade: https://www.patternfly.org/v4/components/table#tablecomposable-examples

This make it easy to compose tables that do not use text based components (which the previous patternfly table model lack a good interface to).

Result is I was able to remove the mix of mutability and stateless input components (all input components are now stateless). I suspect the mix of them cause a hard to reproduce issue where sometimes the skill list of a spot/employee doesn't update in the datatable when it is updated in the component (causing a bad REST response to be sent).

So this new model is simpler and had less bugs, which is a plus.

yurloc (commented):
A few more typos.

Otherwise looks good. Please review the code coverage drop but my impression is that we're getting rid of a lot of complicated code and what remains is more of a UI structure and less of the error-prone logic so not much that would need thorough testing.

yurloc (approved):
Looks good! Let's merge.

PR #565, opened 2021-02-08
Fix process spawn by exec-maven-plugin

dupliaka:
Exec plugin not releasing the process and blocks to proceed with other mvn commands.
So when Jenkins run mvn deploy after test execution ends it got some resource files locked and unable to mvn clean it.

Details on the issue could be found here: https://stackoverflow.com/questions/9937402/process-spawned-by-exec-maven-plugin-blocks-the-maven-process

I use https://github.com/bazaarvoice/maven-process-plugin for both: preparation BE and running cypress tests. So far so good it is working see green deployment here: https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/blue/organizations/jenkins/custom%2Fadupliak%2Fquickstarts-release-pipeline%2Foptaplanner-deploy/detail/optaplanner-deploy/63/pipeline

### Referenced pull requests

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaweb-employee-rostering-standalone/pom.xml
+++ b/optaweb-employee-rostering-standalone/pom.xml
@@ -147,7 +147,7 @@
                   <goal>start</goal>
                 </goals>
                 <configuration>
-                  <name>Run application</name>
+                  <name>ERP backend</name>
                   <healthcheckUrl>http://localhost:${application.port}</healthcheckUrl>
                   <workingDir>${project.build.directory}</workingDir>
                   <processLogFile>${project.build.directory}/${project.artifactId}.log</processLogFile>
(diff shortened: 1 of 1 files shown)


Christopher-Chianelli (commented):
Code changes LGTM; will approve after verifying it works

Christopher-Chianelli:
@dupliaka is the maven build suppose to pass if the integration test fail? Tested the PR on both passing (normal) and failing (changed the test to fail) tests, and the maven build passes for both.

optaweb-employee-rostering-standalone/pom.xml
```diff
@@ -147,7 +147,7 @@
                   <goal>start</goal>
                 </goals>
                 <configuration>
-                  <name>Run application</name>
+                  <name>ERP backend</name>
```
rsynek, line 150: I think that hardly anybody recognizes what "ERP" is; moreover, it's not just the backend, but the entire standalone application bundled in a single JAR file.

dupliaka: True

dupliaka:
> @dupliaka is the maven build suppose to pass if the integration test fail? Tested the PR on both passing (normal) and failing (changed the test to fail) tests, and the maven build passes for both.

Yes, that is expected - the thing is that build should not fail if the cypress test is failing - That was made for CI stabilization full downstream build. Since the integration-tests fails often it should not interrupt fdb. So it does not fail the build - but it issues the test reports - that signals Jenkins to mark the build as unstable but do not interrupt others builds

dupliaka:
> @dupliaka is the maven build suppose to pass if the integration test fail? Tested the PR on both passing (normal) and failing (changed the test to fail) tests, and the maven build passes for both.

I should add that I added it only to ERP
The crucial part here is when the test failed, but didnt left any report - then I will miss that something gone wrong. From other side this situation could happen only if some environment issue have happend. Last time I was asked to do so, but got the same suspicions. So decided to test it first on ERP.

Options:
1. Do not mark build as failed if test failed. Check test report. (Failure can be missed locally, could be missed due to env issue)
2. Mark build as failed if test failed. (Will brake 3 h full downstream build)
3. Remove test from fdb (Bug can be missed completely)
4. If fdb failed with the test fix it as soon as possible or ignore while you fix that(Will require to fdb again - means loss of 4 h)
@rsynek  @Christopher-Chianelli  WDYT?

rsynek:
I think the problem here is that, ideally, if the test fails, locally we want the maven build to fail too. However, on Jenkins, we want the Jenkins build to become `unstable` and not `failed`. The question is if such behavior is achievable.

PR #572, opened 2021-02-17
PLANNER-2198: Migrate backend to Quarkus

Christopher-Chianelli:
Get this error during the build:
[ERROR] 	[error]: Build step io.quarkus.hibernate.orm.deployment.HibernateOrmProcessor#parsePersistenceXmlDescriptors threw an exception: java.lang.IllegalAccessError: class io.quarkus.hibernate.orm.deployment.QuarkusPersistenceXmlParser tried to access private method 'void org.hibernate.jpa.boot.internal.PersistenceXmlParser.<init>(org.hibernate.boot.registry.classloading.spi.ClassLoaderService, javax.persistence.spi.PersistenceUnitTransactionType)' (io.quarkus.hibernate.orm.deployment.QuarkusPersistenceXmlParser and org.hibernate.jpa.boot.internal.PersistenceXmlParser are in unnamed module of loader io.quarkus.bootstrap.classloading.QuarkusClassLoader @5b6edcbf)
[ERROR] 	at io.quarkus.hibernate.orm.deployment.QuarkusPersistenceXmlParser.<init>(QuarkusPersistenceXmlParser.java:36)
[ERROR] 	at io.quarkus.hibernate.orm.deployment.QuarkusPersistenceXmlParser.locatePersistenceUnits(QuarkusPersistenceXmlParser.java:28)
[ERROR] 	at io.quarkus.hibernate.orm.deployment.HibernateOrmProcessor.parsePersistenceXmlDescriptors(HibernateOrmProcessor.java:243)
[ERROR] 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[ERROR] 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[ERROR] 	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[ERROR] 	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
[ERROR] 	at io.quarkus.deployment.ExtensionLoader$2.execute(ExtensionLoader.java:920)
[ERROR] 	at io.quarkus.builder.BuildContext.run(BuildContext.java:277)
[ERROR] 	at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2415)
[ERROR] 	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1426)
[ERROR] 	at java.base/java.lang.Thread.run(Thread.java:834)
[ERROR] 	at org.jboss.threads.JBossThread.run(JBossThread.java:501)

### Referenced pull requests

<details>
<summary>
</summary>

</details>


yurloc:
I know I've seen this too and I resolved it somehow. But I can't remember more details at the moment.

yurloc:
I know! There's a version clash between what Hibernate version is required by Panache and the Hibernate version set somewhere in the POM parent line.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/admin/AdminController.java
```diff
@@ -16,36 +16,28 @@
...
+// @Api(tags = "Admin")
 public class AdminController {
 
     private final AdminService adminService;
 
+    @Inject
     public AdminController(AdminService adminService) {
         this.adminService = adminService;
-        Assert.notNull(adminService, "adminService must not be null.");
+        assert adminService != null;// adminService must not be null.
```
yurloc, line 34: Use [Objects.requireNonNull()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Objects.html#requireNonNull(T,java.lang.String)). Java assertions are turned off by default: https://docs.oracle.com/javase/7/docs/technotes/guides/language/assert.html. We never used them.

triceo: Agree. Java assertions are an obscure JVM feature that is barely ever used by anyone.
First of all because it needs to be enabled to do anything, and if enabled, it just fails the app like any other exception would do.

Christopher-Chianelli:
@yurloc I figured it out yesterday (and can confirm that the issue) when creating a minimum reproducer (didn't push changes yet since several other things are broken in subtle ways, but the backend can run).

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/admin/AdminController.java
Christopher-Chianelli: I am debating if we even need this if we are using Injection (which will cause a compile error if there is no implementation). It only good for catching errors within tests basically, but tests also support injection.

Christopher-Chianelli:
Backend now works in jar mode; dev mode and tests do not run, some error related to enhancing panache entities (might be another version clash)

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/admin/AdminController.java
yurloc: +1. For `@Inject`ed constructors it is a boiler-plate code with no added value.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/OptaWebExceptionHandler.java
```diff
@@ -26,12 +26,8 @@
 @ApplicationScoped
 public class OptaWebExceptionHandler implements ExceptionMapper<Throwable> {
 
-    private ExceptionDataMapper exceptionDataMapper;
-
     @Inject
-    public OptaWebExceptionHandler(ExceptionDataMapper exceptionDataMapper) {
-        this.exceptionDataMapper = exceptionDataMapper;
-    }
+    ExceptionDataMapper exceptionDataMapper;
```
yurloc, line 30: I would prefer if the fields stayed `private final`. It conveys an important piece of information to the reader: the field is initialized when an instance is created and can never ever change. In my opinion this reduces cognitive complexity because I can immediately see it is a dependency the never changes during the instance life cycle.

Christopher-Chianelli: Can't since Quarkus need to create a proxy of this class as it is `@ApplicationScoped` (and Quarkus seems to ignore constructor `@Inject` points and end up attempting to create a default no-args constructor)

Christopher-Chianelli:
Update production and dev work fine, tests are able to run (but there are still errors, seems related to RestAssured config sending plain/text by default instead of json)

Christopher-Chianelli:
Update: all tests are fixed; just need to fix standalone

triceo:
FYI I ran `master` with the Constraint Provider enabled instead of DRL.
I got no backend failures, but I got frontend CCEs coming from indictment utils. (Inidictment contents is probably different between CS and DRL.)

I expect the same to be true on this branch, too.

Christopher-Chianelli:
@triceo those indictments are fixed in this branch. In fact, you can run the frontend + backend without errors. Need to fix the combined "standalone" module.

Christopher-Chianelli:
Update: all modules are fixed and all tests are converted/passing . Some changes need to be done to the distribution and standalone assembly (need to zip quarkus-app, new location of jar file)

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/common/MultipartBody.java
```diff
@@ -0,0 +1,34 @@
...
+public class MultipartBody {
+
+    @FormParam("file")
+    @PartType(MediaType.APPLICATION_OCTET_STREAM)
+    public InputStream file;
+
+    @FormParam("fileName")
+    @PartType(MediaType.TEXT_PLAIN)
+    public String fileName;
+}
```
triceo, line 34: I have to say I do not understand several things here:

- How does this compile? It's a class, not an abstract class, but has methods without bodies.
- What's it for? It sounds like something that RESTEasy should be providing.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/roster/RosterState.java
```diff
@@ -42,7 +42,7 @@
     @NotNull
     private LocalDate firstDraftDate;
     @NotNull
-    private final Integer publishLength; // In number of days
+    private Integer publishLength; // In number of days
```
triceo, line 45: Did Quarkus require this and Spring Boot somehow didn't, or did the change slip in from some other PR?

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/rotation/TimeBucket.java
```diff
@@ -65,7 +65,7 @@
                     @JoinColumn(name = "day_id",
                             referencedColumnName = "id",
                             foreignKey = @ForeignKey(name = "DAY_FK",
-                                    foreignKeyDefinition = "FOREIGN KEY (day_id) references public.time_bucket (id)" +
+                                    foreignKeyDefinition = "FOREIGN KEY (day_id) references public.timebucket (id)" +
```
triceo, line 68: Any reason for this?
If possible, I'd like to keep as many things as possible as similar as possible.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/rotation/TimeBucket.java
```diff
@@ -76,7 +76,7 @@
                     @JoinColumn(name = "seat_id",
                             referencedColumnName = "id",
                             foreignKey = @ForeignKey(name = "SEAT_FK",
-                                    foreignKeyDefinition = "FOREIGN KEY (seat_id) references public.time_bucket (id)" +
+                                    foreignKeyDefinition = "FOREIGN KEY (seat_id) references public.timebucket (id)" +
```
triceo, line 79: Dtto.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/tenant/view/RosterConstraintConfigurationView.java
```diff
@@ -21,28 +21,72 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+
 public class RosterConstraintConfigurationView extends AbstractPersistable {
 
     @NotNull
     private DayOfWeek weekStartDay = DayOfWeek.MONDAY;
 
+    @JsonSerialize(using = HardMediumSoftLongScoreJacksonSerializer.class)
+    @JsonDeserialize(using = HardMediumSoftLongScoreJacksonDeserializer.class)
```
triceo, line 37: Not sure about Jackson, but JSON-B has a type that reduces this boilerplate a lot. (`JsonB adapter`.)
Personally, I don't see a reason for dragging all five tons Jackson into this, when a small  JSON-B impl would suffice.

optaweb-employee-rostering-backend/src/main/resources/solverConfig.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <solver>
   <scoreDirectorFactory>
-    <scoreDrl>org/optaweb/employeerostering/service/solver/employeeRosteringScoreRules.drl</scoreDrl>
-    <!--<constraintProviderClass>org.optaweb.employeerostering.service.solver.EmployeeRosteringConstraintProvider</constraintProviderClass>-->
+    <!--<scoreDrl>org/optaweb/employeerostering/service/solver/employeeRosteringScoreRules.drl</scoreDrl>-->
+    <constraintProviderClass>org.optaweb.employeerostering.service.solver.EmployeeRosteringConstraintProvider</constraintProviderClass>
```
triceo, line 5: Got any numbers for me? How much better/worse are we now?

triceo (commented):
I have left a few comments.

From the type of questions I'm asking, it should be obvious that my experience with Quarkus so far is minimal. Therefore I am tagging in @yurloc , who's been through a similar exercise recently and is therefore the right person for the "official" review.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/OptaWebExceptionHandler.java
yurloc: I can't confirm this. Please revisit the constructor injection option. If it still doesn't work for you, please share the details.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/SwaggerConfig.java
```diff
@@ -16,33 +16,24 @@
...
+     * }
+     * 
+     * @Bean
+     * public UiConfiguration uiConfiguration() {
+     * return UiConfigurationBuilder.builder()
+     * .docExpansion(DocExpansion.LIST)
+     * .build();
+     * }
+     */
 }
```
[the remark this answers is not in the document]
Christopher-Chianelli: I need to figure out how to do the swagger config via Quarkus, so a reminder to actually do that

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/common/MultipartBody.java
Christopher-Chianelli: https://quarkus.io/guides/rest-client-multipart

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/roster/RosterState.java
Christopher-Chianelli: Quarkus requires this for database deserialization

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/rotation/TimeBucket.java
Christopher-Chianelli: Differences in hibernate table name. Spring Boot uses "Time_Bucket", Quarkus uses "TimeBucket"

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/tenant/view/RosterConstraintConfigurationView.java
Christopher-Chianelli: This is purely for testing. For some reason RestAssured doesn't automatically pick up the OptaPlanner module (and registering it doesn't seem to work)

optaweb-employee-rostering-backend/src/main/resources/solverConfig.xml
Christopher-Chianelli: Drools - Constraint Stream performance is almost 1-to-1 according to the benchmark:
```
Drools: Average: 26,411/s Problem 0: 16,665/s Problem 1: 36,158/s
CS     : Average: 25,830/s Problem 0: 16,596/s Problem 1: 35,065/s
```

Christopher-Chianelli: (This is comparing Quarkus Drools to Quarkus Constraint Streams; need to rerun on Spring Boot)

Christopher-Chianelli: Default on Spring boot is faster than both:
```
Spring-Boot Drools: 
Average: 29,662/s Problem 0: 19,758/s  Problem 1: 39,567/s
```

Christopher-Chianelli: CS on Spring boot is also faster than both:
```
Spring-Boot Constraint Stream
Average: 29,105/s Problem 0: 21,401/s  Problem 1: 36,809/s
```

Christopher-Chianelli: Benchmark: Default Solver, warm up 30 seconds, 10 seconds spent limit

triceo: OK, that is good news.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/tenant/view/RosterConstraintConfigurationView.java
triceo: If it's just for testing, that makes the boilerplate even worse. :-(

Christopher-Chianelli: Sadness: it was going to be fixed, but cannot since it will interfere with native tests: https://github.com/quarkusio/quarkus/pull/5430 . I'll see if moving to jsonb fixes it

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/OptaWebExceptionHandler.java
Christopher-Chianelli: It works, but with a warning: `2021-02-22 12:36:19,581 WARN  [io.qua.res.com.dep.ResteasyCommonProcessor] (build-37) Classes annotated with @Provider should have a single, no-argument constructor, otherwise dependency injection won't work properly. Offending class is org.optaweb.employeerostering.OptaWebExceptionHandler`. ExceptionDataMapper isn't an interface and doesn't use CDI, so I can just create it in the constructor

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/tenant/view/RosterConstraintConfigurationView.java
Christopher-Chianelli: There a problem with JSON-B: It doesn't serialize map score values properly (it does serialize field score values properly).

triceo: Maybe a custom adapter needs to be written? 
Anyway, since OptaPlanner persistence does have a JSON-B option, we should understand how to fix this.

Christopher-Chianelli: My guess is it doesn't register map adapters, so it uses the default adapter instead of using the toString adapter; creating test on OptaPlanner core

Christopher-Chianelli: I can confirm JSON-B doesn't work properly for maps in optaplanner-core (They get serialized as `{key: {}}` in tests)

Christopher-Chianelli:
Look like SonarCloud doesn't know how to do code coverage on Quarkus classes (probably due to calls being routed to the generated Quarkus proxies)

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/SwaggerConfig.java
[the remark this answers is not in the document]
Christopher-Chianelli: Done (deleted file, added openapi tags/descriptions to all methods, remove old commented out apioperation annotations)

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/rotation/TimeBucket.java
Christopher-Chianelli: (The alternative is to rename the table name via the `@Entity` annotation)

triceo: I would.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/tenant/view/RosterConstraintConfigurationView.java
Christopher-Chianelli: Moved the adapter registration to tests

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/domain/rotation/TimeBucket.java
Christopher-Chianelli: Done for backwards compatibility with previous databases so no migration would be needed.

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/admin/SystemPropertiesRetriever.java
```diff
@@ -18,45 +18,57 @@
 
 import java.time.DateTimeException;
 import java.time.ZoneId;
+import java.util.Optional;
 
+import javax.enterprise.context.ApplicationScoped;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+@ApplicationScoped
```
yurloc, line 27: This could be done with `@ConfigurationProperties` but I don't think it's important now. We can improve it later.

https://quarkus.io/guides/config-reference

yurloc (commented):
Looks very good! Just a couple of suggestions.

One question: the swagger link in the UI is now broken. Do you plan to fix it or replace it somehow or should it be removed without replacement?

optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/service/roster/RosterController.java
```diff
@@ -35,210 +46,252 @@
...
+    @GET
+    @Path("/shiftRosterView")
+    @Operation(summary = "View Shift Roster", description = "Get a shift roster view between two dates")
+    public ShiftRosterView getShiftRosterView(@PathParam("tenantId") @Min(0) Integer tenantId,
+            @QueryParam("p") Integer pageNumber,
+            @QueryParam("n") Integer numberOfItemsPerPage,
+            @QueryParam("startDate") String startDateString,
+            @QueryParam("endDate") String endDateString) {
+        if (startDateString == null) {
+            throw new IllegalArgumentException("query parameter startDate is required");
```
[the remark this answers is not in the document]
Christopher-Chianelli: This error message is included in the JSON response. The Exception type is important (see ExceptionDataMapper).

optaweb-employee-rostering-standalone/src/main/java/org/optaweb/employeerostering/standalone/WebConfig.java
```diff
@@ -16,21 +16,38 @@
...
-                .setViewName("forward:/");
-    }
+    public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain) throws IOException, ServletException {
+        HttpServletRequest request = (HttpServletRequest) req;
+        HttpServletResponse response = (HttpServletResponse) res;
+        chain.doFilter(request, response);
 
-}
\ No newline at end of file
+        final String PATH = request.getRequestURI();
```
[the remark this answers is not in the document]
Christopher-Chianelli: I think beforehand it was a literal String constant before, but became a final variable sometime later w/o changing it name; changed it to lowercase

triceo:
@yurloc Can you please expedite your review? Let's get this merged this sprint.

yurloc:
@triceo Sure. I'll get back to this as soon as I undraft https://github.com/kiegroup/optaweb-vehicle-routing/pull/445.

@Christopher-Chianelli Can you fix the build and to run the app on OpenShift in the meantime?

Christopher-Chianelli:
@yurloc Done (the build was broken due to temporary broken optaplanner-core SNAPSHOT last week; I fixed runOnOpenshift.sh (forgot to update backend Dockerfile))

optaweb-employee-rostering-backend/Dockerfile
```diff
@@ -1,5 +1,5 @@
-FROM adoptopenjdk/openjdk8:ubi-minimal-jre
-COPY target/*-exec.jar /opt/app/optaweb-employee-rostering.jar
+FROM adoptopenjdk/maven-openjdk11:latest
```
triceo, line 1: Why? UBI for 11 exists just as well:
https://hub.docker.com/layers/adoptopenjdk/openjdk11/ubi-minimal-jre/images/sha256-02e631ef7445a5c84a021f494fc5c831a6b4264307570f0fa67e0fd681beed71?context=explore

Christopher-Chianelli: I copied the one from standalone; it can probabally work for UBI

optaweb-employee-rostering-backend/src/main/resources/application-production.properties
```diff
@@ -13,14 +13,9 @@
...
-# https://stackoverflow.com/questions/43905119/postgres-error-method-org-postgresql-jdbc-pgconnection-createclob-is-not-imple
-spring.jpa.properties.hibernate.jdbc.lob.non_contextual_creation=true
-
-# Disable H2 console enabled in the default profile
-spring.h2.console.enabled=false
+quarkus.datasource.db-kind=postgresql
+quarkus.datasource.jdbc.url=jdbc:postgresql://${DATABASE_HOST:postgresql}:5432/${DATABASE_NAME}
+quarkus.datasource.username=${DATABASE_USER}
+quarkus.datasource.password=${DATABASE_PASSWORD}
+quarkus.hibernate-orm.database.generation=update
```
yurloc, line 21: I think this file is unused. Quarkus only reads `application.properties`. Please confirm and remove this file.

triceo:
Interesting. Code coverage actually went down by 2 percentage points with the JaCoCo-related comment.

Christopher-Chianelli:
@triceo Just pushed changes that fix it (long story short, needed to bump Quarkus version by 1 minor (to 1.11.1.Final, from 1.11.0.Final) since Jacoco will ususally fail due to a file locking error. Quarkus 1.13 adds a Jacoco extension that removes the extra configuration needed).

yurloc (commented):
Pending review comments.

optaweb-employee-rostering-standalone/src/main/resources/application.properties
```diff
@@ -0,0 +1,12 @@
+quarkus.package.type=fast-jar
+
+# Swagger UI
+quarkus.swagger-ui.always-include=true
+quarkus.swagger-ui.path=/swagger-ui.html
+
+# Datasource
+# - using an embedded in-memory DB: http://h2database.com/html/features.html#in_memory_databases
+# - not closing the DB automatically: http://h2database.com/html/features.html#closing_a_database
+quarkus.datasource.db-kind=h2
```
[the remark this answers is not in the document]
yurloc: Ignore the previous comment. I had some pending comments and I couldn't find them, so I submitted the review.

PostgreSQL is configured using the `%postgres` profile - that's OK.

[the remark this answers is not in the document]
Christopher-Chianelli: We do want to keep the structure of the OptaWebs one-to-one if possible. Which means either adding `optaweb-employee-rostering-tests` (which defaults to h2) and change the default here to postgres, or modify `optaweb-vehicle-routing` to use this approach. It feel odd saying "run with the JAR from `optaweb-employee-rostering-tests" in the README, so I probably will add a `runLocally.sh` that does that. (The command changes either way since I am not using the legacy JAR)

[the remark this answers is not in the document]
yurloc: It's a PITA that we can't use the same JAR with different DBs with Quarkus. Will we end up with `standalone-postgres` and `standalone-h2` and switch between them depending on whether you `runLocally` or `runOnOpenShift` or based on a `--database` switch? Or can we only have a single `standalone` module with different profiles for PostgreSQL and H2 and rebuild it each time we `runLocally` or `runOnOpenShift`?

Is it OK to only run ITs against H2 or should we also cover the PostgreSQL configuration?

I think we need to figure out all of that but we can do that later.

yurloc (approved):
Some cleanup might be left to do (removing unused Dockerfiles/reusing Dockerfiles generated by Quarkus, alignment between this and vehicle routing) but that's out of scope of the current time frame for feature development freeze and none of that is a blocking issue.

CI is green, `runOnOpenShift.sh` works with the current CRC (1.23.1).

PR #583, opened 2021-03-03
Polishing

yurloc:
A couple of things I came across during the review that deserved improvement but were of of scope of #558.

<details>
<summary>
</summary>

</details>


optaweb-employee-rostering-frontend/src/ui/components/DataTable.tsx
```diff
@@ -194,21 +203,19 @@ export const DataTable = (props: PaginationData<any> & RouteComponentProps & Dat
...
-                    index: props.sortByIndex,
-                    direction: (props.isReversed ? 'desc' : 'asc') as 'desc' | 'asc',
-                  },
-                  onSort: () => props.onSorterChange(index),
-                  columnIndex: index,
-                },
-              } : {};
+                sortBy: {
+                  index: props.sortByIndex,
+                  direction: (props.isReversed ? 'desc' : 'asc'),
```
Christopher-Chianelli, line 208: This doesn't cause a type error? When I tried it, it type resolved to 'string' in my IDE, but 'direction' only accept type 'desc' | 'asc'.

optaweb-employee-rostering-frontend/src/ui/header/Toolbar.tsx
```diff
@@ -117,48 +117,38 @@ export class ToolbarComponent extends React.Component<Props, ToolbarState> {
...
-          }}
-        >
-          <ToolbarGroup />
-          {bellAndCog}
-        </Toolbar>
-      );
-    }
-
-    const currentTenant = tenantList.find(t => t.id === currentTenantId) as Tenant;
+    const currentTenantName = tenantList.find(t => t.id === currentTenantId)?.name ?? 'TENANT ERROR';
```
Christopher-Chianelli, line 120: Should this be i18n?

Christopher-Chianelli (approved):
LGTM

optaweb-employee-rostering-frontend/src/ui/header/Toolbar.tsx
yurloc: You mean the `TENANT ERROR` string? I don't think so. I think it can only happen as a result of a bug - the component's props would have be a `currentTenantId` and a `tenantList` that doesn't have a tenant with such ID. This shouldn't be possible given enough test coverage. Therefore I consider the fallback string as a hypothetical situation. It will never hit the user's eyes so I think it's not worth translating.

optaweb-employee-rostering-frontend/src/ui/components/DataTable.tsx
yurloc: No, IntelliJ IDEA 2020.3 doesn't complain and `npm run typecheck` passes without any errors.

PR #587, opened 2021-03-09
Configure docker to run as non-root user

dupliaka:
I need to run docker with non-root so to be able to delete files generated inside the container.

Though for podman it is opposite - if you run it as a non-root it is unable to write to the binded volume

I used here 1000:1000 as user and group id since it is a default non-root. So it exist both in container and on host so to not get noisy notifications like:

"groups: cannot find name for group ID"
"I have no name!"
  - your container, complaining
  
That is required for setting up optawebs in the pipeline. So here I made a short version of the pipeline which is also require to be able to remove cypress reports generated in docker

https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/custom/job/adupliak/job/quickstarts-release-pipeline/job/optaplanner-deploy/83/

Pay attention to the owner it is not root anymore
+ ls -la optaweb-employee-rostering-frontend/target/test-reports/
total 4
drwxr-xr-x. 2 jenkins jenkins   30 Mar  9 12:53 .
drwxrwxr-x. 5 jenkins jenkins  165 Mar  9 12:53 ..
-rw-r--r--. 1 jenkins jenkins 1017 Mar  9 12:53 TEST-cypress.xml

### Referenced pull requests

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaweb-employee-rostering-standalone/pom.xml
+++ b/optaweb-employee-rostering-standalone/pom.xml
@@ -191,6 +191,8 @@
                     <argument>${project.parent.basedir}/${frontend.project.name}:/e2e:Z</argument>
                     <argument>-w</argument>
                     <argument>/e2e</argument>
+                    <argument>${user.flag}</argument>
+                    <argument>${user.name.group}</argument>
                     <argument>--entrypoint</argument>
                     <argument>node</argument>
                     <argument>cypress/included:${version.cypress.docker}</argument>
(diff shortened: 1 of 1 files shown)


optaweb-employee-rostering-standalone/pom.xml
```diff
@@ -191,6 +191,8 @@
                     <argument>${project.parent.basedir}/${frontend.project.name}:/e2e:Z</argument>
                     <argument>-w</argument>
                     <argument>/e2e</argument>
+                    <argument>${user.flag}</argument>
+                    <argument>${user.name.group}</argument>
```
Christopher-Chianelli, line 195: What are these parameters in the case of podman?

Christopher-Chianelli (approved):
I use podman-docker (i.e. `alias docker podman`), so I won't be able to check this works for non-root docker, but the code looks good

optaweb-employee-rostering-standalone/pom.xml
dupliaka: Then user.flag and user.name.group are empty so executor plugin skips that and podman launches the same way as before

PR #861, opened 2022-11-23
[BXMSPROD-1881] dependabot for main, 7.67.x and blue branches

Ginxo:


mareknovotny (changes requested):
here we need only main and 8.13.x to be scanned

mareknovotny (approved):

PR #887, opened 2023-04-11
update npm to version 6.14.18

rgdoliveira:
The reason to try to update npm version here is that: Indy/PNC "sidecar" uses TLS v.1.3 and projects **optaweb-employee-rostering** and **optaweb-vehicle-routing** are using an old npm version and for some reason it seems they don't work reliably. So it means that we are having to rebuild those projects on PNC until they pass.

Maybe upgrading npm to a version that was released recently could fix that.

**Related PRs:**
- https://github.com/kiegroup/optaweb-employee-rostering/pull/887
- https://github.com/kiegroup/optaweb-vehicle-routing/pull/861

### DIFF
--- a/.github/workflows/full-downstream.yml
+++ b/.github/workflows/full-downstream.yml
@@ -25,7 +25,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         java-version: [11]
-        maven-version: ['3.8.1']
+        maven-version: ['3.8.6']
       fail-fast: false
     runs-on: ${{ matrix.os }}
     name: Full downstream build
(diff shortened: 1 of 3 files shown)


rgdoliveira:
Not sure if this upgrade will break something. if it breaks, we can just stick with the older version and try to build on PNC until it passes (as we have been doing so far) as the upgrade is not a certain that it will fix the issue and more just a trying.

rgdoliveira:
Build Chain failed for Java 11 and 17 when building kogito-runtimes project:
```[ERROR] Failed to execute commands for kiegroup/kogito-runtimes```

Error was:
```Detected Maven Version: 3.8.1 is not in the allowed range 3.8.6.```

I will push a new commit to update maven-version to 3.8.6 as it is required to build kogito-runtimes.

lampajr (approved):

mareknovotny (approved):

rgdoliveira:
@yurloc @Ginxo do you think we can merge this one?

Ginxo (approved):
LGTM

